PR #1, opened 2015-08-11
Better logging capabilities

pedrobaeza:
This PR adds 2 features:

* log decorator, for putting on methods to automatically log a line when the method is called.
* logged_progress method, to be called inside your scripts for showing a progress bar to know the advance of that part of the migration.

When the transition to the lib will be completed, I will put some examples on OpenUpgrade.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,19 +4,19 @@ History
 -------
 
 0.1.0 (2015-05-04)
----------------------
+------------------
 
 * First release on PyPI.
 
 0.1.1 (2015-05-05)
----------------------
+------------------
 
 * Fixes to the tests
 * Add more badges
 * Fix pip install issue with required.txt
 
 0.1.2 (2015-06-23)
----------------------
+------------------
 
 * Rewrite history with git filter-branch
 * Remove unneeded files from history
@@ -24,3 +24,11 @@ History
   * deferred80.py
   * openupgrade_log.py
   * #openupgrade_loading.py#
+
+0.1.3 (2015-08-11)
+------------------
+
+* Add method decorator log to automatically log a line when the method is
+  called
+* Add method logged_progress for showing a progress bar to know the advance
+  of that part of the migration
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -44,6 +44,7 @@ logger.setLevel(logging.DEBUG)
 
 __all__ = [
     'migrate',
+    'log',
     'load_data',
     'copy_columns',
     'rename_columns',
@@ -832,6 +833,29 @@ def reactivate_workflow_transitions(cr, transition_conditions):
         cr.execute(
             'update wkf_transition set condition = %s where id = %s',
             (condition, transition_id))
+
+
+def log():
+    """Decorator for automatic logging of the executed method.
+    :param custom_logger: for defining an specific logger
+    :param details: if True, more details are given in the log"""
+    def wrap(func, custom_logger=None, details=False):
+        def wrapped_function(*args, **kwargs):
+            msg = "Executing method %s" % func.__name__
+            if details:
+                if args:
+                    msg += " with args %s" % str(args)
+                if kwargs:
+                    if args:
+                        msg += " and kwargs %s" % str(kwargs)
+                    else:
+                        msg += " with kwargs %s" % str(kwargs)
+            custom_logger.info(msg)
+            func(*args, **kwargs)
+        if custom_logger is None:
+            custom_logger = logger
+        return wrapped_function
+    return wrap
 
 
 def migrate(no_version=False):
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -23,6 +23,7 @@ import sys
 import os
 import inspect
 import logging
+import progressbar
 from openerp import release, tools, SUPERUSER_ID
 from openerp.tools.yaml_import import yaml_import
 from openerp.osv import orm
@@ -45,6 +46,7 @@ logger.setLevel(logging.DEBUG)
 __all__ = [
     'migrate',
     'log',
+    'logged_progress',
     'load_data',
     'copy_columns',
     'rename_columns',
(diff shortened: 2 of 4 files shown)


HISTORY.rst
```diff
@@ -24,3 +24,11 @@ History
...
   * openupgrade_log.py
   * #openupgrade_loading.py#
+
+0.1.3 (2015-08-11)
+------------------
+
+* Added method decorator log to automatically log a line when the method is
+  called
+* Added method logged_progress for showing a progress bar to know the advance
+  of that part of the migration
```
bwrsandman: Use imperative: Add

openupgradelib/openupgrade.py
```diff
@@ -23,6 +23,7 @@
 import os
 import inspect
 import logging
+import progressbar
```
bwrsandman: Add progressbar to requirements with the version

openupgradelib/openupgrade.py
```diff
@@ -834,6 +837,48 @@ def reactivate_workflow_transitions(cr, transition_conditions):
...
+        def wrapped_function(*args, **kwargs):
+            msg = "Executing method %s" % func.__name__
+            if details:
+                if args:
+                    msg += " with args %s" % str(args)
+                if kwargs:
+                    if args:
+                        msg += " and kwargs %s" % str(kwargs)
+                    else:
+                        msg += " with kwargs %s" % str(kwargs)
```
bwrsandman: Isn't `str` redundant when you format with `%s`

openupgradelib/openupgrade.py
```diff
@@ -834,6 +837,48 @@ def reactivate_workflow_transitions(cr, transition_conditions):
...
+    if args is None:
+        args = []
+    if kwargs is None:
+        kwargs = {}
+    pbar = progressbar.ProgressBar(widgets=widgets,
+                                   maxval=len(iterator)).start()
+    for i, element in enumerate(iterator):
+        pbar.update(i)
+        func(element, *args, **kwargs)
+    pbar.finish()
```
bwrsandman: Use progressbar 2.3 feature:

```python
    pbar = progressbar.ProgressBar(widgets=widgets)
    for element in pbar(iterator):
        func(element, *args, **kwargs)
```

It's much nicer, especially for this straightforward case.

Also, add a space after the title.

openupgradelib/openupgrade.py
bwrsandman: Also, the egg isntallation seems to fail for certain versions of python.

openupgradelib/openupgrade.py
```diff
@@ -834,6 +837,48 @@ def reactivate_workflow_transitions(cr, transition_conditions):
...
+                        msg += " with kwargs %s" % str(kwargs)
+            custom_logger.info(msg)
+            func(*args, **kwargs)
+        if custom_logger is None:
+            custom_logger = logger
+        return wrapped_function
+    return wrap
+
+
+def logged_progress(func, iterator, title="", args=None, kwargs=None):
```
bwrsandman: Just curious, aren't you trying to use `*args` and `**kwargs`?

Perhaps you should add a comment explaining why it's not like that.

bwrsandman:
Add progressbar>=2.3 to requirements.txt

openupgradelib/openupgrade.py
pedrobaeza: It can't be because the `title` argument, that screw up the normal order of the arguments. I will comment it

openupgradelib/openupgrade.py
pedrobaeza: Done, but I don't understand the egg part

pedrobaeza:
Changes done

openupgradelib/openupgrade.py
bwrsandman: Could you elaborate?

Calling this works
```python
In [4]: def foo(a, i, t="", *args, **kw):
    print(a, i, t, args, kw)
   ...:     

In [5]: foo(1, 2, t=4, y=3)
1 2 4 () {'y': 3}
```

openupgradelib/openupgrade.py
bwrsandman: s/egg/python library/g

bwrsandman:
Unfortunately, progressbar fails all tests for python3 and greater.

It would be a shame to demote this library from python3 compatibility.

Maybe we could fix the incompatibility by getting the progressbar people to update the pypi.

openupgradelib/openupgrade.py
pedrobaeza: Yeah, but if you make:
```
foo(1, 2, 3, t=4, x=5)
```
you got:
```
TypeError: foo() got multiple values for keyword argument 't'
```
with the way I have put it, you should put:
```
foo(1, 2, t=4, args=[3], kwargs={'x': 5})
```

pedrobaeza:
Yeah, we can try to convince/help to bring support to Python3

openupgradelib/openupgrade.py
bwrsandman: Then you can put *args before t:

```python
In [8]: def foo(a, i, *args, t="", **kw):
    print(a, i, t, args, kw)
   ...:     

In [9]: foo(1, 2, 4, t=3, y=3)
1 2 3 (4,) {'y': 3}
```

openupgradelib/openupgrade.py
```diff
@@ -834,6 +837,48 @@ def reactivate_workflow_transitions(cr, transition_conditions):
...
+    """Log the progress of a method that it's run across the elements of an
+    iterator. The passed method should have the individual element to treat
+    as the first argument. It can have more arguments that are passed as
+    a list and a dictionary arguments for the args and kwargs respectively.
+    This is due to the extra title argument, that doesn't allow the
+    "addressing" of the arguments.
+    """
+    widgets = [title + ' ', progressbar.Percentage(), ' ', progressbar.Bar(),
+               ' ', progressbar.ETA()]
```
bwrsandman: Consider stripping the title of white space at least on the right side and also consider the case where title is blank where we would not want a space. 

Also, why is the blank space appended to title, when it could just be it's own element of the list. There could be cases where title has a string representation but cannot be appended to a str.

bwrsandman:
Please make your pull request towards the develop branch

HISTORY.rst
```diff
@@ -24,3 +24,11 @@ History
   * deferred80.py
   * openupgrade_log.py
   * #openupgrade_loading.py#
+
+0.1.3 (2015-08-11)
```
bwrsandman: Make this 1.1.0

setup.py
```diff
@@ -29,7 +29,7 @@
 
 setup(
     name='openupgradelib',
-    version='0.1.2',
+    version='0.1.3',
```
bwrsandman: 1.1.0

PR #5, opened 2015-08-12
 [IMP] set_defaults: Don't use ORM by default.

pedrobaeza:


pedrobaeza:
Now all it's fixed. As stated before, this has been already merged in https://github.com/OCA/OpenUpgrade/pull/340

bwrsandman:
:+1:

PR #6, opened 2015-08-13
Add basic coverage configuration

bwrsandman:


bwrsandman:
Ignore the fact that the coverage goes down. This is normal due to the files being analyzed changing.

openupgradelib/__init__.py
```diff
@@ -2,4 +2,7 @@
 
 __author__ = 'Odoo Community Association'
```
pedrobaeza: To be consistent, it should be `Odoo Community Association (OCA)`

PR #8, opened 2015-08-13
Add basic coverage configuration

bwrsandman:


bwrsandman:
There seems to be a bug with `coverage --include` and pypy3

pedrobaeza:
What can we do then?

bwrsandman:
We can wait for pypy3 or the coverage people to fix this bug. This isn't critical.

pedrobaeza:
But we will get a red cross in all PRs until then...

bwrsandman:
We wont. The only red is in this PR where we try to add --include.

The downside is that the coverage won't be right.

.coveragerc
```diff
@@ -0,0 +1,14 @@
+# .coveragerc to control coverage.py
+[run]
+branch = True
+include =
+    setup.py
+    openupgradelib/*
```
bwrsandman: Failing in pypy3 causing it to stall in `FnmatchMatcher.match`

pedrobaeza:
OK then :+1:

bwrsandman:
Related issues:
* https://bitbucket.org/ned/coveragepy/issues/347/coverage-pypy3-2to3-is-unacceptably-slow
* https://bitbucket.org/pypy/pypy/issues/1871/10x-slower-than-cpython-under-coverage

pedrobaeza:
Thanks for the info

bwrsandman:
I think I have worked around the issue.

bwrsandman:
\o/

Again ignore the decrease in coverage

pedrobaeza:
Cool, I see!

PR #11, opened 2015-08-24
[FIX] lib for working with old API (<= 7.0)

pedrobaeza:
Fix some imports for old versions of OpenUpgrade.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -26,11 +26,16 @@ import logging
 from openerp import release, tools, SUPERUSER_ID
 from openerp.tools.yaml_import import yaml_import
 from openerp.osv import orm
-from openerp.tools.mail import plaintext2html
 from openerp.modules.registry import RegistryManager
 import openerp.osv.fields
-import openerp.fields
 from . import openupgrade_tools
+if release.version_info[0] >= 7:
+    from openerp.tools.mail import plaintext2html
+if release.version_info[0] >= 8:
+    from openerp.fields import Many2many, One2many
+else:
+    Many2many = False
+    One2many = False
 
 if sys.version_info[0] == 3:
     unicode = str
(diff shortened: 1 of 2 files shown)


openupgradelib/openupgrade.py
```diff
@@ -26,11 +26,18 @@
...
 from . import openupgrade_tools
+try:
+    from openerp.tools.mail import plaintext2html
+except ImportError:
+    plaintext2html = False
+try:
+    from openerp.fields.Many2many import Many2many, One2many
+except:
+    Many2many = False
+    One2many = False
```
bwrsandman: Could you use `openerp.release.version`

pedrobaeza:
@bwrsandman, done!

openupgradelib/openupgrade.py
```diff
@@ -26,11 +26,16 @@
...
 from openerp.osv import orm
-from openerp.tools.mail import plaintext2html
 from openerp.modules.registry import RegistryManager
 import openerp.osv.fields
-import openerp.fields
 from . import openupgrade_tools
+if release >= "7.0":
+    from openerp.tools.mail import plaintext2html
+if release >= "8.0":
+    from openerp.fields.Many2many import Many2many, One2many
```
bwrsandman: I'm sorry to be picky, but isn't there a version_info in release that you could use? Like `release.version_info[0] == 7` or something like that?

pedrobaeza: Yeah, it is, but this string comparison is also good and without problems, so I prefer instead importing more variables (that are internals I think and can change)

bwrsandman: You're not importing other variables... `release` is a module which contains `version_info`. `release` is already imported

It would be incredibly irresponsible to change these internals since their purpose is to not change.

I'm not even convinced that this change works.

To make sure, could you change these separate if statements to be `if.. elif..else` where the else warns that no version was found.

pedrobaeza: They're 2 separates imports/logic: one is for a module/method only present for v7 and up, and the other is for new API declarations. It's working as I have tried them.

yvaucher:
@pedrobaeza You might want to re author your commit. author is 'Your Name'

pedrobaeza:
I have "re-branded" my commit and use version_info tuple instead of the release string. I hope this is enough.

pedrobaeza:
Anyone knows about the Travis error `TypeError: 'Mock' object does not support indexing`?

pedrobaeza:
OK, I get how to work with mocks. Now all is green and I've tried in several migrations and it's working correctly, so please merge to continue the path to migrate to openupgradelib.

StefanRijnhart:
Thanks! :+1:

StefanRijnhart:
Merging as @bwrsandman's comments have been honoured.

PR #13, opened 2015-09-11
[IMP] Include merging mechansim

kurkop:
With this improvement, you should be able to merge a module into an existing one. Using the `--init` flag trick you can do this out of the target module's migration scripts.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -94,6 +94,7 @@ __all__ = [
     'deactivate_workflow_transitions',
     'reactivate_workflow_transitions',
     'date_to_datetime_tz',
+    'merge_module_names',
 ]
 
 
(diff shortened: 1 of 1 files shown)


kurkop:
Flake seems to fail due to bad configuration of itself.

bwrsandman:
Could this functionality be put into update `update_module_names` instead?

kurkop:
@bwrsandman not, this is an example:

Two modules a,b should be migrated to c module, the module a can be updated to c with update_module_names, but b can't it because the module c would exist in tables, then updat_module_names can't update the name module in tables if it exists (with a). Merge allows do it.

openupgradelib/openupgrade.py
```diff
@@ -1071,3 +1072,55 @@ def date_to_datetime_tz(
                 AND rp.tz='%(timezone)s';
             """ % values)
     cr.execute("RESET TIMEZONE")
+
+
+def merge_module_names(cr, namespec):
+    """
+    Merge an source model into an existing target model.
```
pedrobaeza: s/an source/a source

pedrobaeza: s/model/module

pedrobaeza:
I see this a very special and rare case, but well, why not? Please check Travis errors.

kurkop:
Ok @pedrobaeza , was missing a newline in file to paseed travis error.

openupgradelib/openupgrade.py
```diff
@@ -1071,3 +1072,55 @@ def date_to_datetime_tz(
...
+
+        # Delete meta entries, that might conflict.
+        # They will be recreated by the new module anyhow.
+        query = ("DELETE FROM ir_model_constraint "
+                 "WHERE module = %s")
+        logged_query(cr, query, [id_old])
+        query = ("DELETE FROM ir_model_relation "
+                 "WHERE module = %s")
+        logged_query(cr, query, [id_old])
+        query = ("DELETE FROM ir_translation "
```
pedrobaeza: You can rename translations instead of removing.

openupgradelib/openupgrade.py
```diff
@@ -1071,3 +1072,55 @@ def date_to_datetime_tz(
...
+                 "WHERE module = %s")
```
pedrobaeza: As this library is for all versions, you have to take care the release version, because module field was added on version 7, so as this is executed in pre-init scripts, you should put `if release.version_info[0] > 7:`

openupgradelib/openupgrade.py
```diff
@@ -1071,3 +1072,55 @@ def date_to_datetime_tz(
...
+                 "(SELECT name FROM ir_model_data "
+                 "WHERE module = %s)")
+        logged_query(cr, query, [new_name, old_name, new_name])
+        query = ("DELETE FROM ir_model_data "
+                 "WHERE module = %s AND name IN "
+                 "(SELECT name FROM ir_model_data "
+                 "WHERE module = %s)")
+        logged_query(cr, query, [old_name, new_name])
+
+        # Update Dependencies and Errase traces of old module.
```
pedrobaeza: s/Errase/erase

openupgradelib/openupgrade.py
kurkop: @pedrobaeza you need that add "if" to all method, because all query has "module" field. In other words:
if not release.version_info[0] > 7:
    return
(Before "for")

pedrobaeza: This is only the line that diverges from versions, so you only need to check here, and make the rest of steps on versions prior to 7

kurkop:
@pedrobaeza I update PR and rebase with master.

pedrobaeza:
There's a conflict with your branch. Please rebase and resolve the conflict.

pedrobaeza:
I have relaunched your idea with a new implementation in https://github.com/OCA/openupgradelib/pull/63. Please check @kurkop @blaggacao

PR #15, opened 2015-09-11
[IMP] docstrings `map_values`

blaggacao:
suggest usage in post-migration.
please merge, if true, please close, if not.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -703,6 +703,8 @@ def map_values(
     """
     Map old values to new values within the same model or table. Old values
     presumably come from a legacy column.
+    Unless special cases, you want to typically use it 
+    in the post-migrations scripts.
 
     :param cr: The database cursor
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -703,6 +703,8 @@ def map_values(
     """
     Map old values to new values within the same model or table. Old values
     presumably come from a legacy column.
+    Unless special cases, you want to typically use it
```
StefanRijnhart: 'You will typically want to use it in post-migration scripts.'

pedrobaeza:
:+1:

StefanRijnhart:
Thanks for the change. If you silently update the code to honour the comments, the reviewers are not notified so next time, please add a comment as well saying that you did so.
:+1:

blaggacao:
OK, cool. I wasn't aware of that github's behaviour. Thanks and Best, David

PR #18, opened 2015-09-11
[IMP] (map_values): Support set & notset selectors

kurkop:
`True` as `old value` represents "is set", whereas `False` represets "is not set"

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -703,6 +703,7 @@ def map_values(
     """
     Map old values to new values within the same model or table. Old values
     presumably come from a legacy column.
+    Old value True represents "is set", False "is not set".
 
     :param cr: The database cursor
     :param source_column: the database column that contains old values to be \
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -734,20 +735,35 @@ def map_values(
             " columns : %s",
             source_column)
     for old, new in mapping:
+
+        new = "'%s'"% new
+
+        if old == True:
```
pedrobaeza: You must use `if old is True`

openupgradelib/openupgrade.py
```diff
@@ -734,20 +735,35 @@ def map_values(
...
+            old = 'NOT NULL'
+            op = 'IS'
+        elif old == False:
```
pedrobaeza: Use `elif old is False:`

openupgradelib/openupgrade.py
```diff
@@ -734,20 +735,35 @@ def map_values(
...
+            old = 'NULL'
+            op = 'IS'
+        else:
+            old = "'%s'"% old
```
pedrobaeza: put an space before `%`

kurkop: @pedrobaeza before which %?

pedrobaeza: `old = "'%s'" % old` (for avoiding lint errors)

openupgradelib/openupgrade.py
```diff
@@ -734,20 +735,35 @@ def map_values(
             " columns : %s",
             source_column)
     for old, new in mapping:
+
+        new = "'%s'"% new
```
pedrobaeza: I forgot to highlight the same problem here:
```
openupgradelib/openupgrade.py:739:21: E225 missing whitespace around operator
```

pedrobaeza: And please make only 1 commit for cleanliness

kurkop: Ok, I made the changes.

pedrobaeza:
:+1:

StefanRijnhart:
Please rewrite the commit's message, which now includes Git's hints on rebasing. Next time, it is fine to leave the commits as they are instead of rebasing your proposal into a single commit after review.

pedrobaeza:
I was the one who ask him to rebase to only 1 commit, because we put a lot of commits for simple patches that makes the commit history very awful. Putting correctly the commit message is enough for ensuring the correctness of the full PR.

StefanRijnhart:
Alright, a matter of taste I guess.

pedrobaeza:
@StefanRijnhart, if code is OK for you, I can rewrite commit message when merging

StefanRijnhart:
@pedrobaeza Deal! :+1:

PR #19, opened 2015-09-18
[FIX] set_defaults: Cope with inherited fields by delegation

pedrobaeza:
Making the work to port @hbrunn's patch for the problem with inherited fields by delegation here to the lib. I'm going to check it right now also on a migration and let you know the result.


StefanRijnhart:
@pedrobaeza did you test it succesfully?

pedrobaeza:
Yeah, I completed sucessfully a migration with project module and I didn't find any error, and default for delegated field "Tasks" was set to True in existing projects, so :+1: by my part.

StefanRijnhart:
Thanks! As the code is verbatim @hbrunn's change to the same code when it was still integrated in the original project I'm counting enough support for a merge.

PR #21, opened 2015-09-26
[FIX] Compatibility for OpenERP versions prior to 6.1

pedrobaeza:
Closes https://github.com/OCA/OpenUpgrade/issues/373

@sebalix I'm afraid you should check if all used methods are totally compatible with these versions, as the used version for all of them has been 8.0, but with this patch you will be able to load OpenUpgrade 6.0 branch (and I have also checked 5.0).

cc @StefanRijnhart

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -23,22 +23,39 @@ import sys
 import os
 import inspect
 import logging
-from openerp import release, tools, SUPERUSER_ID
-from openerp.tools.yaml_import import yaml_import
-from openerp.osv import orm
-from openerp.modules.registry import RegistryManager
-import openerp.osv.fields
 from . import openupgrade_tools
-if release.version_info[0] >= 7:
-    from openerp.tools.mail import plaintext2html
-if release.version_info[0] >= 8:
-    from openerp.fields import Many2many, One2many
+try:
+    from openerp import release
+except ImportError:
+    import release
+Many2many = False
+One2many = False
+if not hasattr(release, 'version_info'):
+    version_info = tuple(map(int, release.version.split('.')))
 else:
-    Many2many = False
-    One2many = False
+    version_info = release.version_info
+if version_info[0] > 6 or version_info[0:2] == (6, 1):
+    from openerp import tools, SUPERUSER_ID
+    from openerp.tools.yaml_import import yaml_import
+    from openerp.osv.orm import except_orm
+    from openerp.modules.registry import RegistryManager
+    from openerp.osv.fields import many2many, one2many
+    if version_info[0] >= 7:
+        from openerp.tools.mail import plaintext2html
+    if version_info[0] >= 8:
+        from openerp.fields import Many2many, One2many
+else:
+    # version < 6.1
+    import tools
+    SUPERUSER_ID = 1
+    from tools.yaml_import import yaml_import
+    from osv.osv import except_osv as except_orm
+    RegistryManager = None
+    from osv.fields import many2many, one2many
 
 if sys.version_info[0] == 3:
     unicode = str
+
 
 # The server log level has not been set at this point
 # so to log at loglevel debug we need to set it
@@ -499,7 +516,7 @@ def set_defaults(cr, pool, default_spec, force=False, use_orm=False):
     for model in default_spec.keys():
         obj = pool.get(model)
         if not obj:
-            raise orm.except_orm(
+            raise except_orm(
                 "Error",
                 "Migration: error setting default, no such model: %s" % model)
         for field, value in default_spec[model]:
@@ -536,7 +553,7 @@ def set_defaults(cr, pool, default_spec, force=False, use_orm=False):
                             field, model))
                     logger.error(error)
                     # this exception seems to get lost in a higher up try block
-                    orm.except_orm("OpenUpgrade", error)
+                    except_orm("OpenUpgrade", error)
             else:
                 write_value(ids, field, value)
 
@@ -593,7 +610,7 @@ def update_module_names(cr, namespec):
         query = ("UPDATE ir_module_module_dependency SET name = %s "
                  "WHERE name = %s")
         logged_query(cr, query, (new_name, old_name))
-        if release.version_info[0] > 7:
+        if version_info[0] > 7:
             query = ("UPDATE ir_translation SET module = %s "
                      "WHERE module = %s")
             logged_query(cr, query, (new_name, old_name))
@@ -625,8 +642,8 @@ def get_legacy_name(original_name):
     :param original_name: the original name of the column
     :param version: current version as passed to migrate()
     """
-    return 'openupgrade_legacy_'+('_').join(
-        map(str, release.version_info[0:2]))+'_'+original_name
+    return 'openupgrade_legacy_' + '_'.join(
+        map(str, version_info[0:2])) + '_' + original_name
 
 
 def m2o_to_x2m(cr, model, table, field, source_field):
@@ -648,13 +665,15 @@ def m2o_to_x2m(cr, model, table, field, source_field):
     .. versionadded:: 8.0
     """
     if not model._columns.get(field):
-        raise orm.except_orm(
+        raise except_orm(
             "Error", "m2o_to_x2m: field %s doesn't exist in model %s" % (
                 field, model._name))
-    if isinstance(model._columns[field], (openerp.osv.fields.many2many,
-                                          Many2many)):
-        rel, id1, id2 = openerp.osv.fields.many2many._sql_names(
-            model._columns[field], model)
+    if isinstance(model._columns[field], (many2many, Many2many)):
+        column = model._columns[field]
+        if version_info[0] > 6 or version_info[0:2] == (6, 1):
+            rel, id1, id2 = many2many._sql_names(column, model)
+        else:
+            rel, id1, id2 = column._rel, column._id1, column._id2
         logged_query(
             cr,
             """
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -648,13 +660,11 @@ def m2o_to_x2m(cr, model, table, field, source_field):
...
-        raise orm.except_orm(
+        raise except_orm(
             "Error", "m2o_to_x2m: field %s doesn't exist in model %s" % (
                 field, model._name))
-    if isinstance(model._columns[field], (openerp.osv.fields.many2many,
-                                          Many2many)):
-        rel, id1, id2 = openerp.osv.fields.many2many._sql_names(
-            model._columns[field], model)
+    if isinstance(model._columns[field], (many2many, Many2many)):
+        rel, id1, id2 = many2many._sql_names(model._columns[field], model)
```
StefanRijnhart: I think you'll want to take self._rel, self._id1, self._id2 for version < 6.1, because the _sql_names method was introduced then.

StefanRijnhart:
Thanks for dealing with this one!

openupgradelib/openupgrade.py
pedrobaeza: Done!

sebalix:
@pedrobaeza I will test that, thank you

openupgradelib/openupgrade.py
StefanRijnhart: Sorry for my confusing use of *self* in the initial comment. You take the fields from *model*, but should that not be the column itself?

pedrobaeza: My bad also for not thinking twice (it's sunday :stuck_out_tongue_winking_eye:). Fixed.

StefanRijnhart:
LGTM ;-) :+1:

sebalix:
Hi @pedrobaeza 

I just tested your branch, and I have a new error now:

```python
ERROR:OpenUpgrade:base: error in migration script /home/openerp/oerp/server/bin/addons/base/migrations/6.0.1.3/pre-migration.py: 'module' object has no attribute 'version_info'
[2015-09-28 16:44:25,664][database] ERROR:OpenUpgrade:'module' object has no attribute 'version_info'
  File "/usr/local/lib/python2.7/dist-packages/openupgradelib-1.1.0-py2.7.egg/openupgradelib/openupgrade.py", line 926, in wrapped_function
    func(cr, version)
  File "base/migrations/6.0.1.3/pre-migration.py", line 108, in migrate
    cr, module_namespec
  File "/usr/local/lib/python2.7/dist-packages/openupgradelib-1.1.0-py2.7.egg/openupgradelib/openupgrade.py", line 608, in update_module_names
    if release.version_info[0] > 7:
AttributeError: 'module' object has no attribute 'version_info'
[2015-09-28 16:44:25,665][database] ERROR:migration:module base: Each pre-migration file must have a "migrate(cr, installed_version)" function
```

The `release` module does not contain the `version_info` variable in the 6.0 branch, just this:
```python
version = '6.0.4'
major_version = '6.0'
```

And the `release.version_info` pattern is used at differents places in `openupgradelib`:
```sh
./tests/test_openupgradelib.py:19:openerp_mock.release.version_info = (8, 0, 0, 'final', 0)
./openupgradelib/openupgrade.py:39:    if release.version_info[0] >= 7:
./openupgradelib/openupgrade.py:41:    if release.version_info[0] >= 8:
./openupgradelib/openupgrade.py:608:        if release.version_info[0] > 7:
./openupgradelib/openupgrade.py:641:        map(str, release.version_info[0:2]))+'_'+original_name
./openupgradelib/openupgrade.py:1065:    if release.version_info[0] < 7:
```

Maybe we should encapsulate the detection of the version in a new function?

pedrobaeza:
Yeah, I missed that one. I already did it in this one: https://github.com/OCA/openupgradelib/pull/21/files#diff-3d7675eb1386de028dab1f879bdc4681R33, but there are some other places where vendor_info is used.

The problem I see encapsulating the version detection is that each new Odoo version will need to add some logic to this encapsulation method, but if you see this correct, we can make it via CONSTANTS with growing integers that allows to make comparisons with <, =, >, <=, and we can fallback to a higher value in case none of the current versions are present. What do you think?
```
VERSION_42 = 1
VERSION_50 = 2
VERSION_60 = 3
VERSION_61 = 4
VERSION_70 = 5
VERSION_80 = 6
VERSION_90 = 7
VERSION_UNKNOWN = 99
...
if not hasattr(release, 'version_info'):
    if x:
        version = VERSION_60
   else:
...
```

sebalix:
The problem this with system is the `get_legacy_name` function which needs the real version number to generate the field's name, no?

So I was thinking about a function like this:
```python
def get_version():
    if not hasattr(release, 'version_info'):
        return tuple(map(int, release.version.split('.')))    # add ('final, 0) if it really matters
    else:
        return release.version_info()
```
And replacing `release.version_info` everywhere by a call to `get_version()`. What do you think?

pedrobaeza:
Good idea! I think this can be good enough to overpass this problem, although I think defining a method prior imports is not allowed by flake8.

sebalix:
If flake8 complained about this, we could put the function in another (new) module?

pedrobaeza:
I'll check tonight and tell you

pedrobaeza:
OK, let's hope this one is the last needed modification! I have saved in version_info module variable the tuple (previously obtained according version), and thus the conditional can be the same in all places.

cc @StefanRijnhart @sebalix

sebalix:
Hi @pedrobaeza 
Now I can launch the 6.0 server, great.

I have another error which occurs later but I don't know if its related to `openupgradelib` or the `base` module migration script of the 6.0 branch or the database itself. All I know is that the migration 5.0 -> 6.0 was successful before the integration of the `openupgradelib`, so maybe that some errors was not raised before or something like that. If it can help, the traceback:

```python
DEBUG:OpenUpgrade:model res.users, field company_ids: setting default value of resources [4, 3, 2, 1] to [1]
[2015-09-29 12:30:09,176][database] ERROR:db.cursor:Programming error: column "company_ids" of relation "res_users" does not exist
LINE 1: UPDATE res_users SET company_ids = ARRAY[1] WHERE id IN (1, ...
                             ^
, in query UPDATE res_users SET company_ids = %s WHERE id IN %s
[2015-09-29 12:30:09,176][database] ERROR:OpenUpgrade:base: error in migration script /home/openerp/oerp/server/bin/addons/base/migrations/6.0.1.3/post-migration.py: column "company_ids" of relation "res_users" does not exist
LINE 1: UPDATE res_users SET company_ids = ARRAY[1] WHERE id IN (1, ...
                             ^

[2015-09-29 12:30:09,176][database] ERROR:OpenUpgrade:column "company_ids" of relation "res_users" does not exist
LINE 1: UPDATE res_users SET company_ids = ARRAY[1] WHERE id IN (1, ...
                             ^
  File "/usr/local/lib/python2.7/dist-packages/openupgradelib-1.1.0-py2.7.egg/openupgradelib/openupgrade.py", line 930, in wrapped_function
    func(cr, version)
  File "base/migrations/6.0.1.3/post-migration.py", line 305, in migrate
    openupgrade.set_defaults(cr, pool, defaults)
  File "/usr/local/lib/python2.7/dist-packages/openupgradelib-1.1.0-py2.7.egg/openupgradelib/openupgrade.py", line 548, in set_defaults
    obj._defaults[field](obj, cr, user_id, None))
  File "/usr/local/lib/python2.7/dist-packages/openupgradelib-1.1.0-py2.7.egg/openupgradelib/openupgrade.py", line 514, in write_value
  File "/home/openerp/oerp/server/bin/sql_db.py", line 78, in wrapper
  File "/home/openerp/oerp/server/bin/sql_db.py", line 131, in execute
ProgrammingError: column "company_ids" of relation "res_users" does not exist
LINE 1: UPDATE res_users SET company_ids = ARRAY[1] WHERE id IN (1, ...
                             ^

  File "/home/openerp/oerp/server/bin/openerp-server.py", line 121, in <module>
    db,pool = pooler.get_db_and_pool(dbname, update_module=tools.config['init'] or tools.config['update'], pooljobs=False)
  File "/home/openerp/oerp/server/bin/pooler.py", line 39, in get_db_and_pool
    addons.load_modules(db, force_demo, status, update_module)
  File "/home/openerp/oerp/server/bin/addons/__init__.py", line 975, in load_modules
    processed_modules.extend(load_module_graph(cr, graph, status, perform_checks=(not update_module), registry=registry, report=report, skip_modules=skip_modules))
  File "/home/openerp/oerp/server/bin/addons/__init__.py", line 903, in load_module_graph
    migrations.migrate_module(package, 'post')
  File "/home/openerp/oerp/server/bin/addons/__init__.py", line 598, in migrate_module
    mod.migrate(self.cr, pkg.installed_version)
  File "/usr/local/lib/python2.7/dist-packages/openupgradelib-1.1.0-py2.7.egg/openupgradelib/openupgrade.py", line 930, in wrapped_function
    func(cr, version)
  File "base/migrations/6.0.1.3/post-migration.py", line 305, in migrate
  File "/usr/local/lib/python2.7/dist-packages/openupgradelib-1.1.0-py2.7.egg/openupgradelib/openupgrade.py", line 548, in set_defaults
    obj._defaults[field](obj, cr, user_id, None))
  File "/usr/local/lib/python2.7/dist-packages/openupgradelib-1.1.0-py2.7.egg/openupgradelib/openupgrade.py", line 514, in write_value
  File "/home/openerp/oerp/server/bin/sql_db.py", line 78, in wrapper
  File "/home/openerp/oerp/server/bin/sql_db.py", line 131, in execute
psycopg2.ProgrammingError: column "company_ids" of relation "res_users" does not exist
LINE 1: UPDATE res_users SET company_ids = ARRAY[1] WHERE id IN (1, ...
```

StefanRijnhart:
That is a regression of changing the default method of set_defaults to SQL, which does not handle x2many fields properly. It could be improved on that point, or you could call set_defaults for these fields separately with an additional use_orm=True argument.

pedrobaeza:
As said by @StefanRijnhart, it's not something of this PR. We can go on then to fix this and then make another PR for fixing the other problem.

sebalix:
:+1:

pedrobaeza:
@StefanRijnhart, can you please merge?

StefanRijnhart:
With pleasure!

PR #27, opened 2016-02-16
Remove wrong docs

Yajo:
Looking at the code, it seems like it gets quoted automatically.


Yajo:
Closing to reopen from right repo.

PR #28, opened 2016-02-16
Avoid broken updates

Yajo:
If a migration fails, it will get reverted.


Yajo:
Sorry, reopening from right repo.

PR #29, opened 2016-02-16
Add new context manager allow_pgcodes.

Yajo:
It allows the migrator to ignore some expected error codes from Postgres.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -108,6 +108,18 @@ def allow_pgcodes(cr, *codes):
         with allow_pgcodes(cr, psycopg2.errorcodes.UNIQUE_VIOLATION):
             cr.execute("INSERT INTO me (name) SELECT name FROM you")
 
+    .. warning::
+        **All** sentences inside this context will be rolled back if **a single
+        error** is raised, so the above example would insert **nothing** if a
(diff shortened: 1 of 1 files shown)


Yajo:
Please some help with travis? Seems like pypy versions cannot compile psycopg2, but it makes no sense to me because without that you simply cannot migrate...

pedrobaeza:
Well, Travis don't perform a migration for now, so it doesn't need it. It only makes a flake8/pylint check. It seems that now that you have introduced it, it raises an error. You can enclose it in a `try: except ImportError:` clause to avoid it.

Yajo:
I deferred the import and enabled the pypy alternative for it, travis is :green_heart: now.

StefanRijnhart:
Nice one :+1:

pedrobaeza:
Using this feature, the operation enclosed the context manager won't be performed, isn't it? If so, I'm not sure that is a good idea to silently ignore the problem. If don't, then the DB will get an inconsistent state. I don't see this change as something good for the health of the script migrations.

StefanRijnhart:
Yes, that point could be made more clear (in the docstring and maybe with loglevel info instead of debug).

Yajo:
You're right, better now?

pedrobaeza:
:+1:

PR #30, opened 2016-02-16
Avoid broken updates

Yajo:
Reopening #28. If a migration fails, it will get reverted.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -945,7 +945,8 @@ def migrate(no_version=False):
                 (module, stage, version))
             try:
                 # The actual function is called here
-                func(cr, version)
+                with cr.savepoint():
+                    func(cr, version)
             except Exception as e:
                 logger.error(
                     "%s: error in migration script %s: %s" %


pedrobaeza:
Shouldn't you need to restore the save point in case of error?

Yajo:
That's what the context manager does

pedrobaeza:
OK, I'm not very used to the internal upgrade mechanism. I let @hbrunn or @StefanRijnhart to judge.

StefanRijnhart:
We tried to make sure that the upgrade of every individual module is run in its own transaction, because of the interaction between pre and post migration scripts. If that is broken somehow, it should be fixed in the core files of the OpenUpgrade distribution.

With this change, you run the risk of having a successful pre-migration and module loading, but a broken post-migration. If only the post-migration is reverted, you are probably stuck with a broken database unless the pre-migration happens to be idempotent (most of them are not).

So if the problem is there, it should be fixed but IMHO not in this particular bit of code.

hbrunn:
agreed with @StefanRijnhart

Yajo:
Actually I need to perform some migrations that fall outside the scope of the OpenUpgrade project, so the idea behind this is to be able to make this library more reusable for those scenarios.

When writing a pre script, I found that if the whole script is not bundled inside a transaction, you would end up having to restore the database to test the script again, which makes the developer lose a lot of time.

Please don't confuse the `with` with a `try`. If an exception happens, it will be effectively raised. The only difference is that all sentences inside the block will be discarded.

According to my tests, when that exception is raised, it makes Odoo retry the migration in an endless loop. Without this fix, each try will yield a different failure because the transaction is not rolled back; with it, all tries will yield the same error. This allows you to just fix the script and reboot Odoo to make it continue.

Besides, AFAIK savepoints can be found inside other savepoints. If you say OpenUpgrade uses one transaction for the whole pre+post process in OpenUpgrade, then in the worst case this patch should not hurt, don't you think?

StefanRijnhart:
Sorry I misunderstood. The change I spoke of was in OpenUpgrade core distribution. I now understand that this change is very useful for migrations outside of this context. :+1: and thanks!

pedrobaeza:
@Yajo, now I understand what you said about the context manager.

:+1: for me too

hbrunn:
this fails miserably for OpenERP <= 7.0. I'll post a patch soon

PR #32, opened 2016-02-24
[WIP] New object-oriented migrator tools collection.

Yajo:
Beware: Methods decorated with `@wip` are just a concept proof! The other ones are the ones I actually tested. I'll keep removing that decorator as I keep doing migrations and testing those.

This is a proof of concept of some work that I already had done before I realized openupgradelib existed. However I think it can be useful/interesting to others, so I'm sharing it here (maybe you'll see it and think it's absolutely unneeded, but anyway I'm interested in your opinions :wink:).

It works by creating an instance of the `Migrator` class that represents an addon being migrated, and allows partial access to an environment, from where it can read some info. However, writes are done directly by DB cursor to avoid breakages.

You can see this [example migration](https://github.com/OCA/social/pull/45/files#diff-3f103bf6a08dcc9eb06d977073d13472) to get the point of how migrations would look with this.

FTR this contiunes https://github.com/OCA/server-tools/pull/369.

### DIFF
--- a/openupgradelib/migrator.py
+++ b/openupgradelib/migrator.py
@@ -1,0 +1,431 @@
+# -*- coding: utf-8 -*-
+# © 2016 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+from contextlib import contextmanager
+from functools import wraps
+from psycopg2 import errorcodes, ProgrammingError
+from openerp import api, SUPERUSER_ID
+
+_logger = logging.getLogger(__name__)
+
+
+# TODO Remove when module ends WIP stage
+def wip(function):
+    """This indicates methods that are currently WIP."""
+    @wraps(function)
+    def wrapper(*args, **kwargs):
+        _logger.warning(
+            "BEWARE! You are using an untested function: %s. If you find it "
+            "works fine, please remove the `@wip` decorator.",
+            function.__name__)
+        return function(*args, **kwargs)
+    return wrapper
+
+
+class MigrationException(Exception):
+    """Custom exception raised only by the migration toolkit.
+
+    Odoo handles some exceptions like :class:`AttributeError` and
+    :class:`psycopg2.OperationalError` in some places of the upper stack. By
+    defining a custom class for this toolkit's exceptions, you make Odoo not
+    handle them and abort the migration process with a useful error message.
+    """
+    def __init__(self, old):
+        """Create exception and link it to the older one.
+
+        :param Exception old:
+            Old exception that was caught before this one.
+        """
+        _logger.exception(old)
+        self.old = old
+        super(MigrationException, self).__init__(*old.args)
+
+
+def manage(function):
+    """Run sentences inside a transaction and environment.
+
+    If anything fails inside :param:`function`, all will be rolled back; and
+    you have access to an Odoo environment like when using models normally.
+
+    This decorator is prepared to be applied to your ``migrate`` method (the
+    required method for all migrations), but if you want to use it in another
+    context, at least methods decorated with this must have ``cr`` as their
+    first parameter.
+    """
+    @wraps(function)
+    def wrapper(cr, *args, **kwargs):
+        try:
+            with cr.savepoint():
+                with api.Environment.manage():
+                    return function(cr, *args, **kwargs)
+        except Exception as error:
+            raise MigrationException(error)
+    return wrapper
+
+
+class Migrator(object):
+    def __init__(self, cr, addon, table_names=None, context=None):
+        """Start a migrator instance for a given addon.
+
+        :param openerp.sql_db.Cursor cr:
+            Database cursor to use for the migration.
+
+        :param str addon:
+            This is the name of the addon where data is attached before running
+            the migration.
+
+        :param dict table_names:
+            Dictionary mapping model names to table names, such as::
+
+                {"res.partner": "res_partner"}
+
+            Only useful when a model's table **is not** the model's name
+            replacing dots by low lines **and** the model name cannot be found
+            in Odoo's regitry anymore (maybe because you are deleting it).
+
+        :param dict context:
+            Default context for the environment created, in case you need it.
+        """
+        self.env = api.Environment(cr, SUPERUSER_ID, context or dict())
+        self.addon = addon
+        self.table_names = table_names or None
+
+    @wip
+    @contextmanager
+    def _allow_pgcodes(self, *codes):
+        """Context manager that will omit specified error codes.
+
+        :param *str codes:
+            Undefined amount of error codes found in :mod:`psycopg2.errorcodes`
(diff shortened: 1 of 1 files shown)


openupgradelib/migrator.py
```diff
@@ -0,0 +1,549 @@
...
+        self._execute("DELETE FROM %s WHERE id IN %%s" % table, (ids,))
+        if xmlid:
+            self._execute(
+                """ DELETE FROM ir_model_data
+                    WHERE module = %s AND model = %s AND res_id IN %s
+                """,
+                (self.addon, model_name, ids))
+
+    @wip
+    def _table_fk(self, table_name):
```
blaggacao: I would very much like to see this deeply integrated into openupgradelib. For now it is quite limited for to 2many fields :/ @Yajo @StefanRijnhart 

We used the mentioned DB query quite extensively in the past.

StefanRijnhart: :+1: for adding this to openupgradelib, although the model data table can be cleaned with the cleanup_database module. For mapping lists, does rename_xmlids not work for you?

@yajo You'll probably want to drop the selection by module. Any references to the record in ir_model_data can be purged safely.

Yajo: But you mean you want the whole migrator or just this method in a standalone function?

blaggacao: @StefanRijnhart My comment was actually meant for the `def _table_fk(self, table_name):` function. Sorry, github doesn't display as I intended. 

@Yajo I think the your migrator can be analyzed one by one as I might guess there are overlapping features. I was pleased to see `_table_fk(self, table_name)` and it would be nice if a `copy_table` or `rename_table` would in some distant future be aware of 2many relational tables.

openupgradelib/migrator.py
```diff
@@ -0,0 +1,549 @@
...
+
+                    JOIN information_schema.constraint_column_usage AS ccu
+                    ON tc.constraint_name = ccu.constraint_name
+                WHERE
+                    kcu.constraint_type = 'FOREIGN KEY' AND
+                    ccu.table_name = %s
+            """,
+            (table_name,)
+        )
+        return self.env.cr.dictfetchall()
```
blaggacao: I actually meant this query. Properly packaged this could enable a whole new class of openupgradelib functionality.

blaggacao: I think @StefanRijnhart can best estimate where such a method would be useful. Integrating it would then be rather easy I guess. The difficulty here probably was the query construction itself.

Yajo: Feel free to take it and add your own PR for that please :wink:. This method is mostly untested and I have no plans on testing or updating this soon

StefanRijnhart: I think it would be fine to have this in OpenUpgrade, but as always I won't be proposing until I need it.

blaggacao: I've a few migrations touching this topic on my todo list, so I'll see what I can do.

blaggacao:
@Yajo I got interested in this. Could you state the main pros and cons of your approach. I think this helps understanding and maybe propel integrating efforts. Not sure.

openupgradelib/migrator.py
```diff
@@ -0,0 +1,549 @@
...
+    This decorator is prepared to be applied to your ``migrate`` method (the
+    required method for all migrations), but if you want to use it in another
+    context, at least methods decorated with this must have ``cr`` as their
+    first parameter.
+    """
+    @wraps(function)
+    def wrapper(cr, *args, **kwargs):
+        try:
+            with cr.savepoint():
+                with api.Environment.manage():
```
blaggacao: @Yajo How would usage of this look like? How would I include that into the current migrate decorator? I think this would be rather easy to include, wouldn't it?

blaggacao: From the docs I see:

```
 For instance, the statements::

        model = self.pool.get(MODEL)
        ids = model.search(cr, uid, DOMAIN, context=context)
        for rec in model.browse(cr, uid, ids, context=context):
            print rec.name
        model.write(cr, uid, ids, VALUES, context=context)

    may also be written as::

        env = Environment(cr, uid, context) # cr, uid, context wrapped in env
        model = env[MODEL]                  # retrieve an instance of MODEL
        recs = model.search(DOMAIN)         # search returns a recordset
        for rec in recs:                    # iterate over the records
            print rec.name
        recs.write(VALUES)                  # update all records in recs
```

So the idea is that any model called with
Env['modelname'] has cr,uid,context included and we would not need to care about this for the rest of the script? Am I correct in assuming that this would mean the whole openupgradelib could be slowly migrated to new api style, which obviously would make things easier for newcomers who only grew up with the new api?

blaggacao:
I'm at the moment going through the ORM and I wonder, if you're alternative approach probably is more contemporary... While I do not see the immediate benefits of working in an oo rather than procedural way, (maybe you can point me out the pros), I do see benefits in doing migration tasks through the ORM rather than at the low DB level.

Now, the large code base of openupgradelib is rather DB centric, than working through the ORM, when we want to benefit from some more advanced ORM-like functionality, this means basically rebuilding it, which might not be the idea. What I'm wondering is, if this is for legacy reasons (because in version 6 you just had to do it that way) or if basically any migration through SQL could also be theoretically be done through a consistent wrapping of ORM methods?

openupgradelib/migrator.py
Yajo: This method's idea got into https://github.com/OCA/openupgradelib/pull/30, I think it should be deleted from here.

Yajo:
That was my point since the begining. We have an ORM, why not use it for migrations?

Well, that and not knowing openupgradelib at that time... :sweat_smile: 

I don't have any pending migrations to test this new migrator tool, so it got stuck at WIP stage. But I remember it as much more pleasant to work with than current procedural-oriented openupgradelib. Maybe you want to test it in your next migrations and send me PR until we are able to remove all `@wip` decorators.

BTW, those methods *not* decorated with that are supposed to be stable.

StefanRijnhart:
Yes, in theory the ORM would give the most consistent results. In practice, it is often too slow for real world databases and it's benefits (recomputation of stored function fields, firing of workflow triggers) are often redundant or again performance wise disastrous. I will still happily write post migration scripts using the ORM, but quite a lot of them have been patched in the meantime to either be run directly in the database or have field recomputation or workflow triggers disabled when they run.

blaggacao:
@StefanRijnhart Good explanation, I think many inclined people might ask themselves exactly this, just as @Yajo and I did. I suspected a reason, without knowing it. Maybe this can make it into the documentation? That would be awesome...

As a conclusion, I think openupgradelib could probably embrace a sql-orm dichotomy? Where sql methods can be used for speed, exactly knowing what one is doing, and orm methods, which provides more consistent results and expanded use cases (m2m fields) at the cost of speed.

pedrobaeza:
I'm closing this as there explained reasons and other techniques (like use_env parameter in migrate) for having ORM in migration scripts.

PR #33, opened 2016-04-02
[FIX] support OpenERP versions that don't have cr.savepoint

hbrunn:
which is <= 7.0. The problematic change was introduced by #30


pedrobaeza:
Thanks for this :+1:

sebalix:
:+1:

StefanRijnhart:
:+1:, thanks!

PR #35, opened 2016-06-17
[IMP] Add xml mapping functionality

blaggacao:
- maps a fully qualified xml dict to it's database ids
- migrates the values via map_value
- includes an optional cke to verify if all source column values have a mapping
- otherwise ignores faulty tuples of the mapping resolution
- but has a magical external id named 'NULL' which matches a database NULL in source and destination column
- as per map_value behaviour: does not verify if there are multiple conflicting mappings (one old with several new)

Please give feedback. Especially, I need to know how, where to include the imports and the imd in order to make it openupgradelib-ish. Also feedback in terms of functionality (completeness and scope) would be nice.

I think the use case is rather obvious, just ping me for any doubts.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -799,13 +799,12 @@ def float_to_integer(cr, table, field):
 
 
 def map_values(
-        cr, source_column, target_column, mapping,
-        model=None, table=None, write='sql'):
+        cr, source_column, target_column, mapping, xml_mapping=False,
+        model=None, table=None, write='sql', check_completeness=False):
     """
     Map old values to new values within the same model or table. Old values
     presumably come from a legacy column.
     You will typically want to use it in post-migration scripts.
-
     :param cr: The database cursor
     :param source_column: the database column that contains old values to be \
     mapped
@@ -813,15 +812,47 @@ def map_values(
     'orm') that the new values are written to
     :param mapping: list of tuples [(old value, new value)]
         Old value True represents "is set", False "is not set".
+    :param xml_mapping: If True, mapping is expected to be a fully qualified \
+    external ID as for example in "account.1_account_tax_1234".
     :param model: used for writing if 'write' is 'orm', or to retrieve the \
     table if 'table' is not given.
     :param table: the database table used to query the old values, and write \
     the new values (if 'write' is 'sql')
     :param write: Either 'orm' or 'sql'. Note that old ids are always \
     identified by an sql read.
+    :param check_completeness: True verifies, that all source entries \
+    have been accounted for in the mapping table and raises otherwise.
 
+    This moethod does not support mapping values across tables, neither \
+    relational nor property fields (it does not support resiliant \
+    handling of their respective relational tables - you still can use \
+    this method on those relational tables in sql mode).
     .. versionadded:: 8.0
     """
+    def use_xml_mapping(cr, mapping):
+        from openerp.modules.registry import RegistryManager
+        gid = RegistryManager.get(cr.dbname)['ir.model.data'].xmlid_to_res_id
+        result = []
+
+        def get_id(cr, xml):
+            if xml == 'NULL':
+                return 'NULL'
+            try:
+                return gid(cr, SUPERUSER_ID, xml, raise_if_not_found=True)
+            except:
+                logger.error((
+                    "'map_values_by_xml' faild to retrieve a Database ID "
+                    "for External ID {0}. This mapping will be ignored."
+                ).format(xml))
+                pass
+        for rec in mapping:
+            val = (get_id(cr, rec[0]), get_id(cr, rec[1]))
+            if val[0] and val[1]:
+                result.append(val)
+        return result
+
+    if xml_mapping:
+        mapping = use_xml_mapping(cr, xml_mapping)
 
     if write not in ('sql', 'orm'):
         logger.exception(
@@ -836,6 +867,24 @@ def map_values(
             "map_values is called with the same value for source and old"
             " columns : %s",
             source_column)
+    if check_completeness:
+        values = {
+            'table': table,
+            'source_column': source_column,
+        }
+        query = ("SELECT {source_column!s} FROM {table!s}").format(**values)
+        cr.execute(query)
+
+        missing = set([i[0] for i in cr.fetchall()]) - \
+                  set([i[0] for i in mapping])
+        if (missing - set([None])):
+            for dbid in missing:
+                logger.error((
+                    "'map_values_by_xml' has detected a missing mapping for "
+                    "Database ID {0} in Source Column (check_completness "
+                    "enabled)."
+                ).format(dbid))
+            raise
     for old, new in mapping:
         new = "'%s'" % new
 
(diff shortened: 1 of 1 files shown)


blaggacao:
I also would appreciate feedback in code style / layout / control flow ...

blaggacao:
@Yajo with this and a careful integration of your FK-query, managed-data instances get within reach! I refer to managed-data instances, when you offer clients to "life-update" certain country specific metadata such as a comprehensive set of default partners, taxes, chart of account changes, and other metadata who constantly changes at the country level.

hbrunn:
the obvious use case eludes me to be honest. Especially why we need a new function for this, I do see how it's useful if `map_values` also understands xmlids.

Before looking at style myself, travis already has enough to say: https://travis-ci.org/OCA/openupgradelib/jobs/138266082#L193

blaggacao:
Thanks @hbrunn for looking at this. How map_value understands xml_ids eludes me indeed. I should have added, that I didn't expect it to be mergeable, as I wanted to get feedback on the idea and style first. (fixing travis becomes a priority to get it merge ready)

hbrunn:
use some type that will break as well the sql case and the orm case and use this as marker for an xmlid. Tuples of the form `('module', 'id')` come to mind.

Fixing travis for me is a preliminary to take a closer look at code in the first place.

blaggacao:
Could I do: `(module.externalid)` within parenthesis to keep the notation of regular odoo xml files?

blaggacao:
Keeping on thinking about this, it would be even very useful to have `map_values` beeing able to cross tables. For if you want map value X of Column Y of Table Z into Value X' of Column Y' in Table Z'. @hbrunn Any objections or enthusiasm for this idea?

hbrunn:
no enthusiasm, but also no objections. I fear you're overengineering stuff for theoretical reasons. This library developed around needs that arose when writing OpenUpgrade, so most people anchor a contribution here with a contribution needing it in OCA/Openupgrade

blaggacao:
Maybe I'm trapped in my line of thinking. I haven't yet clear a sensible scope of this. It shows that when I need to copy data from a migrated 2many field, like in 
>`field res_partner.ciiu` was referenced to model `account.tax` and now goes to model `ciiu.codes`
things become quickly messed up, because in a `*_rel` table, the following breaks, which is actually a undocumented limitation of the function (sidenote to myself: I should document this in another PR):
```
 else:
            query = """SELECT id FROM %(table)s
                       WHERE %(source)s %(op)s %(old)s""" % values
```
And I haven't clear at all, how `*_rel` table handling would look like in the ORM method.
That's why I'm motivated to discuss on "functionality (completeness and scope)" first.

blaggacao:
To organize things and get a little free sight into the foggy thoughts, I could shell out `use_xml_mapping` into `map_values` as you proposed. I liked the idea of type checking for this purpose.
I also could include the `check_completeness` into `map_values`, as it is also helps constructing a mapping dict with Database IDs there.

blaggacao:
I think now it's a bad idea to make the detection, because this would make the control flow more complex, and I descart the probable existence the use case wanting to mix Database IDs and External IDs in one mapping. So I'll go with the flags.

blaggacao:
:boom: :tada: :blossom: 
Ok, it's now refactored, as discussed above.This flake8 is annoying :smile: , will fix now.

I would be happy if we could include this quickly, so I can see how to tackle the thing with the relational tables.

blaggacao:
:tada: :tada: :tada: maybe -> https://github.com/google/yapf can do good.

StefanRijnhart:
Looks as if you want to make your functionality a wrapper around map_values instead of integrating it.

blaggacao:
@StefanRijnhart it was my first approach, as I thought better not touch existing methods, yet as @hbrunn  argumented he sees it more as an improvement to `map_values` and finally I agree, I already changed the PR in the more later commits. Now it's integrated.

StefanRijnhart:
That first version does look cleaner, actually.

blaggacao:
Every time I do a refactoring here, I also need to do an internal refactoring of my actual use case, and I wanted to move on to tackle the problem of 2many fields which is blocking a pending client migration. I would be happy, if for the sake of @hbrunn 's previous opinion, we could stick with that, if the more recent integration would not be blocking.

blaggacao:
@bwrsandman @pedrobaeza @legalsylvain 
As we have a situation of differing opinions between core maintainers here, could you probably give your opinion, please? I would appreciate if this would be merged, to be able to move on and build upon this.

It would be therefore nice if we could proceed to review process quickly.

blaggacao:
@StefanRijnhart A strong argument to integrate comes to my mind: I'm planning to make a wrapper to map_values to deal with many2many fields, which is a bit more complicated. It would be quite ugly to make two wrapper, one for external ids and one for DB ids. Would this resolve your objection?

openupgradelib/openupgrade.py
```diff
@@ -836,6 +867,26 @@ def map_values(
...
+        missing = set1 - set2
+
+        if (missing - set([None])):
+            for dbid in missing:
+                logger.error((
+                    "'map_values_by_xml' has detected a missing mapping for "
+                    "Database ID {0} in Source Column (check_completness "
+                    "enabled)."
+                ).format(dbid))
+            raise
```
blaggacao: @hbrunn Bad idea, is ist?

openupgradelib/openupgrade.py
```diff
@@ -799,29 +799,60 @@ def float_to_integer(cr, table, field):
...
+    have been accounted for in the mapping table and raises otherwise.
 
+    This moethod does not support mapping values across tables, neither \
+    relational nor property fields (it does not support resiliant \
+    handling of their respective relational tables - you still can use \
+    this method on those relational tables in sql mode).
     .. versionadded:: 8.0
     """
+    def use_xml_mapping(cr, mapping):
+        from openerp.modules.registry import RegistryManager
```
blaggacao: I think we get that from global imports already...

openupgradelib/openupgrade.py
```diff
@@ -799,29 +799,60 @@ def float_to_integer(cr, table, field):
 
 
 def map_values(
-        cr, source_column, target_column, mapping,
-        model=None, table=None, write='sql'):
+        cr, source_column, target_column, mapping, xml_mapping=False,
```
pedrobaeza: Please add new arguments at the end, for keeping compatibility with those that pass arguments positionally

openupgradelib/openupgrade.py
```diff
@@ -799,29 +799,60 @@ def float_to_integer(cr, table, field):
...
     :param model: used for writing if 'write' is 'orm', or to retrieve the \
     table if 'table' is not given.
     :param table: the database table used to query the old values, and write \
     the new values (if 'write' is 'sql')
     :param write: Either 'orm' or 'sql'. Note that old ids are always \
     identified by an sql read.
+    :param check_completeness: True verifies, that all source entries \
+    have been accounted for in the mapping table and raises otherwise.
 
+    This moethod does not support mapping values across tables, neither \
```
pedrobaeza: s/moethod/method

pedrobaeza: I don't get the meaning of this paragraph. Can you explain better?

openupgradelib/openupgrade.py
pedrobaeza: Making this, you are restricting the use of this function on post_migration script. Maybe you can do it with plain SQL.

openupgradelib/openupgrade.py
pedrobaeza: Well, if the option is True, you should stop the migration, but better to pass the previous text to an `Exception`.

blaggacao: I'm not sure about this as it's per line level, so raising would raise on every missing mapping, so you would need to run the script x times in order to get all x faulty mappings, instead of a simple look at the logs.

pedrobaeza: No if you raise an Exception with a text that contains the full list, that can be simply made with a `.join()`.

openupgradelib/openupgrade.py
blaggacao: Aha, I'm about to understand the SQL proneness, can you give an example how this can break in post-migration scripts for me to better understand?

openupgradelib/openupgrade.py
blaggacao: Did I miss something? I wasn't aware of that keyword arguments can be passed positionally... Is that so?

openupgradelib/openupgrade.py
pedrobaeza: If you make it on pre-migration scripts, the models that this module uses are not yet loaded. Making in plain SQL, this doesn't matter.

openupgradelib/openupgrade.py
pedrobaeza: Yes, this is Python basics.

openupgradelib/openupgrade.py
blaggacao: Actually this method as is does not support mapping of m2m fields. from a source-m2m to a destination-m2m. I was having that use case and was trying confidently this method as it didn't advert it's not made for this.

openupgradelib/openupgrade.py
blaggacao: Ok so you're saying that using this in pre-migration script is limited when the fields of the module the specs refer to are not yet loaded. But then I wonder, what would be a good reason to do that?
If you want to migrate to a new model, you probably would do that after the model is properly created and copy the column meanwhile in order to keep the values until post-migrate. As stated in variously I'm working on a method to do that consistently also with m2m fields.

openupgradelib/openupgrade.py
blaggacao: Thanks! Didn't know. :dancer:

blaggacao:
What's happening with flake? Those last two commits don't even touch those lines...

blaggacao:
@StefanRijnhart It's carved out again. :smile: 

The recent commit also homogenates the use of "magic values" in the mapping table:

Pythons `True`, `False` and `None` are needed to be mapped to the corresponding SQL values for retrieving the source ids from the tables.

On the other hand SQL's `NULL`, `FALSE`, `TRUE`, `NOT NULL` should also be allowed and handled properly by switching the SQL operator.

On the target side, when writing through "orm" `False`, `True`, `None` are allowed, whereas while writing through sql `NULL`, `FALSE`, `TRUE`, `NOT NULL` should and are to be handled by the correct operator gracefully.

openupgradelib/openupgrade.py
```diff
@@ -836,18 +868,39 @@ def map_values(
...
+                "following values in Source Column (check_completness "
+                "enabled):\n{0}".format('\n'.join(missing))
+            )
+
     for old, new in mapping:
         new = "'%s'" % new
 
-        if old is True:
-            old = 'NOT NULL'
+        if old in [True, False, None, 'NULL', 'TRUE', 'FALSE', 'NOT NULL']:
```
StefanRijnhart: What if you want to match a string value 'NULL' in a char field?

StefanRijnhart:
No, the m2m future extension does not change my opinion. In fact, it sounds like this extension would obfuscate this method even more if integrated.

blaggacao:
replaced by #53

PR #36, opened 2016-06-18
[IMP] Add dropping m2m tables

blaggacao:
- Drops many2many tables after migration through the ORM

Use case:
Dropping a many2many field in a migration is not automatically cleaned by the ORM (only deleting the whole module does that). So if you removed a m2m field in a model through migration you want to clean up after you thanks to this small wrapper.

Thanks in advance for your time, feedback, collaboration and eventual merging.

### DIFF
--- a/openupgradelib/cleanup.py
+++ b/openupgradelib/cleanup.py
@@ -28,7 +28,7 @@ logger.setLevel(logging.DEBUG)
 
 ###########################################
 #
-# NOTE: Please try to homologate this library with
+# NOTE: Please try to syntesize this library with
 # database_cleanup module as you contribute
 # You can find it here:
 # https://github.com/OCA/server-tools/tree/9.0/database_cleanup
@@ -54,7 +54,7 @@ def drop_m2m_table(env, table_spec):
     :param cr: The database cursor
     :param table_spec: list of strings ['table one', 'table two']
 
-    .. versionadded:: 9.0
+    Note: You can use this method from version 8.0 onwards.
     """
     drop = env['ir.model.relation']._module_data_uninstall
     for table in table_spec:
@@ -63,6 +63,9 @@ def drop_m2m_table(env, table_spec):
         env.cr.execute(query)
         ids = [x[0] for x in env.cr.fetchall()]
         drop(ids)
+        query = """DELETE FROM ir_model_relation
+                   WHERE name='{0}'""".format(table)
+        env.cr.execute(query)
 
 
 def drop_columns(cr, column_spec):
--- a/openupgradelib/cleanup.py
+++ b/openupgradelib/cleanup.py
@@ -1,0 +1,84 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    This module copyright (C) 2011-2013 Therp BV (<http://therp.nl>)
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import logging
+# from .openupgrade import version_info
+
+logger = logging.getLogger('OpenUpgradeCleanup')
+logger.setLevel(logging.DEBUG)
+
+###########################################
+#
+# NOTE: Please try to homologate this library with
+# database_cleanup module as you contribute
+# You can find it here:
+# https://github.com/OCA/server-tools/tree/9.0/database_cleanup
+#
+# Feel yourself encouraged to contact the author of a
+# similar method in databse_cleanup (git blame).
+#
+###########################################
+
+
+__all__ = [
+    'drop_m2m_table',
+    'drop_columns',
+]
+
+
+def drop_m2m_table(env, table_spec):
+    """
+    Drop a many2many relation table properly.
+    You will typically want to use it in post-migration scripts after you \
+    have migrated the values of your many2many fields.
+
+    :param cr: The database cursor
+    :param table_spec: list of strings ['table one', 'table two']
+
+    .. versionadded:: 9.0
+    """
+    drop = env['ir.model.relation']._module_data_uninstall
+    for table in table_spec:
+        query = """SELECT id FROM ir_model_relation
+                   WHERE name='{0}'""".format(table)
+        env.cr.execute(query)
+        ids = [x[0] for x in env.cr.fetchall()]
+        drop(ids)
+
+
+def drop_columns(cr, column_spec):
+    """
+    Drop columns but perform an additional check if a column exists.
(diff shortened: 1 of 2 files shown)


blaggacao:
I'll do the CI cleaning right away, as I think this is mergeable. One second..

blaggacao:
:tada: :boom: :blossom: CI green.

openupgradelib/openupgrade.py
```diff
@@ -398,6 +398,27 @@ def drop_columns(cr, column_spec):
                         table, column)
 
 
+def drop_m2m_table(cr, table_spec):
```
blaggacao: @hbrunn need this be included into the `_all_` dict? I'm inclined to guess so without being certain about it.

pedrobaeza:
There's 2 reasons why not do this:

* There can be other migration scripts (of dependent modules) that need this table (and you don't know it), so it's not a good practice to try to drop tables.
* This task can be accomplished with the module database_cleanup.

So for me, this is :-1:

blaggacao:
@pedrobaeza There is a use case for module-only migration (not the whole database). On the other hand, database_cleanup is an external module, which does not reside in migrations scripts. However in order to have a planable consistent state over several, possibly timely disperse migrations cycle, I want to be able to be very specific about my cleaning homework.

pedrobaeza:
Sorry, but my opinion is that one. Including this function in the library will encourage its use with not the best practices, as stated before. If you want this, implement in your own library + migration scripts.

blaggacao:
Ok fair. Well we could probably leave it open, but without merging? I'm still thinking of openupgradelib when it comes to migration tasks rather than my own fork, and I would be glad to do so in future.

blaggacao:
@pedrobaeza I had a closer look at `database_cleanup`. It's terribly single purposed (a module, seriously?) and I would have no idea how to incorporate TransientModel ClassMethods into my migration scripts. It is clear, that some migrations tasks in pre require housekeeping in post. Note that you presumably have a different understanding of "migration tasks", than I do. However, I think you're main objection is still valid. Being consequent yourself, you should therfore also rise on stuff like [this](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L78-L79), I guess.

One path to solution could be to isolate cleanup functions into a cleanup.py, by this the documented standard import `import openupgradelib.openupgrade` would hide this away, and if you really need stuff like this you would be glad to find it in `import openupgradelib.cleanup`. People who contribute on `cleanup` could be encouraged to homologize with corresponding functions in `databse_update` and can even leave a ping to the blamed contributor of the functions in `database_update` to check if he/she could refactor towards openupgradelib.

This would elegantly address your concern, yet, without doubt, implies extra work. If you really care about your concern, we should team up on this or something similar. If not, then you're probably not warranted by matter of fairness to enforce your objection on behalf of this argument.

StefanRijnhart:
Moving cleanup core functions to openupgradelib seems fine to me.

pedrobaeza:
@StefanRijnhart, we shouldn't allow auto-clean methods that removes the possibility to use that data lately in other scripts.

StefanRijnhart:
Well, I always viewed the database_cleanup module as a part of the openupgrade project. As it is potentially useful for other purposes, it is included in the more general server-tools repository. Now, @blaggacao implies that having the methods available in other contexts can be useful too, which can be achieved by moving the methods to its own lib, or as I prefer, to openupgradelib (openupgradelib.cleanup as suggested above). I still agree with you that database cleaning is to be done in some post migration phase as per OpenUpgrade policy, and not as an integral part of running the OpenUpgrade server but that does not mean that we cannot facilitate whatever other people like to do with their Odoo databases.

blaggacao:
@pedrobaeza I the last commit pays due respect to your valid objections. In line with my previous comment, I also shelled out `drop_columns` as it bears similar categorial use, maintaining backwards compatibility, yet with logging a warning message inline with your concerns.
@StefanRijnhart Does this head into a welcomed direction?

openupgradelib/cleanup.py
```diff
@@ -0,0 +1,85 @@
...
+def drop_m2m_table(env, table_spec):
+    """
+    Drop a many2many relation table properly.
+    You will typically want to use it in post-migration scripts after you \
+    have migrated the values of your many2many fields.
+
+    :param cr: The database cursor
+    :param table_spec: list of strings ['table one', 'table two']
+
+    .. versionadded:: 9.0
```
StefanRijnhart: This is meaningless now that openupgradelib is shared between all versions. But you will want to make a remark that this method will not work before 7.0 as m2m relations were not registered in this way back then.

openupgradelib/openupgrade.py
```diff
@@ -379,23 +379,15 @@ def add_xmlid(cr, module, xmlid, model, res_id, noupdate=False):
...
-    for (table, column) in column_spec:
-        logger.info("table %s: drop column %s",
-                    table, column)
-        if column_exists(cr, table, column):
-            cr.execute('ALTER TABLE "%s" DROP COLUMN "%s"' %
-                       (table, column))
-        else:
-            logger.warn("table %s: column %s did not exist",
-                        table, column)
+    logger.warning(
```
StefanRijnhart: Mabye 
```
import warnings
warnings.warn("your message", DeprecationWarning)
```

openupgradelib/cleanup.py
```diff
@@ -0,0 +1,85 @@
...
+###########################################
+
+
+__all__ = [
+    'drop_m2m_table',
+    'drop_columns',
+]
+
+
+def drop_m2m_table(env, table_spec):
```
StefanRijnhart: Revert to cr please, and rewrite method to use SQL only. There is no env in 7.0.

openupgradelib/cleanup.py
```diff
@@ -0,0 +1,85 @@
...
+import logging
+from .openupgrade import column_exists
+# from .openupgrade import version_info
+
+logger = logging.getLogger('OpenUpgradeCleanup')
+logger.setLevel(logging.DEBUG)
+
+###########################################
+#
+# NOTE: Please try to homologate this library with
```
StefanRijnhart: You keep using that word. I don't think it means what you think it means. https://en.wikipedia.org/wiki/Homologation

openupgradelib/cleanup.py
```diff
@@ -0,0 +1,85 @@
...
+
+    .. versionadded:: 9.0
+    """
+    drop = env['ir.model.relation']._module_data_uninstall
+    for table in table_spec:
+        query = """SELECT id FROM ir_model_relation
+                   WHERE name='{0}'""".format(table)
+        env.cr.execute(query)
+        ids = [x[0] for x in env.cr.fetchall()]
+        drop(ids)
```
StefanRijnhart: Also delete the entry from ir_model_relation.

openupgradelib/openupgrade.py
blaggacao: Nice

openupgradelib/cleanup.py
blaggacao: Is there a strong use case for 7.0, still? I mean in not supporting 7.0 this PR does not break things...

openupgradelib/cleanup.py
blaggacao: Done.

pedrobaeza:
Is this going to be continued? Why not copying the database_cleanup code?

pedrobaeza:
Closing as no response

blaggacao:
That's allright, Pedro...  Thanks for the cleaning! 

Meanwhile we dropped openupgrade lib completely and work pure SQL for speed and reducing external dependency for the team. 

What's missing though is a general approach to coordinate maintenance which are not strictly module migrations (like module installations, renamings with db interactions, etc.) We feel anthem songs from camptocamp are just the right quick'n'dirty tool for that.

PR #37, opened 2016-06-19
[WIP] make map_value relation aware

blaggacao:
Work in Progress.
- Splits up map_value in map_value_orm (which can handle relation fields) and map_value_sql, which cannot
- Proposes new mappin_spec which reduces the function signature and improves readability of migration code (as all info is found where it belongs to in the declaration section)
- Implements first draft of map_value compatibility wrapper
- Implements first draft of map_value_orm
- Deprecates necessarily the table keyword argument, as I could not find a reliable way to transform a table name into a `model._name`. If you know a SQL query, please share it as deprecating inadvertently map_value signature is really really bad and ugly.
- TODO map_value_sql, if it is not decided that it would not be needed and could be changed to deprecated in it's current form and functionality.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -798,14 +798,35 @@ def float_to_integer(cr, table, field):
         })
 
 
-def map_values(
-        cr, source_column, target_column, mapping,
-        model=None, table=None, write='sql'):
+def use_xml_mapping(cr, mapping):
+    from openerp.modules.registry import RegistryManager
+    gid = RegistryManager.get(cr.dbname)['ir.model.data'].xmlid_to_res_id
+    result = []
+
+    def get_id(cr, xml):
+        if xml == 'NULL':
+            return 'NULL'
+        try:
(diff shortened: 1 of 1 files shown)


blaggacao:
Noteworthy: This is atop of #35 , in order to keep each changeset to its smallest possible impact in order to facilitate reviewing. Therefore, please help reviewing #35 first. On the other hand, structural feedback is already very welcome in order to avoid multiple refactorings on almost finished code. Thanks!

Use Case: We have swapped the model behind a many2many field and need to update our databases. We do not want to invest in one time scripts and want our effort to benefit a wider audience.

Openupgradelib, although initially devised for upgrading between major Odoo versions is being used and (has being advised to do so by prominent OCA members) to also be used in minor module migrations. This justifies our attempt to use it for module migration and the inclusion of typical minor migration needs such as the one explained above.

:beers:

blaggacao:
One preliminary question exists, that I want to tackle with you, please:
For what particular reason (end is this reason still valid), map_value has an "sql" mode?

Thanks!

blaggacao:
replaced by #53

PR #38, opened 2016-06-19
[IMP] Add Environment manager for version >=8

blaggacao:
- Includes an environment manager in the @migrate() decorator as of odoo version 8
- Makes it possible to access to the well known new-style `self.env` via `Env`

For instance, the statements:

```
model = self.pool.get(MODEL)
ids = model.search(cr, uid, DOMAIN, context=context)
for rec in model.browse(cr, uid, ids, context=context):
    print rec.name
model.write(cr, uid, ids, VALUES, context=context)
```
may also be written as:
```
model = Env[MODEL]                  # retrieve an instance of MODEL
recs = model.search(DOMAIN)         # search returns a recordset
for rec in recs:                    # iterate over the records
    print rec.name
recs.write(VALUES)  
```

@Yajo Is that mas o menos right, what I did with your idea?

Note the context argument on the migrate decorator for whatever it might be good in the future...

Migrations scripts would need to pass `Env.cr` on behalf of `cr`, but we could easily include the `cr` for legacy reasons.

Of course, once this is codewise mergeable, we need to proceed to proper documentation.

Thanks for your time, collaboration, effort, review and good vibes!

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -962,18 +962,18 @@ def reactivate_workflow_transitions(cr, transition_conditions):
             (condition, transition_id))
 
 
-def migrate(no_version=False, v8api=False, context=None, *args, **kwargs):
+def migrate(no_version=False, v9env=False, context=None, *args, **kwargs):
     """
     This is the decorator for the migrate() function in migration scripts.
     It returns if `version` is not defined and `no_version` is False,
     and logs execeptions. It also retrieves debug context data from the
     frame above for logging purposes. Optionally it provides a fully fledged
-    Odoo Environment (with `v8api` = True). An arbitrary context can be
+    Odoo Environment (with `v9env` = True). An arbitrary context can be
     preloaded into the Environment via `context` dict.
 
     :param no_version: Set to `True` if the migrate method has to be taken
     into account if the module is installed during a migration.
-    :param v8api: Set to `True` gives you a fully fledged Odoo Environment.
+    :param v9env: Set to `True` gives you a fully fledged Odoo Environment.
     The wrapped function signature changes to `migrate(env, *args, **kwargs):`
         Obtain Version: env.context['migrate_version']
         Obtain Cursor: env.cr
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1009,12 +1009,12 @@ def migrate(no_version=False, v8api=False, context=None, *args, **kwargs):
                 "%s: %s-migration script called with version %s" %
                 (module, stage, version))
             versioncheck = (version_info[0] >= 8 and v9env) or version_info > 9
-            if not versioncheck and not v9env:
+            if not versioncheck:
                 logger.warning(
                     "You are are using the decorator without the Environment "
                     "creation feature. As of Odoo Version 10 Migrations, the "
                     "Environment will be enabled by default. Make sure to "
-                    "gat aquiainted with the slightly different migrate() "
+                    "get acquainted with the slightly different migrate() "
                     "signature in time: `def migrate(env, *args, **kwargs)`"
                 )
             try:
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -962,16 +962,29 @@ def reactivate_workflow_transitions(cr, transition_conditions):
             (condition, transition_id))
 
 
-def migrate(no_version=False, context=None, *args, **kwargs):
+def migrate(no_version=False, v8api=False, context=None, *args, **kwargs):
     """
-    This is the decorator for the migrate() function
-    in migration scripts.
-    Set argument 'no_version' to True if the method as to be taken into account
-    if the module is installed during a migration.
-    Return when the 'version' argument is not defined and no_version is False,
-    and log execeptions.
-    Retrieve debug context data from the frame above for
-    logging purposes.
+    This is the decorator for the migrate() function in migration scripts.
+    It returns if `version` is not defined and `no_version` is False,
+    and logs execeptions. It also retrieves debug context data from the
+    frame above for logging purposes. Optionally it provides a fully fledged
+    Odoo Environment (with `v8api` = True). An arbitrary context can be
+    preloaded into the Environment via `context` dict.
+
+    :param no_version: Set to `True` if the migrate method has to be taken
+    into account if the module is installed during a migration.
+    :param v8api: Set to `True` gives you a fully fledged Odoo Environment.
+    The wrapped function signature changes to `migrate(env, *args, **kwargs):`
+        Obtain Version: env.context['migrate_version']
+        Obtain Cursor: env.cr
+        To genuinely support calling legacy methods you should do:
+        ```
+        def migrate(env, *args, **kwargs):
+            version = env.context['migrate_version']
+            cr = env.cr
+        ```
+    :param context: An arbitrary odoo context you might want to set up. You
+    can access this through `env.context`, where the Environment is present.
     """
     def wrap(func):
         @wraps(func)
@@ -995,11 +1008,21 @@ def migrate(no_version=False, context=None, *args, **kwargs):
             logger.info(
                 "%s: %s-migration script called with version %s" %
                 (module, stage, version))
+            def call_funct
+            versioncheck = (version_info[0] >= 8 and v9env) or version_info > 9
+            if not versioncheck and not v9env:
+                logger.warning(
+                    "You are are using the decorator without the Environment "
+                    "creation feature. As of Odoo Version 10 Migrations, the "
+                    "Environment will be enabled by default. Make sure to "
+                    "gat aquiainted with the slightly different migrate() "
+                    "signature in time: `def migrate(env, *args, **kwargs)`"
+                )
             try:
                 # The actual function is called here
                 if hasattr(cr, 'savepoint'):
                     with cr.savepoint():
-                        if version_info[0] >= 8:
+                        if versioncheck:
                             with api.Environment.manage():
                                 context['migrate_version'] = version
                                 env = api.Environment(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is not a bad idea, but there's too much to be done before this can merged:

* Use another decorator, not the `migrate` one, that is for other purposes.
* Prevent their use on versions < 8
* Document it properly
* With a decorator, how do you get the variable? (this is only inexperience in decorators...)

Anyway, as this is very easy to achieve with one line of code inside of migration scripts, I'm not sure if this deserves the time.

blaggacao:
@pedrobaeza That's definitely discouraging. However, I'll try:

- What's the purpose of the @migrate decorator, that excludes this there.
- < v8 is excluded, or did I do something wrong here? Or probably I don't understand what you mean.
- Documentation is pending as stated, but first the code itself should be in a mergeable state, a documentation commit is the least difficult thing about it, I guess.
- Your last point, I honestly do not understand it.

blaggacao:
@pedrobaeza I think this is getting closer to "review worthy" in 
["review worthy", "code ready", "docs wip", "docs ready", "final review"]
The new backwards compatible decoration signature would be:
```
@manage(context={'test': 'nice'})
def migrate(Env):
    version = Env.context['migrate_version']
    cr = Env.cr
```
This makes me think for being completely backwards compatible, that maybe a context flab would be needed in order to enable this feature deliberately and therefor not to break existing migration scripts. Fro version 10 it could then be scheduled as default. Would that be good?

blaggacao:
There is something I don't understand about: http://stackoverflow.com/questions/37915879/index-out-of-range-in-python-in-an-apparently-inconsistent-way
@Yajo Any idea? Could this be related to the Environment.manage() function? Or to concurrency?

StefanRijnhart:
Good idea, if it can be made compatible with existing calls like https://github.com/OCA/OpenUpgrade/blob/8.0/addons/account/migrations/8.0.1.1/pre-migration.py#L25

blaggacao:
@StefanRijnhart Done. The current commit, propagates this style to default as of version 10 migration scripts and logs a proper and concise warning. I think, this is in the interest in encouraging best practices and forward compatibility.

openupgradelib/openupgrade.py
```diff
@@ -991,17 +1008,44 @@ def wrapped_function(cr, version):
...
+                            with api.Environment.manage():
+                                context['migrate_version'] = version
+                                env = api.Environment(
+                                    cr, SUPERUSER_ID, context
+                                )
+                                func(env, *args, **kwargs)
+                                cr.execute('RELEASE SAVEPOINT "%s"' % name)
+                        else:
+                            func(cr, version)
+                            cr.execute('RELEASE SAVEPOINT "%s"' % name)
```
Yajo: You have the `with cr.savepoint():` for this stuff.

Yajo:
This looks great, I think it will encourage new migrations to become more developer-friendly, while still retaining the old behavior intact.

pedrobaeza:
This option is always available by developers with 2 lines of code, but I agree it can be easier integrating it in the decorator. I told about using another decorator to avoid the compatibility problem.

openupgradelib/openupgrade.py
```diff
@@ -960,19 +962,34 @@ def reactivate_workflow_transitions(cr, transition_conditions):
             (condition, transition_id))
 
 
-def migrate(no_version=False):
+def migrate(no_version=False, v9env=False, context=None, *args, **kwargs):
```
Yajo: How about defaulting to `v9env=None`? Then, if it is True or (None and version > 9), enable it. So, you allow old-style lovers to pass `v9env=False`.

openupgradelib/openupgrade.py
StefanRijnhart: @Yajo no, because we want this to keep working in Odoo versions < 8.0.

Yajo: ah ok

openupgradelib/openupgrade.py
StefanRijnhart: Apply argument inspection?

Yajo: Just check if it is `False` or `None`.

blaggacao: Good idea, thought about something even more versatile, but didn't come up with the idea to set it `None`.

blaggacao:
@Yajo I added your suggestion.

blaggacao:
@pedrobaeza @StefanRijnhart @hbrunn @Yajo 

I think we can proceed to last feedback round as to functionality and code ... If there are no further objections, please give me your "LGTM for code and functionality" , if you are in favor of this.

Anything that needs additional proper documentation? I'm not sure if it would be a good idea if it is the first and only item in the features list (it would be a bit out of context), but I can put it there if you want.

openupgradelib/openupgrade.py
```diff
@@ -960,19 +962,35 @@ def reactivate_workflow_transitions(cr, transition_conditions):
...
-    This is the decorator for the migrate() function
-    in migration scripts.
-    Set argument 'no_version' to True if the method as to be taken into account
-    if the module is installed during a migration.
-    Return when the 'version' argument is not defined and no_version is False,
-    and log execeptions.
-    Retrieve debug context data from the frame above for
-    logging purposes.
+    This is the decorator for the migrate() function in migration scripts.
+    It returns if `version` is not defined and `no_version` is False,
```
StefanRijnhart: I find this less clear than the original explanation of 'no_version'. Every function always 'returns'.

openupgradelib/openupgrade.py
```diff
@@ -960,19 +962,35 @@ def reactivate_workflow_transitions(cr, transition_conditions):
             (condition, transition_id))
 
 
-def migrate(no_version=False):
+def migrate(no_version=False, v9env=None, context=None, *args, **kwargs):
```
StefanRijnhart: Why is it called v9env? Environments are available from 8.0 on.

openupgradelib/openupgrade.py
```diff
@@ -991,17 +1009,46 @@ def wrapped_function(cr, version):
...
+                    "signature in time: `def migrate(env, *args, **kwargs)`"
+                )
             try:
                 # The actual function is called here
                 if hasattr(cr, 'savepoint'):
                     with cr.savepoint():
-                        func(cr, version)
+                        if versioncheck:
+                            with api.Environment.manage():
+                                context['migrate_version'] = version
```
StefanRijnhart: It looks like context could still be 'None' at this point.

openupgradelib/openupgrade.py
```diff
@@ -991,17 +1009,46 @@ def wrapped_function(cr, version):
...
+                                return func(env, *args, **kwargs)
+                        else:
+                            return func(cr, version)
                 else:
                     name = uuid.uuid1().hex
                     cr.execute('SAVEPOINT "%s"' % name)
                     try:
-                        func(cr, version)
-                        cr.execute('RELEASE SAVEPOINT "%s"' % name)
+                        if versioncheck:
```
StefanRijnhart: This code duplication bothers me. I would suggest to either using cr.savepoint completely (I would prefer this) or put the duplicated code in an inline method. Note that the savepoint release does not have to be scoped under the with statement below, so it does not have to be included in such a function.

openupgradelib/openupgrade.py
blaggacao: You're right. It should rather be just `env`.

openupgradelib/openupgrade.py
blaggacao: Ok, will homologate it.

openupgradelib/openupgrade.py
blaggacao: Thanks for the pair of eyes!

openupgradelib/openupgrade.py
blaggacao: Yep, is ugly like it is. Thanks for the hint about cursor release! Stuff I'm not aware yet... :smile:

blaggacao: On a closer look: If `cr` has no attribute savepoint, it won't have an environment either. So this was basically a no-op.

openupgradelib/openupgrade.py
```diff
@@ -960,19 +962,35 @@ def reactivate_workflow_transitions(cr, transition_conditions):
...
+            cr = env.cr
+        ```
+    :param context: An arbitrary odoo context you might want to set up. You
+    can access this through `env.context`, where the Environment is present.
     """
     def wrap(func):
-        def wrapped_function(cr, version):
+        @wraps(func)
+        def wrapped_function(cr, version, no_version=no_version,
+                             context=context, *args, **kwargs):
```
blaggacao: @StefanRijnhart Here was another one `env=env` was missing. It looks like this has become necessary because of the way the `@wraps` decorator works.

openupgradelib/openupgrade.py
```diff
@@ -960,19 +962,38 @@ def reactivate_workflow_transitions(cr, transition_conditions):
...
-    and log execeptions.
-    Retrieve debug context data from the frame above for
-    logging purposes.
+    This is the decorator for the migrate() function in migration scripts.
+    It retrieves debug context data from the frame above for logging purposes.
+    Optionally it provides a fully fledged Odoo Environment
+    (with `env` = True). An arbitrary context can be preloaded into the
+    Environment via `context` dict. Otherwise returns, if the 'version'
+    argument is not defined and no_version is False, and logs exception.
+
```
blaggacao: @StefanRijnhart Is this more or less what you were missing?

openupgradelib/openupgrade.py
blaggacao: added defaulting to `{}` in line 994

StefanRijnhart:
Thank you for the followup. All looks good now, but migration fails

2016-06-27 07:47:32,868 4333 ERROR demo80all OpenUpgrade: base: error in migration script /home/bruno/buildout/ou90/parts/odoo/openerp/addons/base/migrations/9.0.1.3/pre-migration.py: migrate() takes exactly 2 arguments (1 given)
2016-06-27 07:47:32,868 4333 ERROR demo80all OpenUpgrade: migrate() takes exactly 2 arguments (1 given)
  File "/home/bruno/buildout/ou90/openupgradelib/openupgradelib/openupgrade.py", line 1045, in wrapped_function
    return func(env, *args, **kwargs)

StefanRijnhart:
That is the migration of the base module of 9.0

StefanRijnhart:
obsoleted by #51

PR #40, opened 2016-06-19
[IMP] prep get_legacy_name for orm copy and include module version

blaggacao:
- prefixes columns with "x_", the orm standard for manual fields. 
This enables a "pre-post migration persistent" field copy function through the ORM, which in turn enables m2m copying.
- Include module version, as openupgradelib is increasingly used in individual module migration.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -692,8 +692,8 @@ def get_legacy_name(original_name):
     :param original_name: the original name of the column
     :param version: current version as passed to migrate()
     """
-    return 'openupgrade_legacy_' + '_'.join(
-        map(str, version_info[0:2])) + '_' + original_name
+    return 'x_ou_legacy_' + '_'.join(
+        map(str, version_info[0:])) + '_' + original_name
 
 
 def m2o_to_x2m(cr, model, table, field, source_field):


pedrobaeza:
`x_` is for custom fields added by interface, not for this function, that let the DB columns renamed, but at SQL level, not at Odoo level (they are not accessible via ORM), so this is not useful at all.

:-1: 

Thanks anyway for proposing.

blaggacao:
@pedrobaeza This is a prerequisite for the PR I'm preparing, so it would be very nice of you if you could explain your objections for the general use case. I think thumbs down should be properly argued.

pedrobaeza:
My arguments are already said. You can rename to what you want in your scripts, so please do it there, not changing the main functionality for `None` value.

blaggacao:
@pedrobaeza As I said, it is a prerequisite for a PR I'm planning, on which value you cannot judge yet. Don't you think it would be nicer just wait for this PR tu judge on the whole picture? And reopen this in the meantime. Please, don't be unnecessarily disrespectful.

blaggacao:
Well, I'll carve out the version part as this breaks in module migrations if you use the advised `get_legacy_column()` method. With more detailed version, the module-migration case is better supported without breaking anything on the major version migration. Is that correct?

PR #43, opened 2016-06-19
[IMP] detail version on legacy columns

blaggacao:
- Accounts for the minor-version use case
- Avoids conflicting and breaking column names in  minor-version upgrades

This is the undisputed part of #40 so I carve it out here.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -693,7 +693,7 @@ def get_legacy_name(original_name, version=None):
     :param version: current version as passed to migrate()
     """
     if version:
-        vstr = '_'.join('.',split(version))
+        vstr = '_'.join(version.split('.'))
     else:
         vstr = '_'.join(map(str, version_info[0:2]))
     return 'openupgrade_legacy_' + vstr + '_' + original_name
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -683,7 +683,7 @@ def add_ir_model_fields(cr, columnspec):
         logged_query(cr, query, [])
 
 
-def get_legacy_name(original_name):
+def get_legacy_name(original_name, version=None):
     """
     Returns a versioned name for legacy tables/columns/etc
     Use this function instead of some custom name to avoid
@@ -692,8 +692,11 @@ def get_legacy_name(original_name):
     :param original_name: the original name of the column
     :param version: current version as passed to migrate()
     """
-    return 'openupgrade_legacy_' + '_'.join(
-        map(str, version_info[0:])) + '_' + original_name
+    if version:
+        vstr = '_'.join('.',split(version))
+    else:
+        vstr = '_'.join(map(str, version_info[0:2]))
+    return 'openupgrade_legacy_' + vstr + '_' + original_name
 
 
 def m2o_to_x2m(cr, model, table, field, source_field):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is also not useful, as the version_info is for Odoo/OpenERP, not for the module itself, so you won't find 8.1, 8.2 versions, only 8, 9, 10....

:-1:

blaggacao:
Does this mean that `version_info` drops this valuable?
```
openerp.modules.migration: module l10n_co_terceros: Running migration [>9.0.1.2] pre-migration
OpenUpgrade: l10n_co_terceros: pre-migration script called with version 9.0.1.1
```
I think with version as in:
```
logger.info(
                "%s: %s-migration script called with version %s" %
                (module, stage, version))
```
things could work.

openupgradelib/openupgrade.py
```diff
@@ -692,8 +692,11 @@ def get_legacy_name(original_name):
     :param original_name: the original name of the column
     :param version: current version as passed to migrate()
```
blaggacao: @pedrobaeza please note this line.

blaggacao:
Isn't there a way to make this globally available? So that it would not need to be passed to every function that would need it? Kind of include it in the `@migrate` decorator?

blaggacao:
@StefanRijnhart This brings back the version parameter as per your initial docstring. A more detailed version is warranted as for example when openupgradelib is in a minor module migration series. being more detailed about the version avoids already the most common naming conflicts. However, there is an edge case which would be not covered by this: When I treat a foreign model, in one's modlu migration script, and then afterwards happen to treat the same model in another module with coinciding small version numbers, without doing a cleanup in the meantime. I think that's generally acceptable as being not supported, but minor version update series should come back.

Sidenote/Idea:
As the minor version would be available in the hopefully accepted Environment's context, I'm thinking of a way to decorate this function in order to keep the old signature, yet enable the use of the environment. I would guess that the Environment is managed by Environment.manage() and therefore calling it from the courser might retrieve the very same environment which holds the version info. If that would be possible we could easily overcharge several methods with an environment while fully remaining compatible with the old signature. This would mean that we would be possible to add this functionality magically all over the codebase without changing a single line of code...

StefanRijnhart:
Explicit is better than implicit. I would rather have it that the
signature is allowed to be either (cr, version) or (env, cr, version).

blaggacao:
Yeay! good point, this would then look like for the generic decorator to make the decorated function accept either (cr, *args, **kwargs) or (env, cr, *args, **kwargs) and mimicking the exact signature of the decorated method, the signature of which can be defaukted to:

@modernizer()
def method(env, cr, otherarg)

this preserves currebtbusage of 'cr' within the function, but enables env addotions while beeing fully backward compatible and while permittig both styled function calls. 

I'll need to check if my assumptions are right and this is possible at all.

This is getting really cool :smile:

StefanRijnhart:
Was I posting on the wrong thread?

This breaks things. The legacy name needs to be consistent from module to module within the same migration. The modules will have different minor version numbers of course.

This method exists to give a meaningful, but still predictable name to obsolete columns.

:-1:

blaggacao:
Ok, I understand. The minor version use case can be covered in a different way through clean up strategies after each minor migration. I deduced initially, that this would be the undisputed part of #40, but now I know better. Preparing the discussion for #45 I would carve out the other part of #40 which seems to be acceptable under this policy and is preliminary for #45.

PR #44, opened 2016-06-19
[IMP] clarifying docstring

blaggacao:
- Clarify that some fields are not supported. 
(it is implicit in columns != fields, but explicitly is more friendly)

I wasn't 100% sure about property fields, but as they are stored in a separate table, I assumed that it's the case like for m2m.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -820,6 +820,9 @@ def map_values(
     :param write: Either 'orm' or 'sql'. Note that old ids are always \
     identified by an sql read.
 
+    This moethod does not support mapping m2m, o2m or property fields. \
+    For o2m you can migrate the inverse field's column instead.
+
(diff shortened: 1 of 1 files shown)


blaggacao:
Carved out from #35 for better separation of concern.

openupgradelib/openupgrade.py
```diff
@@ -820,6 +820,9 @@ def map_values(
     :param write: Either 'orm' or 'sql'. Note that old ids are always \
     identified by an sql read.
 
+    This moethod does not support mapping m2m, o2m or property fields. \
```
hbrunn: typo

StefanRijnhart:
Cherrypicked with typo corrected as https://github.com/OCA/openupgradelib/commit/a9e81757a807b69267e638cc73d89c60ba1599e7

blaggacao:
Thanks!

PR #45, opened 2016-06-19
[WIP]  add orm copy for m2m fields

blaggacao:
- Adds support for conveniently copy m2m fields
- Can only be used in pre-migrations scripts
- This is because it prepares the ir_model_fields and plays hand in hand with the ORM magic during actual update
- When ORM is recreating the renamed column/field this can act in conjunction with map_values_orm in post-migtration script like a "copy" function
- Preserves full backwards compatibility
- TODO Test for supported TTYPES
- TODO Merge a variant of #40 as we need to use manual fields ("x_" - prefixed) in order to fool (circumvent) the loading mechanism
- TODO General Testing
- TODO Getting Env, which can serve as a signature distinguisher for orm and sql methods. That would organize things a bit.

Note: It's WIP

@pedrobaeza I hope this explains you're doubts on the other commits/issues. You can see this as a connected a series of commits, but I split them up to discuss things individually and not have to deal with extremely "fat" commits, which wouldn't be nice for reviewers.

Please ask questions, if you have doubts about the why of this. Maybe it's not necessary to solve copying m2m it this way, so you might want to share your approach and we can discuss in more detail.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -256,10 +256,75 @@ def copy_columns(cr, column_spec):
             })
 
 
-def rename_columns(cr, column_spec):
+def rename_columns_orm(Env, field_spec):
+    """
+    Copy model fields and it's values using the orm. Only works in the 
+    pre script. This method supports m2m fields.
+
+    :param Env: The Environment created y the @migrate() decorator.
+    :param field_spec: a dict with model names as in 'res.partner', with
+        lists of old names. To ensure ORM 'x_' prefix, construction of 
+        new names is exclusively done by get_legacy_name()
+
+    .. versionadded:: 9.0
+    """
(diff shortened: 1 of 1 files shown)


blaggacao:
This is kind of the ultimate explanation for:
#40
#43 
#38 
#37 
#36 
and somewhat #35 for my particular use case.

This being said,  it obviously doesn't need to be done this way. So please if you would argue to architect it differently, please let me know. Thanks for your time and effort, for your collaboration and understanding and ultimately for (hopefully) great contributing experience.

In any case as per some past experiences of mine: Please, please, please don't feel annoyed, this is a serious attempt (as always) to share this in a manner, that more people can benefit from it and give feedback to it. Thanks! :heart:

hbrunn:
depends on #47

blaggacao:
replaced by #54

PR #47, opened 2016-06-27
[IMP] make legacy names orm aware

blaggacao:
- preserves version logic
- reduces legacy prefix (just to save on characters)
- includes odoo's magical `x_` prefix

Rationale:

By prefixing with `x_`, ORM understands `manual` fields, where applicable.

This behaviour enables the creation of backup columns through the ORM in pre migrations scripts, while being able to access this information in post migration.

Without this behaviour, during the actual module upgrade, the ORM would sync DB and Registry with the module's model definition. Any "ORM aware" columns/fields created in pre migration would be wiped away.

A typical use case is the copying of a m2m field together with its values, while being able to access it with the convenience of (metonymy:) recordsets and `mapped()` functions in post migration (what is referred to before as "ORM aware" fields). Pending #45 relies on this behaviour.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -690,9 +690,8 @@ def get_legacy_name(original_name):
     collisions with future or past legacy tables/columns/etc
 
     :param original_name: the original name of the column
-    :param version: current version as passed to migrate()
     """
-    return 'openupgrade_legacy_' + '_'.join(
+    return 'x_ou_' + '_'.join(
         map(str, version_info[0:2])) + '_' + original_name
 
 


hbrunn:
are you sure you don't need records in `ir.model.fields` to have the orm load the appropriate field?

StefanRijnhart:
@hbrunn I think the idea is that you can now create fields in the interface to access the data through the ORM. There is a check on manually created fields that the column starts with the x_ prefix.

This creates ambiguity when cleaning up databases. The x_ prefix is a strong signal for __do not delete this column__. I would prefer to keep the verbose 'openupgrade_legacy' instead, as a strong signal meaning: __after a succesful migration, go ahead and drop this cruft__.

blaggacao:
@StefanRijnhart I absolutely agree with the strong signal, however, especially in the case of large names, I am afraid of hitting this size limit, I don't quite remember anymore. This is why I chose `x_ou` in this commit with `x_` singalling as per rational and `ou_` signalling "migration".

blaggacao:
@hbrunn Honestly, I don't quite understand your question. But @StefanRijnhart observation about the check of the "x_" column is correct. If it has an x_ in front, it is not touched between the migration scripts.

At the risk of totally not targeting your question, the proposed workflow is:
pre-migration.py: create a x_ field via orm where you want to store some temporal data in
between migration: odoo recognizes x_ as untouchable, so the stored information will be preserved.
post-migration.py: Now you can use tha information also via ORM in a convenient way. Eventually, you might want to clean up after yourself. Nonetheless, you can later rely on the `x_ou_` signature, as well.

pedrobaeza:
This feature will need later to make a manual manipulation creating the field in interface, so I think this library is not for that, because it provides tools for automating a migration process. If you need extra columns for you ORM work, then declare then on your module instead of the migration script.

PR #48, opened 2016-06-27
[ADD] rename references to the model in mail related records

hbrunn:


pedrobaeza:
I think this is too specific for mail module and should be done in the mail module, not here

hbrunn:
The mail module can't do this because it is not aware of renamed models. If it was not a dependency of nearly everything in Odoo, I'd agree with you, but with the dependencies being the way they are, I think it's very useful to have this here.
Another option would be to have this function collect a list of model renames, and a mechanism of registering a callback to be executed at the end of the migration, so that for example the mail module could react on renamed models for `mail.message` and `mail.followers`, and everyone else do the same. This would be more beautiful, but I don't see the advantage currently to be honest.

The `ir_translation` part is definitely well placed here.

sebalix:
:+1: If there are many modules like `mail` in the future, a callback approach will scale nicely, but I do not see the point for now.

StefanRijnhart:
Absolutely. Thanks! :+1:

StefanRijnhart:
Needed this for the sale module migration 9.0 as well, so I'm merging this.

StefanRijnhart:
Oh wait, needs a rebase.

hbrunn:
@StefanRijnhart done

StefanRijnhart:
Thanks!

PR #49, opened 2016-07-05
[ADD] lift_constraints

hbrunn:
a nice side effect of this is that we can say `openupgrade.AsIs` when this is merged

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -25,6 +25,7 @@ import inspect
 import uuid
 import logging
 from contextlib import contextmanager
+from psycopg2.extensions import AsIs
 from . import openupgrade_tools
 try:
     from openerp import release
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -25,6 +25,7 @@
 import uuid
 import logging
 from contextlib import contextmanager
+from psycopg2.extensions import AsIs
```
pedrobaeza: Is this a heavy dependency that should be declared on requirements?

StefanRijnhart:
:+1:

openupgradelib/openupgrade.py
hbrunn: no, it comes with standard psycopg2, so everyone has this anyways

sebalix:
:+1:

StefanRijnhart:
Merging, because @pedrobaeza's comment was addressed already.

PR #50, opened 2016-07-06
Update module field in ir_model_fields when calling rename_models.

diggy128:
rename_models() does not change 'module' field in ir_model_fields.
This PR adds 'module' query to properly rename model fields.
A separate query is used from the one that updates 'relation' field as it's not always the same.


pedrobaeza:
:+1: Thanks

StefanRijnhart:
:+1: Thanks!

PR #51, opened 2016-07-21
[ADD] allow to create an environment automatically

hbrunn:


StefanRijnhart:
I think this includes code from the other PR. Can you keep a separate commit for that code so that we don't lose attribution?

hbrunn:
no, all mine

StefanRijnhart:
Thanks for clarifying! :+1:

hbrunn:
if anybody has an idea what's wrong on travis, help would be very appreciated

.travis.yml
```diff
@@ -17,7 +17,7 @@ install:
 
 # command to run tests, e.g. python setup.py test
 script:
-    - python setup.py flake8
+    - flake8 OCA/openupgradelib
```
hbrunn: it's weird, since the travis troubles this doesn't work any more: https://travis-ci.org/OCA/openupgradelib/jobs/147235136
I figure that's just fine too

hbrunn:
@OCA/openupgrade-maintainers yay, green. But have a look at 5488c74, I must admin that I've no clue why this goes wrong if flake8 is called via setup.py

StefanRijnhart:
Thanks for fixing the tests. Here is the related bug report for flake8 https://gitlab.com/pycqa/flake8/issues/164

hbrunn:
thanks for the link, I googled my fingers bloody and couldn't find this. Can we merge this, then the scripts I'm planning to produce soon could use it already?

sebalix:
:+1:

blaggacao:
@hbrunn Would you consider creating a "unittest infrastructure" for openupgradelib? I refer to setting the right mockups etc to ease testing migration functions. I'm quite stuck in the quest for setting up a mock for `env['my.model']`, nor do I know what would be the right approach...

PR #52, opened 2016-07-28
[ADD] rename_property

hbrunn:
this is used in https://github.com/OCA/OpenUpgrade/pull/576

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -120,6 +120,7 @@ __all__ = [
     'reactivate_workflow_transitions',
     'date_to_datetime_tz',
     'lift_constraints',
+    'rename_property',
 ]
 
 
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -1320,3 +1321,22 @@ def savepoint(cr):
             cr.execute('RELEASE SAVEPOINT "%s"' % name)
         except:
             cr.execute('ROLLBACK TO SAVEPOINT "%s"' % name)
+
+
+def rename_property(cr, model, old_name, new_name):
+    """Rename property old_name owned by model to new_name. This should happen
```
pedrobaeza: You can add the parameter documentation in docstring format with `:param`

pedrobaeza:
I think this should be part of `rename_columns` method, to be transparent for the user if the column is a property or not.

hbrunn:
this would involve a lot of auto-detection-foo that I would not want a low level library to do. do you insist on this or can you live with the extra function if there's no overwhelming support for your suggestion?

pedrobaeza:
OK, if no one supports me, then you can continue with this isolated method.

StefanRijnhart:
:+1: I get @pedrobaeza's suggestion, but rename_columns is a very simplistic sql operation (it is not a proper rename_*fields*).

PR #54, opened 2016-08-08
Feature/rename columns orm

blaggacao:
This is a somewhat more versatile ORM copy and rename method set.
replaces https://github.com/OCA/openupgradelib/pull/45


pedrobaeza:
Please don't refactor all code each time you add a feature. Better to have this as an argument like other methods (use_orm=True or similar). Closing waiting for the new PR.

PR #56, opened 2016-08-10
[FIXIMP] fix docstring, bug and imp [Merge Ready]

blaggacao:
- Fix Docstring
- Fix wrong condition
- Set up useful defaults in context
  -- version and module can be used throughout scripts to construct unique hashes of the current precise migration context (eg. for naming schemes)

@hbrunn

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1013,15 +1013,15 @@ def migrate(no_version=False, use_env=None, uid=None, context=None):
     Set argument `no_version` to True if the method as to be taken into account
     if the module is installed during a migration.
 
-    Set argument `pass_env` if you want an v8+ environment instead of a plain
+    Set argument `use_env` if you want an v8+ environment instead of a plain
     cursor. Starting from version 10, this is the default
 
     The arguments `uid` and `context` can be set when an evironment is
     requested. In the cursor case, they're ignored.
 
     The migration function's signature must be `func(cr, version)` if
-    `pass_env` is `False` or not set and the version is below 10, or
-    `func(env, version)` if `pass_env` is `True` or not set and the version is
+    `use_env` is `False` or not set and the version is below 10, or
+    `func(env, version)` if `use_env` is `True` or not set and the version is
     10 or higher.
 
     Return when the `version` argument is not defined and `no_version` is False
@@ -1037,7 +1037,7 @@ def migrate(no_version=False, use_env=None, uid=None, context=None):
             filename = 'unknown'
             with ExitStack() as contextmanagers:
                 contextmanagers.enter_context(savepoint(cr))
-                use_env2 = use_env is None and version_info[0] > 10 or use_env
+                use_env2 = use_env is None and version_info[0] >= 10 or use_env
                 if use_env2:
                     assert version_info[0] >= 8, 'you need at least v8'
                     contextmanagers.enter_context(api.Environment.manage())
@@ -1056,6 +1056,9 @@ def migrate(no_version=False, use_env=None, uid=None, context=None):
                 logger.info(
                     "%s: %s-migration script called with version %s" %
                     (module, stage, version))
+                # Set up useful default context values
+                context['migrate_version'] = version and version or None
+                context['module_name'] = module and module or None
                 try:
                     # The actual function is called here
                     func(
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1035,6 +1035,7 @@ def migrate(no_version=False, use_env=None, uid=None, context=None):
             stage = 'unknown'
             module = 'unknown'
             filename = 'unknown'
+            
             with ExitStack() as contextmanagers:
                 contextmanagers.enter_context(savepoint(cr))
                 use_env2 = use_env is None and version_info[0] >= 10 or use_env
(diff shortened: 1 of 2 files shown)


blaggacao:
@hbrunn the last checks seem to have to do with something not related.

I think te conclusion from this series of changes is valid, that if one value is specified in the decorator (such as in the test), at least then the default values of the decorator were not passed through.

This might have resulted in quite awkward behaviour, I guess...

openupgradelib/openupgrade.py
```diff
@@ -1056,12 +1060,16 @@ def wrapped_function(cr, version):
                 logger.info(
                     "%s: %s-migration script called with version %s" %
                     (module, stage, version))
+                # Set up useful default context values
+                context = context or {}
+                context['migrate_version'] = version and version or None
+                context['module_name'] = module and module or None
```
hbrunn: I don't see how those are useful

openupgradelib/openupgrade.py
```diff
@@ -1031,13 +1032,16 @@ def migrate(no_version=False, use_env=None, uid=None, context=None):
     logging purposes.
     """
     def wrap(func):
-        def wrapped_function(cr, version):
+        @wraps(func)
+        def wrapped_function(cr, version, no_version=no_version, use_env=use_env,
+                             uid=uid, context=context):
```
hbrunn: are you sure this does what you expect?

hbrunn:
have a look at travis

openupgradelib/openupgrade.py
blaggacao: yep, this is how my local fork has been working for some time now.

hbrunn: I didn't ask if it breaks anything, it obviously doesn't. The question is: What do you expect from returning a function that takes parameters which will never be passed and always fall back to their default value, which you set from names available in the closure anyways?

openupgradelib/openupgrade.py
blaggacao: - "useful" as in "serve a current use case better"? -- As to my knowledge, it isn't.

or

- "useful" as in "potentially enable future use cases"? -- This was the design goal.


Class of future use cases, which this facilitates, are characterized by the need to determine and persist the exact migration context, whereby "exact migration context" is constituted by a hash of `migrate_version` and `module_name`. Only if those are not `None`, this promise can be kept, but I see no other way to strengthen the promise further, so user has to take care of this.

Basically, this enables tagging records or values or columns/fields with the hash and refer to them precisely afterwards.

Some illustrative examples:
- External ids created by migration scripts
- Reliable Field/column prefixing

blaggacao:
Travis all green, but 2.6. I cannot make sense of the error. Can you have a quick look? https://travis-ci.org/OCA/openupgradelib/jobs/151131493

openupgradelib/openupgrade.py
hbrunn: I propose you come up with a PR on OpenUpgrade where such a thing is used. For the time being, I strongly disagree with cluttering this lib with stuff that might be used some time somewhere.

This goes for most of your input up to now: I personally won't even have a look at huge change proposals that are not grounded in some already existing PR in OpenUpgrade.

openupgradelib/openupgrade.py
blaggacao: Before, test where failing. This is what I was referring to by it has been working on my local machine: "it works now". IIRC, by trial and error from my previous PR about env support, this is necessary. What alternative explanation would you draw from: https://travis-ci.org/OCA/openupgradelib/jobs/151125711

hbrunn: You can read global variables, but as soon as you want to assign a value, creating the local one will clash with the global one. You circumvent this by creating a local variable with the same name as the global one in the function definition already.
The `use_env` parameter has the same problem, which is why I used `use_env2` for assignments.

openupgradelib/openupgrade.py
blaggacao: It's entirely the maintainers' call. I have seen disagreement among maintainers about the single-purposeness of openupgradelib already in the past. This is definitely a question of "product strategy" - not my business, unless someone would ask me for my opinion.

openupgradelib/openupgrade.py
blaggacao: Ahh. Some enlightenment comes to my mind. Thanks for the explanation! :+1: 
So you would rather say:
Not doing reassignment in the function signature, but avoid naming clashes?

hbrunn: that depends. If you actually want to change the variable from the other scope, use `var $yourvariablename`. If not, declare another name. For mutable types like the context, a copy is basically mandatory.
In your proposal (which I disagree with as detailed below), passing `context=dict(context or {}, yourkey1=value1, ...)` would seem the best choice to me

openupgradelib/openupgrade.py
blaggacao: As a side note: The last time I proposed interdependent changes, it resulted in days wasted in useless refactoring, as anyhow no duly consensus could be drawn and my own fork and upstream code were terribly torn apart and almost impossible to integrate. Yet, completely walking away is not yet warranted and not a real option, anyhow... :smile:

pedrobaeza:
Please change one thing per PR. Make a PR for changing docstring, which will be quickly merged, and others for each option to be discussed separately.

PR #57, opened 2016-10-07
[RFR] Move column_exists so it can be used during loading

StefanRijnhart:
Needed for https://github.com/OCA/OpenUpgrade/pull/640

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -241,6 +241,7 @@ def load_data(cr, module_name, filename, idref=None, mode='init'):
 # for backwards compatibility
 load_xml = load_data
 table_exists = openupgrade_tools.table_exists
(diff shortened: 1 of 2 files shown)


legalsylvain:
Thanks. :+1: 
Merging this trivial PR.

pedrobaeza:
I'm afraid it's not so trivial, because there are a lot of scripts that use this method, so now they need to change the import. Please add an alias inside openupgrade.py for not losing compatibility.

legalsylvain:
@pedrobaeza  There is an alias in this PR.
https://github.com/OCA/openupgradelib/pull/57/files#diff-3d7675eb1386de028dab1f879bdc4681R244

pedrobaeza:
Sorry, I didn't see it. Then it's good.

PR #59, opened 2016-10-11
[FIX] make openupgradelib 1.2.0 work with openupgrade 8

hbrunn:


pedrobaeza:
Aren't we increasing a lot the test time with this?

hbrunn:
yes

pedrobaeza:
Should we then take another approach?

hbrunn:
well, we've seen that none of us noticed the problem the latest commit caused, and this is what tests are meant for. I'm now working on make this specific test only run once, but once it should run in my opinion

pedrobaeza:
OK, it's fair to increase the time but avoid these things. We should improve drastically the coverage of the lib also.

StefanRijnhart:
Yes, sorry about this. Thanks for working on tests.

hbrunn:
@pedrobaeza @StefanRijnhart yay!
I can't squash myself as I don't have access to my ssh keys at the moment. But you could squash and merge right away.
Something that really puzzles me is that the check for `TRAVIS_PYTHON_VERSION` doesn't work for the pypy case, that's why I had to introduce the second check in https://github.com/OCA/openupgradelib/pull/59/files#diff-f1aa76070daee9c8fad189a76355176bR34 - do imports behave differently in pypy? I think this is mergeable as is, even though it's not beautiful, but if anyone can shed light on what's going on here, I'd learn a valuable lesson I guess

StefanRijnhart:
Thanks for the hard work! :+1

StefanRijnhart:
No clue about pypy BTW., sorry.

pedrobaeza:
Me neither I'm afraid

StefanRijnhart:
I just released version 1.2.1 so that the default Pypi version of Openupgradelib works with Odoo 8.0 again.

PR #60, opened 2016-10-11
[FIX] openupgrade is loaded earlier than openerp.modules

StefanRijnhart:
Fixes https://travis-ci.org/OCA/OpenUpgrade/builds/166773305


hbrunn:
my take on this was https://github.com/OCA/openupgradelib/pull/59 and the rest is noise because I'd like to have this tested too

StefanRijnhart:
Closing in favour of #59

PR #62, opened 2016-12-01
[FIX] installing a library must not fix requirements versions

sbidoul:
Installing a library must not fix requirements versions, unless strictly necessary

In this case, the problem is that installing openupgradelib would
downgrade the wheel package, which is undesirable.


sbidoul:
I suggest removing requirements.txt too, unless there is a good reason for openupgradelib to depend on a specific version of the wheel package

PR #68, opened 2017-01-24
[ADD] new logging decorator

legalsylvain:
Add a new decorator that can be added on each function that take times. (or for debug / development purpose)

**simple log**

        @openupgrade.logging()
        def my_function(*args):

**log every x calls**

        @openupgrade.logging(logging_step=x)
        def my_function_called_many_times(*args)

Try to fix : https://github.com/OCA/OpenUpgrade/issues/326

Regards.

CC : @pedrobaeza, @StefanRijnhart, @sebastienbeau, @OCA/openupgrade-maintainers

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,14 +3,6 @@
 History
 -------
 
-1.2.3 (2017-01-24)
-------------------
-* [ADD] Add New decorator 'logging' to log function calls
-* [RFR] Local logger; don't force debug level
-* [IMP] docstring for 'rename_models'
-* [FIX] 'update_module_names' : Rename non updated XML-ID occurences
-* [ADD] add 'convert_binary_field_to_attachment' (8.0 -> 9.0)
-
 1.2.2 (2016-12-27)
 ------------------
 * New argument merge_modules in update_module_names for merging several
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,7 +5,11 @@ History
 
 1.2.3 (2017-01-24)
 ------------------
-* New decorator to log function calls.
+* [ADD] Add New decorator 'logging' to log function calls
+* [RFR] Local logger; don't force debug level
+* [IMP] docstring for 'rename_models'
+* [FIX] 'update_module_names' : Rename non updated XML-ID occurences
+* [ADD] add 'convert_binary_field_to_attachment' (8.0 -> 9.0)
 
 1.2.2 (2016-12-27)
 ------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,10 @@
 History
 -------
 
+1.2.3 (2017-01-24)
+------------------
+* New decorator to log function calls.
+
 1.2.2 (2016-12-27)
 ------------------
 * New argument merge_modules in update_module_names for merging several
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1056,7 +1056,7 @@ def reactivate_workflow_transitions(cr, transition_conditions):
 openupgrade_call_logging = {}
 
 
-def logging(args_details=False, logging_step=False):
+def logging(args_details=False, step=False):
     """
     This is a decorator for any sub functions called in an OpenUpgrade script.
     (pre or post migration script)
@@ -1067,7 +1067,7 @@ def logging(args_details=False, logging_step=False):
     is called.
 
     :param args_details: if True, arguments details are given in the log
-    :param logging_step: The log will be done only every logging_step times.
+    :param step: The log will be done only every step times.
 
     Typical use::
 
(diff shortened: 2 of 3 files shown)


pedrobaeza:
I already did something similar here: https://github.com/OCA/openupgradelib/pull/1. Please check it for possible completeness.

legalsylvain:
@pedrobaeza. Thanks for pointing that closed PR I have not seen. 

I added args_details feature from your PR and increment version, and complete history.

Regards.

hbrunn:
to my taste, OpenUpgrade already is too chatty. Especially with lavish use of `logged_query`. Don't you think this will just add more noise to the log file?

legalsylvain:
Hi @hbrunn. Thanks for your review.

Indeed, I do not read a lot the result of logged_query and it generates noises for me. But I guess, it can be disabled, setting openupgrade logger level to info. (logged_query generated debug log).

About that new decorator :  
- this is a little noise. ;-) less noice than logs generated by logged_query.
- I really think that it should not be used for each function in openupgrade, only for the function / module that could take times. (stock + account in V8 for exemple)
- For me, it's a mixed tools, to be included in openupgrade migration scripts of course, but for developpers too. If a developper has any problem during the pre or post migration, it can easier to identify which function take times, just adding some decorators. Writing migration script is a mess and need a lot of works, any extra tools is welcome.

If the noise still is a problem, I can set the level of the log to debug.

pedrobaeza:
I feel the same like @legalsylvain. OpenUpgrade is too verbose in several things that maybe shouldn't be, but very silent in another ones that need to be debugged or tracked. With this, you have the possibility to add some basic logging at the method call, but that's not mean to be done everywhere. This can even be done by someone on a concrete installation modifying codebase for tracing an error or times. I know that you can add a logging call, but this is more direct.

I developed the original proposal for the same purpose than Sylvain: stock post-migration is very slow and there are at least 8 methods without no feedback.

The progress bar was the other technique for estimating times in the quant creation: you have the total number of moves, so it's easy to estimate after some steps in the process which is the remaining time. And this one is very useful for entire OpenUpgrade. @legalsylvain, maybe you can find another similar lib with Python3 support for having the same possibilities?

legalsylvain:
Hi @pedrobaeza, thanks for giving your point of view.

About progressbar, indeed, it should be great, but it ask more work (finding and implementing a stable lib) so it's out of the scope of this PR. I'm fighting with quants for the time being !
Morever, it could be implemented later.

Do you agree to accept this PR as a first step ?

kind regards.

HISTORY.rst
```diff
@@ -3,6 +3,10 @@
 History
 -------
 
+1.2.3 (2017-01-24)
+------------------
+* New decorator to log function calls.
```
pedrobaeza: You should see the commit log for other changes that have been included meanwhile

bealdav:
:+1:

openupgradelib/openupgrade.py
```diff
@@ -1051,6 +1052,77 @@ def reactivate_workflow_transitions(cr, transition_conditions):
...
+                # Compute unique name
+                unique_name = '%s.%s' % (func.__module__, func.__name__)
+                if unique_name not in openupgrade_call_logging:
+                    openupgrade_call_logging[unique_name] = 0
+                openupgrade_call_logging[unique_name] += 1
+                current = openupgrade_call_logging[unique_name]
+                if current == 1 or current % logging_step == 0:
+                    msg += " ; Calls quantity : %d" % current
+                    if current == 1:
+                        msg += " ; Logging Step : %d" % logging_step
```
StefanRijnhart: I would slightly prefer that logging_step be renamed to step.

openupgradelib/openupgrade.py
```diff
@@ -1051,6 +1052,77 @@ def reactivate_workflow_transitions(cr, transition_conditions):
...
+                else:
+                    to_log = False
+
+            # Log Args
+            if args_details:
```
StefanRijnhart: Can you only perform these string operations if to_log evaluates true?

HISTORY.rst
```diff
@@ -3,6 +3,14 @@
 History
 -------
 
+1.2.3 (2017-01-24)
```
StefanRijnhart: Can you remove all this unrelated release stuff? By convention it is committed directly on the branch and that commit is than tagged. I think merging such a commit might be messy.

legalsylvain:
Hi @StefanRijnhart,

Thanks for your review ! Sorry for the unrelated history and version changes. I reverted.
(First time I commit in a lib. I have no clear ideas about how versions should work).

regards.

mourad-ehm:
:+1:

legalsylvain:
Hi @StefanRijnhart , is it now ok for you ? 
regards.

hbrunn:
ping @StefanRijnhart

StefanRijnhart:
@pedrobaeza is this OK for you as well?

pedrobaeza:
Yes, we can proceed merging this and letting the other features as wishlist.

PR #71, opened 2017-05-16
[FIX] m2o_to_x2m: Compatible with Odoo v10

pedrobaeza:
* _columns doesn't exist on v10
* _sql_names dissapears also
* On >= 8.0, env is available for getting classes objects


hbrunn:
@pedrobaeza why work in progress?

pedrobaeza:
Because it was in test in https://github.com/OCA/OpenUpgrade/pull/828. I haven't checked results yet.

pedrobaeza:
Now this is tested and working, so I'm going to merge it.

PR #74, opened 2017-07-04
Added suggestion for latest version install in docs

PCatinean:


PCatinean:
That was my initial thought as well, I then left it so people know it's on pip as well but either is fine as long as the information is specified imo.

hbrunn:
i don't care one way or the other to be honest

PR #75, opened 2017-07-31
[FIX] when renaming/deleting a module, rename/delete its xmlid

hbrunn:
fixes OCA/OpenUpgrade#934

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -754,6 +754,12 @@ def update_module_names(cr, namespec, merge_modules=False):
         if merge_modules:
             query = "DELETE FROM ir_module_module WHERE name = %s"
             logged_query(cr, query, [old_name])
+            logged_query(
+                cr,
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -754,6 +754,23 @@ def update_module_names(cr, namespec, merge_modules=False):
...
+            logged_query(
+                cr,
+                "DELETE FROM ir_model_data WHERE module = 'base' "
+                "AND model='ir.module.module' AND name = %s",
+                ('module_%s' % old_name,),
+            )
+        else:
+            logged_query(
+                cr,
+                "UPDATE ir_model_data SET name = %s "
```
pedrobaeza: This is already done in https://github.com/OCA/openupgradelib/pull/75/files#diff-3d7675eb1386de028dab1f879bdc4681R730

hbrunn: indeed, thanks. Done & rebased

PR #79, opened 2017-09-18
[FIX][set_defaults] New api adaptation

mreficent:
This PR is to adapt `set_defaults` method to new api.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -592,6 +592,7 @@ def set_defaults(cr, pool, default_spec, force=False, use_orm=False):
     Set default value. Useful for fields that are newly required. Uses orm, so
     call from the post script.
 
+    :param pool: In v10 and newer, you have to pass the 'env' instead.
     :param default_spec: a hash with model names as keys. Values are lists \
     of tuples (field, value). None as a value has a special meaning: it \
     assigns the default value. If this value is provided by a function, the \
@@ -613,12 +614,19 @@ def set_defaults(cr, pool, default_spec, force=False, use_orm=False):
         if use_orm:
             for res_id in ids:
                 # Iterating over ids here as a workaround for lp:1131653
-                obj.write(cr, SUPERUSER_ID, [res_id], {field: value})
+                if isinstance(pool, api.Environment):
+                    obj.write({field: value})
+                else:
+                    obj.write(cr, SUPERUSER_ID, [res_id], {field: value})
         else:
             query, params = "UPDATE %s SET %s = %%s WHERE id IN %%s" % (
                 obj._table, field), (value, tuple(ids))
             # handle fields inherited from somewhere else
-            if field not in obj._columns:
+            if version_info[0] >= 10:
+                columns = obj._fields
+            else:
+                columns = obj._columns
+            if field not in columns:
                 query, params = None, None
                 for model_name in obj._inherits:
                     if obj._inherit_fields[field][0] != model_name:
@@ -643,13 +651,16 @@ def set_defaults(cr, pool, default_spec, force=False, use_orm=False):
                 cr.execute(query, (params[0], sub_ids))
 
     for model in default_spec.keys():
-        obj = pool.get(model)
-        if not obj:
+        obj = pool.get(model, False)
+        if obj == False:
             do_raise(
                 "Migration: error setting default, no such model: %s" % model)
         for field, value in default_spec[model]:
             domain = not force and [(field, '=', False)] or []
-            ids = obj.search(cr, SUPERUSER_ID, domain)
+            if isinstance(pool, api.Environment):
+                ids = obj.search(domain).ids
+            else:
+                ids = obj.search(cr, SUPERUSER_ID, domain)
             if not ids:
                 continue
             if value is None:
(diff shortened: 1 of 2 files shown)


openupgradelib/openupgrade.py
```diff
@@ -643,13 +644,19 @@ def write_value(ids, field, value):
                 cr.execute(query, (params[0], sub_ids))
 
     for model in default_spec.keys():
-        obj = pool.get(model)
-        if not obj:
-            do_raise(
-                "Migration: error setting default, no such model: %s" % model)
+        if version_info[0] >= 10:
```
pedrobaeza: No need for this `if`. It's the same for all versions: `pool/env.get(model)`

hbrunn: I beg to differ with the estimated colleague @pedrobaeza. Only since 10, `Environment` inherits from `Mapping` which implements the `.get` function. In <= 9, this will raise.
As 8 doesn't even support the `in` operator (https://github.com/OCA/OCB/blob/8.0/openerp/api.py#L779), we're really out of luck here.

I think btw. it's not a good idea anymore to check for the version - we still can request a registry if we want or an env in pre 10: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L1152 Let's better check if `pool` is an `api.Envrionment`, then you can first check `model in pool.registry` and then use the itemgetter on `pool` itself

pedrobaeza:
@hbrunn You can make `pool.get('model')` on <= v9. See for example:

https://github.com/odoo/odoo/blob/69fa111e95f1697a007aae8be32fe51188789da2/addons/stock_account/models/stock.py#L220

Any way, the strategy can be what you suggest.

mreficent:
A new problem is the line `if not obj:` that becomes true in v10 :sweat:

mreficent:
What do you think now?

openupgradelib/openupgrade.py
```diff
@@ -643,13 +644,16 @@ def write_value(ids, field, value):
...
-        if not obj:
+        obj = pool.get(model, False)
+        if obj == False:
             do_raise(
                 "Migration: error setting default, no such model: %s" % model)
         for field, value in default_spec[model]:
             domain = not force and [(field, '=', False)] or []
-            ids = obj.search(cr, SUPERUSER_ID, domain)
+            if isinstance(pool, api.Environment):
+                ids = obj.search(domain).ids
```
hbrunn: won't all hell break lose later when `write_value` tries to do old style calls again?

openupgradelib/openupgrade.py
```diff
@@ -643,13 +644,16 @@ def write_value(ids, field, value):
                 cr.execute(query, (params[0], sub_ids))
 
     for model in default_spec.keys():
-        obj = pool.get(model)
-        if not obj:
+        obj = pool.get(model, False)
```
hbrunn: I just checked with

```shell
grep -R use_env addons/*/migrations --files-with-matches | xargs grep --after-context=2 set_defaults
addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py:    openupgrade.set_defaults(env.cr, env.registry, column_defaults)
addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py-    update_barcodes_nomenclatures(env)
addons/point_of_sale/migrations/9.0.1.0.1/post-migration.py-    openupgrade.load_data(
--
addons/website/migrations/9.0.1.0/post-migration.py:    openupgrade.set_defaults(
addons/website/migrations/9.0.1.0/post-migration.py-        env.cr, pool, {'website': [('cdn_filters', None)]}, use_orm=True,
addons/website/migrations/9.0.1.0/post-migration.py-    )
```

that this will not break in the <= 9.0, but that's only coincidence. But on the other hand, we can also just see it as mistake to pass an env in <= 10

openupgradelib/openupgrade.py
mreficent: Yes, the same with _defaults. I didn't tested that yet, but sure it will break.

openupgradelib/openupgrade.py
mreficent: Why is a mistake to pass an env? I don't understand.

openupgradelib/openupgrade.py
mreficent: Ok, now I fixed the _defaults, _columns, and the write_value.

What do you think?

mreficent:
Why Travis fails?

Please, update review 🙂

mreficent:
The master branch is already broken

openupgradelib/openupgrade.py
```diff
@@ -592,6 +592,7 @@ def set_defaults(cr, pool, default_spec, force=False, use_orm=False):
     Set default value. Useful for fields that are newly required. Uses orm, so
     call from the post script.
 
+    :param pool: In v10 and newers, you have to pass the 'env' instead.
```
StefanRijnhart: `newer`

openupgradelib/openupgrade.py
StefanRijnhart: Fine by me to suppose that pool is not an env before 10.0, especially as it is documented above.

mreficent:
Yes, thanks, now is :green_apple: !!
update review please

pedrobaeza:
@hbrunn can you update your review?

StefanRijnhart:
@pedrobaeza I'm guessing that @hbrunn is too busy to review right now. I believe his points are sufficiently addressed, so if we can have your approval I think it is alright to merge.

pedrobaeza:
Merging then.

PR #80, opened 2017-10-03
[ADD] support update mode where we never try to create records

hbrunn:
this is meant to allow loading `noupdate="1"` records and being sure nothing bad happens when doing this.

Currently, if you `load_data` with records that were noupdate, forcecreate=False and deleted, odoo will generate a `create`, which fails if there are any required fields that are not included in the xml

cc @pedrobaeza

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -24,6 +24,10 @@ import os
 import inspect
 import uuid
 import logging
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
 from contextlib import contextmanager
 try:
     from contextlib import ExitStack
@@ -46,6 +50,7 @@ except ImportError:
                 self._cms.pop().__exit__(exc_type, exc_value, traceback)
 
 from psycopg2.extensions import AsIs
+from lxml import etree
 from . import openupgrade_tools
 
 core = None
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
Can you introduce a new value for mode, so that no confusion arises with the original modes from the Odoo method (I'm thinking `init_no_create`). In fact, I am already confused. What happens if you want to update a record that has the noupdate flag in ir_model_data? It looks like the record does not get updated. Maybe you need to pass mode='init' to the upstream method.

Can you move that phrase 'will not try to create a record with partial data in case the record was removed in the database.' to the upper method docstring, so that it shows up in the documentation?

hbrunn:
if you update a record with the noupdate flag, it depends on the mode what happens: nothing in mode=update, an update in init. Or a create if the record doesn't exist, and that's what breaks here.

openupgradelib/openupgrade.py
```diff
@@ -267,11 +274,45 @@ def load_data(cr, module_name, filename, idref=None, mode='init'):
                 cr, module_name, pathname, fp.read(), idref, mode, noupdate)
         elif ext == '.yml':
             yaml_import(cr, module_name, fp, None, idref=idref, mode=mode)
+        elif mode == 'init_no_create':
+            for fp2 in _get_existing_records(cr, fp, module_name):
+                tools.convert_xml_import(
+                    cr, module_name, fp2, idref, mode=mode
```
pedrobaeza: Change this to `init`. Please rebase also on latest branch for seeing Travis result.

hbrunn: good one. done

StefanRijnhart:
CI is still broken. I'm restarting Travis.

pedrobaeza:
Yes, sorry, I'm in the travel back with limited access. Holger, squash at your taste and please merge.

StefanRijnhart:
Thanks @pedrobaeza, have a good journey!

PR #84, opened 2017-11-05
[ADD] rename_fields: New method for renaming completely fields

pedrobaeza:
This, in contrast of ``rename_columns``, performs all the steps for completely rename a field from one name to another. This is needed for making a complete renaming of a field with all their side features:
translations, filters, exports...

Call this method whenever you are not performing a pure SQL column renaming for other purposes (preserve a value for example).

This method performs also the SQL column renaming, so only one call is needed.

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -132,6 +132,7 @@ __all__ = [
     'load_data',
     'copy_columns',
     'rename_columns',
+    'rename_fields',
     'rename_tables',
     'rename_models',
     'rename_xmlids',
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This will be soon used in a PR changing all the calls in OpenUpgrade 9 and 10.

pedrobaeza:
Used in https://github.com/OCA/OpenUpgrade/pull/1159

openupgradelib/openupgrade.py
```diff
@@ -380,6 +381,89 @@ def rename_columns(cr, column_spec):
...
+            SET name = %s
+            WHERE name = %s
+            """, (old_field, new_field),
+        )
+        # Rename appearances on filters
+        # Example of replaced domain: [['field', '=', self], ...]
+        # TODO: Rename when the field is part of a submodel (ex. m2one.field)
+        cr.execute("""
+            UPDATE ir_filters
+            SET domain = replace(domain, $$'%s'$$, $$'%s'$$)
```
Yajo: You could use `\m%s\M` to match whole words, because in case there's a `id` field, we don't want `partner_id` to match!

Check it out: https://www.postgresql.org/docs/current/static/functions-matching.html#posix-constraint-escapes-table

pedrobaeza:
@mreficent if you want to do it... :wink: 

They are put as TODO as they are very hard to be done, and well, we didn't have it previously, so it's not something critical.

openupgradelib/openupgrade.py
pedrobaeza: It's not the case for this, @Yajo . See that `'` are surrounding the word.

Yajo: It's expected that using my fix works for full words, no matter if they are surrounded by `'`, `"` or found in a relational field such as `parent_id.matching_field`. Isn't that what you'd want here? 🤔

pedrobaeza: NO, in this case it's a simple substitution. See that we are matching simple domains for the same model, so that's exactly what we want to match. 

The rest is not possible for multiple reasons:

* The model of the filter will be another.
* We have to determine if the submodel is for the model with the replaced field, and not another with a field with the same old name.

That's why the TODO is there, because it can't be resolved simply with a regex.

pedrobaeza:
Merging this and OpenUpgrade PRs for continuing proposing other features.

mreficent:
One silly question:

Will the `Model name` and the `Table name` always be the same? If the answer is yes, then it should be more simple to just pass one of those parameters, for example the `Table name`, and then obtain the `Model name` by just using:

       model_name = table_name.replace("_", ".")

pedrobaeza:
No, you need both, as you can redefine the table name in your model. That's the default value.

PR #85, opened 2017-11-16
[IMP] rename_fields: Several improvements

pedrobaeza:
Split by commits for better comprehension:

* Include mail.alias table
* Restrict update to matching records in existing replaces
* Include col_group_by key in the filters contexts.

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -456,7 +456,8 @@ def rename_fields(env, field_spec, no_deep=False):
         # {'group_by': ['field', 'other_field'], 'other_key':value}
         # {'group_by': ['date_field:month']}
         # {'other_key': value, 'group_by': ['other_field', 'field']}
-        cr.execute("""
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -444,24 +444,49 @@ def rename_fields(env, field_spec, no_deep=False):
...
-                $$'group_by':(.*)'%s(:day|:week|:month|:year){0,1}'(.*?\])$$,
-                $$'group_by':\1'%s\2'\3$$
+                context, %(old_pattern)s, %(new_pattern)s
             )
             WHERE model_id = %%s
-            """ % (old_field, new_field), (model, ),
+                AND context ~ %(old_pattern)s
+            """ % {
+                'old_pattern': (
+                    "$$('group_by'|'col_group_by'):([^\]]*)"
```
Yajo: Don't you need raw strings?

pedrobaeza: Yeah, changed!

PR #86, opened 2017-11-23
[ADD] Replace account types in Odoo 9.0

StefanRijnhart:
Used in https://github.com/OCA/OpenUpgrade/pull/1203

### DIFF
--- a/openupgradelib/openupgrade_90.py
+++ b/openupgradelib/openupgrade_90.py
@@ -5,9 +5,12 @@
 """This module provides simple tools for OpenUpgrade migration, specific for
 the 8.0 -> 9.0 migration.
 """
+import logging
+import threading
+
+from psycopg2.extensions import AsIs
 
 from openupgradelib import openupgrade
-import logging
 
 
 def convert_binary_field_to_attachment(env, field_spec):
@@ -30,7 +33,7 @@ def convert_binary_field_to_attachment(env, field_spec):
     :param env: Odoo environment
     :param field_spec: A dictionary with the ORM model name as key, and as
         dictionary values a tuple with:
-        
+
         * field name to be converted as attachment as first element.
         * SQL column name that contains actual data as second element. If
           the second element is None, then the column name is taken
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_90.py
```diff
@@ -66,3 +68,29 @@ def convert_binary_field_to_attachment(env, field_spec):
...
+
+def replace_account_types(env, type_spec, unlink=True):
+    """ Replace old account types with their replacements.
+    :param type_spec: list of tuples (xmlid of old account.account.type, \
+xmlid of new account.account.type)
+    :param unlink: attempt to unlink the old account type
+    """
+    logger = logging.getLogger('OpenUpgrade')
+    for old_type, new_type in type_spec:
+        type8 = env.ref(old_type)
```
pedrobaeza: Maybe one of these types has been deleted on the target DB, so you should make:
```
type8 = env.ref(old_type, False)
if not type8:
    continue
```

StefanRijnhart: I prefer not to allow for this. The account types should exist because they are not defined with FORCECREATE=0. If they don't exist the user can upgrade the modules on the source version of the migration first. If we don't allow for non-existing types there is an implicit check on the correct spelling of the xmlids.

pedrobaeza: But this is a general utility. You can use this method for custom or other uses than the one you have thought, so it's better to not block migrations by this detail. Put a warning if you want to be verbose, but don't make this to fail.

StefanRijnhart: No, I want it to fail. Leave it up to the caller to only pass valid account types. For the record, are there even any account types in all of the Odoo localization modules that are defined with forcecreate=false?

StefanRijnhart: Answering my own question: no there are no account types defined with forcecreate=false in the Odoo localization modules.

pedrobaeza: I'm thinking in the case you haven't updated previous version before starting migration (that's not a requirement till now), and you removed one of the types.

pedrobaeza:
@StefanRijnhart can you check my last comment to finish the PR?

openupgradelib/openupgrade_90.py
StefanRijnhart: You are right that it is not a formal requirement at the moment. Do you think it is reasonable to add it as a requirement? It is not just for this particular case. There may be other cases which would make it easier for the developers if module data can be considered consistent.

pedrobaeza: I would prefer to avoid it for this special case, as the check code is very simple, and till now I haven't found any other thing that requires to be fully updated, but we can include at least a reference on documentation.

StefanRijnhart: Then I'll modify the code here and only raise for missing records if the testing flag is present on the thread.

pedrobaeza:
Everything in place now?

StefanRijnhart:
@pedrobaeza as far as I am concerned, yes!

PR #88, opened 2017-11-27
[ADD] disable_invalid_filters

pedrobaeza:
Avoid errors on migrated databases making a non-destructive disabling.

It analyzes all the existing active filters to check if they are still correct. If not, they are disabled for avoiding errors when clicking on them, or worse, if they are default filters when opening the model/action.

To be run at the base end-migration script for having a general scope. Only assured to work on >= v8.

When merged, I will make PRs for OpenUpgrade versions calling this method on end-migration/deferred step.

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -161,6 +161,7 @@ __all__ = [
     'lift_constraints',
     'rename_property',
     'delete_record_translations',
+    'disable_invalid_filters',
 ]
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can we merge then?

pedrobaeza:
There's a conflict now due to a merge operation.

StefanRijnhart:
Well I started the merge operation because there was a conflict in the first place.

StefanRijnhart:
But I don't know what went wrong. Maybe you can rebase your own commit onto master and discard my edits?

pedrobaeza:
Done!

openupgradelib/openupgrade.py
```diff
@@ -1670,6 +1671,7 @@ def rename_property(cr, model, old_name, new_name):
         "update ir_property set name=%s where fields_id in %s",
         (new_name, field_ids))
 
+<<<<<<< HEAD
```
StefanRijnhart: Conflict markers were merged.

pedrobaeza:
Resubmitted.

StefanRijnhart:
Success!

PR #91, opened 2017-12-14
Fix logging with non-ascii characters in exception.

### LINKED ISSUE #90: Crash in error logging with UTF-8 message
str(e) will crash if there is special characters in the exception message (because I am in fr_CH.UTF8 with my user on my system I think). Then the error that occurs is not displayed in the log. I will try to run the process with another temporary session locale but this is not very clean.

  File "/[…]/9.0/server/openerp/service/server.py", line 897, in preload_registries
    registry = RegistryManager.new(dbname, update_module=update_module)
  File "/[…]/9.0/server/openerp/modules/registry.py", line 390, in new
    openerp.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/[…]/9.0/server/openerp/modules/loading.py", line 342, in load_modules
    loaded_modules, processed_modules = load_module_graph(cr, graph, status, perform_checks=update_module, report=report, upg_registry=upg_registry)
  File "/[…]/9.0/server/openerp/modules/loading.py", line 189, in load_module_graph
    migrations.migrate_module(package, 'post')
  File "/[…]/9.0/server/openerp/modules/migration.py", line 169, in migrate_module
    mod.migrate(self.cr, pkg.installed_version)
  File "/[…]/openupgradelib/openupgradelib/openupgrade.py", line 1391, in wrapped_function
    message = str(e)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xab' in position 69: ordinal not in range(128)

hmartinet:
str() should never be used in Python 2 if you can not know the encoding
of the string.

In this case it can log and raise a UnicodeEncodeError in migration.log
instead of the exception causing the error, and you cannot know why
your migration failed.

I use repr() for Python 2. It's not perfect but the exception is
displayed and the stacktrace is in the log because of logger.exception(),
which is bypassed if str() cause another exception.

Resolve #90

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1390,9 +1390,7 @@ def migrate(no_version=False, use_env=None, uid=None, context=None):
                             cr, uid or SUPERUSER_ID, context or {})
                         if use_env2 else cr, version)
                 except Exception as e:
-                    message = str(e)
-                    if sys.version_info[0] == 2:
-                        message = message.decode('utf8')
+                    message = repr(e) if sys.version_info[0] == 2 else str(e)
                     logger.error(
                         "%s: error in migration script %s: %s",
                         module, filename, message)


hmartinet:
@StefanRijnhart Yes it's not really my solutions it's a known trick in Python 2 community. :wink:

Thanks for the quick merge !

PR #93, opened 2018-01-10
[RFC] rename field xmlids when renaming models

hbrunn:
so every field has an xmlid of the form `field_$modelname_$fieldname` where in the model name, periods are replaced by underscores. When renaming a model, we have to update this too, because Odoo only created those xmlids when *creating* a column, and our columns already exist: https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L428

Now, the code below works fine for my use case, but imagine renaming a model whose name is a substring of other existing models (we want to rename `res.partner`, but there's also `res.partner.category` for example), then the below blows up. Any ideas? A join over ir_model_fields, ir_model is the only thing I can think of, so any better ideas?

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -532,12 +532,28 @@ def rename_models(cr, model_spec):
     for (old, new) in model_spec:
         logger.info("model %s: renaming to %s",
                     old, new)
+        _old = old.replace('.', '_')
+        _new = new.replace('.', '_')
         cr.execute('UPDATE ir_model SET model = %s '
                    'WHERE model = %s', (new, old,))
         cr.execute('UPDATE ir_model_fields SET relation = %s '
                    'WHERE relation = %s', (new, old,))
         cr.execute('UPDATE ir_model_data SET model = %s '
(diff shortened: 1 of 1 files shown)


pedrobaeza:
The only way I see is to go through `ir_model_fields` table for the old model and perform exact queries. I think this can be done in an

pedrobaeza:
Updated previous comment (sent by mistake) with the proper SQL.

hbrunn:
thanks Pedro, I'll use this when I work on this specific upgrade again and come back

pedrobaeza:
Have you used this, Holger?

hbrunn:
first time I encountered this I fixed it manually and wrote this RFC. Next time I'll run across this I'll fix it. We have a few migration projects in the pipeline, so this probably happens rather sooner than later

pedrobaeza:
OK, then I'm going to add a commit with my proposal here because I'm migrating right now one DB and I can try also, it it's OK to you.

hbrunn:
of course!

pedrobaeza:
FYI, it has worked flawlessly

PR #94, opened 2018-01-15
Disable filters with invalid context.

mvaled:
I've found SyntaxError in the context of custom filters:

```
2018-01-15 17:24:42,154 29768 WARNING db1 openupgrade: FILTER DISABLED: Filter 'xxxxx for user 'yyyyy for model 'purchase.requisition' has been disabled as it contains an invalid context {'default_user_id': 5, ['account_analytic_id']}.
```

This commit invalidates them.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1742,7 +1742,15 @@ def disable_invalid_filters(env):
             f.active = False
             continue
         # CONTEXT GROUP BY
-        context = safe_eval(f.context, {'time': time, 'uid': env.uid})
+        try:
+            context = safe_eval(f.context, {'time': time, 'uid': env.uid})
+        except Exception:
+            logger.warning(
+                format_message(f) + "as it contains an invalid context %s.",
+                f.context
+            )
+            f.active = False
+            continue
         keys = ['group_by', 'col_group_by']
         for key in keys:
             if not context.get(key):


pedrobaeza:
But this is because you don't have your DB in previous version well maintained. For sure if you open that filter in that version, you will find an error. This can't be bullet proof with all the broken data in previous version. At least, that's my opinion.

Fix it in that version, and then start the migration.

mvaled:
Hi @pedrobaeza,

While I see your point, I would like you to consider another argument: Migrating a DB is a rather lengthy process.  It could take hours each time, and whenever it fails you would probably have to start all over.  Also, the only reasonable solution for such broken data in the filter is also to disable it and then ask the user their intentions for the filter so that we can reasonably replace it for a good one.

I still don't know how does SyntaxError ever get there.

pedrobaeza:
Just a note: you don't need to restart the migration. Just launch again the server and it will continue updating - migrating - remaining modules.

I let other colleagues to decide to relax the rule I have stated.

mvaled:
While developing the migration procedure I do continue with the migration at that point.  However, I keep a log of things to be done, so that when the *true* migration happens I perform all steps needed to ensure a good out-come.

Maybe I'm too cautious...

mvaled:
:fireworks: :smile:

Issue #95, opened 2018-02-11
[6.1->7.0] error when running the script due to new api adaptation


bodi000:
Hello,

This appears to be a regression from commit 22a6409, as I did a few upgrades last year without the error. 
The line
https://github.com/OCA/openupgradelib/blob/9b5d8f9710b19339438fcd62e3f98c631086a8cf/openupgradelib/openupgrade.py#L811

produces this error:

```
2018-02-11 10:27:31,701 15701 INFO bs_05_migrated OpenUpgrade: base: post-migration script called with version 6.1.1.3
2018-02-11 10:27:31,701 15701 ERROR bs_05_migrated OpenUpgrade: base: error in migration script base/migrations/7.0.1.3/post-migration.py: TypeError('get() takes exactly 2 arguments (3 given)',)
2018-02-11 10:27:31,702 15701 ERROR bs_05_migrated OpenUpgrade: get() takes exactly 2 arguments (3 given)
Traceback (most recent call last):
  File "/var/tmp/openupgrade/openupgradelib/openupgradelib/openupgrade.py", line 1391, in wrapped_function
    if use_env2 else cr, version)
  File "/var/tmp/openupgrade/7.0/server/openerp/addons/base/migrations/7.0.1.3/post-migration.py", line 514, in migrate
    openupgrade.set_defaults(cr, pool, force_defaults, force=True)
  File "/var/tmp/openupgrade/openupgradelib/openupgradelib/openupgrade.py", line 811, in set_defaults
    obj = pool.get(model, False)
TypeError: get() takes exactly 2 arguments (3 given)
2018-02-11 10:27:31,702 15701 ERROR bs_05_migrated openerp.modules.loading: Error executing post migration script for module base
: get() takes exactly 2 arguments (3 given)


2018-02-11 10:27:31,718 15701 ERROR bs_05_migrated openerp: Failed to initialize database `bs_05_migrated`.
Traceback (most recent call last):
  File "/var/tmp/openupgrade/7.0/server/openerp/cli/server.py", line 97, in preload_registry
    db, registry = openerp.pooler.get_db_and_pool(dbname,update_module=update_module)
  File "/var/tmp/openupgrade/7.0/server/openerp/pooler.py", line 33, in get_db_and_pool
    registry = RegistryManager.get(db_name, force_demo, status, update_module)
  File "/var/tmp/openupgrade/7.0/server/openerp/modules/registry.py", line 203, in get
    update_module)
[...]

PR #96, opened 2018-02-11
[FIX][set_defaults] New api adaptation

bodi000:
Proposition for issue #95

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -818,7 +818,7 @@ def set_defaults(cr, pool, default_spec, force=False, use_orm=False):
                     
         for field, value in default_spec[model]:
             domain = not force and [(field, '=', False)] or []
-            if version_info[0] >= 8:
+            if version_info[0] > 8:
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Hi @bodi000, thanks for noting and suggesting a solution! We are in dire need of tests that validate these functions across different versions of Odoo to prevent regressions. For instance, your version would not work with an Odoo 8.0 pool so I'm cooking something up at https://github.com/OCA/openupgradelib/pull/99.

pedrobaeza:
I have been thinking about the same for a long time in my mind, and I have a different approach for the problem as you have taken for #99. I can put a PoC in the following days.

StefanRijnhart:
@pedrobaeza Mine is nearly finished now, but there might be room for improvement.

StefanRijnhart:
@pedrobaeza was going to reject this in favour of #99. Have you had time to work out your concept of functional tests though?

pedrobaeza:
Fixed with the alternative proposal #99

PR #99, opened 2018-04-02
[ADD] Functional tests

StefanRijnhart:
Continued from https://github.com/OCA/openupgradelib/pull/96 by @bodi000 

* Fixes set_defaults in 7.0
* Fix flake8 not being run on the actual code
* Flake8 fixes
* Add a dedicated test module with tests for 6.1 up to 11.0
* Replace python loading test with running an actual migration from 6.1 to 11.0


StefanRijnhart:
I think I'm getting this travis test scripting all wrong, so if you have any suggestions please come forth.

StefanRijnhart:
In spite of the rough approach, I'm quite pleased with the fact that we can test openupgradelib methods across Odoo versions nw.

StefanRijnhart:
@pedrobaeza @hbrunn still curious what you think. Functional tests for openupgradelib, what's not to like ;-)

openupgradelib/openupgrade.py
```diff
@@ -808,42 +815,38 @@ def write_value(ids, field, value):
                 cr.execute(query, (params[0], sub_ids))
 
     for model in default_spec.keys():
-        obj = pool.get(model, False)
-        if obj == False:
+        if api and isinstance(pool, api.Environment):
+            obj = pool[model]
+        else:
+            try:
```
hbrunn: why not this unconditionally?

paulius-sladkevicius:
@StefanRijnhart would be possible to have `Fixes set_defaults in 7.0` separately done, because now it's included in a functional tests that is not directly related and w/o this fix it's hard to migrate, needs to do manual patching.

mreficent:
@StefanRijnhart what do you say of @paulius-sladkevicius comment? I think there are several people in v6.1 and v7 with the set_defaults problem...

StefanRijnhart:
Well, I still have to see any better proposal than this so let me just rebase this and maybe we can merge it. It can always be refactored later.

StefanRijnhart:
Rebase done.

@pedrobaeza please reconsider to merge this until you have put forward your alternative test strategy.

@hbrunn I've ignored your suggestion about different jobs per version so far because this PR was blocked by Pedro anyway but I have to say I do like the fact that the migration runs an upgrade across all supported versions.

pedrobaeza:
OK, let's go with this approach for now until we have a split way and covering all versions.

agb80:
Can we manke a new release for be able to get this patch easily?

pedrobaeza:
Get it with `pip install --ignore-installed git+https://github.com/OCA/openupgradelib.git@master`

PR #100, opened 2018-04-07
[ADD] add_fields: Add a field definition

pedrobaeza:
This method adds all the needed stuff for having a new field populated in the DB (SQL column, ir.model.fields entry, ir.model.data entry...).

It's intended for being run in pre-migration scripts for pre-populating fields that are going to be declared later in the module.

NOTE: This only works in >=v8 and is not needed in >=v12, as now Odoo always add the XML-ID entry:
https://github.com/odoo/odoo/blob/9201f92a4f29a53a014b462469f27b32dca8fc5a/odoo/addons/base/models/ir_model.py#L794-L802

:param: field_spec: List of tuples with the following expected elements for each tuple:

  * field name
  * model name
  * SQL table name: Put `False` if the model is already loaded in the registry and thus the SQL table name can be obtained that way.
  * field type: binary, boolean, char, date, datetime, float, html, integer, many2many, many2one, monetary, one2many, reference, selection, text, serialized. The list can vary depending on Odoo version or custom added field types.
  * SQL field type: If the field type is custom or it's one of the special cases (see below), you need to indicate here the SQL type to use (from the valid PostgreSQL types):
    https://www.postgresql.org/docs/9.6/static/datatype.html
  * module name: for adding the XML-ID entry.

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -130,6 +130,7 @@ __all__ = [
     'migrate',
     'logging',
     'load_data',
+    'add_fields',
     'copy_columns',
     'rename_columns',
     'rename_fields',
@@ -1685,6 +1686,7 @@ def rename_property(cr, model, old_name, new_name):
         "update ir_property set name=%s where fields_id in %s",
         (new_name, field_ids))
 
+
 def delete_record_translations(cr, module, xml_ids):
     """Cleanup translations of specific records in a module.
 
@@ -1705,6 +1707,7 @@ def delete_record_translations(cr, module, xml_ids):
             WHERE module = %s AND name LIKE %s AND res_id = %s;
         """)
         logged_query(cr, query, (module, row[0] + ',%', row[1],))
+
 
 def disable_invalid_filters(env):
     """It analyzes all the existing active filters to check if they are still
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@hbrunn this has been created due to your comment in https://github.com/OCA/openupgradelib/pull/93#issue-162200382

openupgradelib/openupgrade.py
```diff
@@ -1785,3 +1788,91 @@ def format_message(f):
...
+        'integer': 'int4',
+        'many2many': False,  # No need to create SQL column
+        'many2one': 'int4',
+        'monetary': 'numeric',
+        'one2many': False,  # No need to create SQL column
+        'reference': 'varchar',
+        'selection': 'varchar',  # Can be sometimes integer. Force it manually
+        'text': 'text',
+        'serialized': 'text',
+    }
```
hbrunn: can't you call https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L170 here?

openupgradelib/openupgrade.py
```diff
@@ -1785,3 +1788,91 @@ def format_message(f):
...
+    for vals in field_spec:
+        field_name, model_name, table_name, field_type, sql_type, module = vals
+        # Add SQL column
+        if not table_name:
+            table_name = env[model_name]._table
+        sql_type = sql_type_mapping.get(field_type)
+        if sql_type:
+            logged_query(
+                env.cr, """ALTER TABLE %s ADD COLUMN %s %s""",
+                (AsIs(table_name), AsIs(field_name), AsIs(sql_type)),
```
hbrunn: I myself don't care a lot about comments in the database, other people do. Maybe allow an optional comment and if this exists in the spec, say https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L2582 ?

openupgradelib/openupgrade.py
pedrobaeza: That methods is not available anymore in newer versions. It has been replaced by a property on each field class, which sometimes it's a direct value (example: https://github.com/odoo/odoo/blob/9961bbf4eb87f281e8631d5471bdcad06eb35440/odoo/fields.py#L1120) and sometimes a method that requires a proper instance (example: https://github.com/odoo/odoo/blob/9961bbf4eb87f281e8631d5471bdcad06eb35440/odoo/fields.py#L1820), so it can't be easily reproducible. That's why I opted for this hardcoded option, being limited its scope in the future (it's not needed for >=v12).

openupgradelib/openupgrade.py
pedrobaeza: Me neither. Should we complicate this adding a new tuple item for that? Is it someone to feed that data in the scripts? I don't think so...

hbrunn: yes, whoever wants it, adds it

openupgradelib/openupgrade.py
hbrunn: makes sense, thanks for the links

PR #102, opened 2018-04-23
[IMP] rename_models: Handle properties that reference to the old model

pedrobaeza:
Use case: You are renaming from model A to model B, and there are a many2one company dependent (property) field whose relation is model A. The field "value_reference" of these properties are of the form
'A,id'. We have to change them to 'B,id'.

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -564,6 +564,24 @@ def rename_models(cr, model_spec):
                    "name=%s || substr(name, strpos(name, ',')) "
                    'where name like %s',
                    (new, old + ',%'),)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Yeah, right. Both have marginal use, and thus it hasn't made too much noise, but these are rough edges in a "pixel perfect" migration.

pedrobaeza:
You can merge when you want :wink:

pedrobaeza:
Related PR where I have found the problem: https://github.com/OCA/bank-payment/pull/470

mreficent:
Fixed in https://github.com/OCA/openupgradelib/pull/104.

PR #104, opened 2018-06-06
[FIX] rename_models method

mreficent:
This PR does:

- [x] Fix `rename_models` method

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -538,8 +538,6 @@ def rename_models(cr, model_spec):
         _new = new.replace('.', '_')
         cr.execute('UPDATE ir_model SET model = %s '
                    'WHERE model = %s', (new, old,))
-        cr.execute('UPDATE ir_model_fields SET relation = %s '
-                   'WHERE relation = %s', (new, old,))
         cr.execute('UPDATE ir_model_data SET model = %s '
                    'WHERE model = %s', (new, old,))
         cr.execute(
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Curious what your use case is for forcefully deleting models?

pedrobaeza:
Yeah, me too. @mreficent you tend to mix things together in the same PR. Please make one per "task", so that we are not forced to process them together. Right now, we could merge already the patch for the model renaming if it was alone.

mreficent:
@pedrobaeza Well, I put them together in the same PR because the two methods are related. Don't tell me they are not. The queries in both methods are similar. In one, they are renaming the model but in the other are deleting it. But by comparing the two, it's easier to understand if the new method consider all the cases or not.

@StefanRijnhart I am retaking the migration of the `mrp/mrp_operations` and by doing some local examples, I realized I need this.

StefanRijnhart:
OK, thanks for the clarification.

pedrobaeza:
It's not the same to be related than to be similar. As they are not related, they shouldn't be in the same PR. I see the delete operation very dangerous. We have been always conservative in data in OpenUpgrade. I prefer a rename operation than a removal.

StefanRijnhart:
Yes, maybe it is better if you file this feature separately when we can review your use case in the mrp migration at the same time.

mreficent:
I want to rename model A to B, but if B already exists the renaming queries give errors because unique contrains. So before renaming A to B, I need to delete B. And it doesn't make sense to rename model B to C.

pedrobaeza:
Yes, because you are destroying information that can be important if you want to recreate another thing. But as said:

- Please leave this PR with only the fix, and we merge.
- Create a new one, and point us to the use case in MRP, so that we can analyze it.

mreficent:
Done.

mreficent:
Can be merged :)

StefanRijnhart:
Thanks ;-)

PR #105, opened 2018-06-08
[ADD] merge_records: New method for merging several records into a target one

pedrobaeza:
New method for merging several records into a target one

    NOTE: This should be executed in end migration scripts for assuring that
    all the possible relations are loaded and changed. Tested on v11.

    :param env: Environment variable
    :param model_name: Name of the model of the records to merge
    :param record_ids: List of IDS of records that are going to be merged.
    :param target_record_id: ID of the record where the rest records are going
      to be merge in.
    :param field_spec: Dictionary with field names as keys and forced operation
      to perform as values. If a field is not present here, default operation
      will be performed.

      Possible operations by field types:

      * Char, Text and Html fields:
        - 'merge' (default): content is concatenated with an ' | ' as separator
        - other value: content on target record is preserved
      * Integer, Float and Monetary fields:
        - 'sum' (default): Sum all the values of the records.
        - 'avg': Perform the arithmetic average of the values of the records.
        - 'max': Put the maximum of all the values.
        - other value: content on target record is preserved
      * Binary field:
        - any value: content on target record is preserved
      * Boolean field:
        - 'and': Perform a logical AND over all values.
        - 'or': Perform a logical OR over all values.
        - other value (default): content on target record is preserved
      * Date and Datetime fields:
        - 'max': Put the maximum of all the values.
        - other value (default): content on target record is preserved
      * Many2one fields:
        - any value: content on target record is preserved
      * Many2many fields:
        - 'merge' (default): combine all the values
        - other value: content on target record is preserved
      * One2many fields:
        - 'merge' (default): combine all the values
        - other value: content on target record is preserved
      * Reference fields:
        - any value: content on target record is preserved
      * Selection fields:
        - any value: content on target record is preserved
    :param method: Specify how to perform operations. By default or specifying
      'orm', operations will be performed with ORM, maybe slower, but safer, as
      related and computed fields will be recomputed on changes, and all
      constraints will be checked.

cc @Tecnativa

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -909,20 +909,23 @@ def set_defaults(cr, pool, default_spec, force=False, use_orm=False):
                 write_value(ids, field, value)
 
 
-def logged_query(cr, query, args=None):
+def logged_query(cr, query, args=None, skip_no_result=False):
     """
     Logs query and affected rows at level DEBUG.
 
     :param query: a query string suitable to pass to cursor.execute()
-    :param args: a list, tuple or dictionary passed as substitution values \
-to cursor.execute().
+    :param args: a list, tuple or dictionary passed as substitution values
+      to cursor.execute().
+    :param skip_no_result: If True, then logging details are only shown
+      if there are affected records.
     """
     if args is None:
         args = ()
     args = tuple(args) if type(args) == list else args
     cr.execute(query, args)
-    logger.debug('Running %s', query % args)
-    logger.debug('%s rows affected', cr.rowcount)
+    if not skip_no_result or cr.rowcount:
+        logger.debug('Running %s', query % args)
+        logger.debug('%s rows affected', cr.rowcount)
     return cr.rowcount
 
 
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -1,0 +1,355 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import logging
+import functools
+from psycopg2.extensions import AsIs
+from .openupgrade import logged_query
+from .openupgrade_tools import column_exists, table_exists
+
+logger = logging.getLogger('OpenUpgrade')
+logger.setLevel(logging.DEBUG)
+
+
+def _change_many2one_refs_sql(env, model_name, record_ids, target_record_id):
+    cr = env.cr
+    cr.execute("""
+        SELECT name, model
+        FROM ir_model_fields
+        WHERE ttype='many2one' AND relation=%s
+        """, (model_name, ))
+    for row in cr.dictfetchall():
+        try:
+            model = env[row['model']]
+        except KeyError:
+            continue
+        if not model._auto:  # Discard SQL views
+            continue
+        table = model._table
+        if not table_exists(cr, table):
+            continue
+        column1 = row['name']
+        if not column_exists(cr, table, column1):
+            # TODO: It can be a company dependent field
+            continue
+        logged_query(
+            cr, "UPDATE %s SET %s = %s WHERE %s IN %s", (
+                AsIs(table), AsIs(column1), target_record_id, AsIs(column1),
+                tuple(record_ids),
+            ), skip_no_result=True,
+        )
+
+
+def _change_many2one_refs_orm(env, model_name, record_ids, target_record_id):
+    fields = env['ir.model.fields'].search([
+        ('ttype', '=', 'many2one'),
+        ('relation', '=', model_name),
+    ])
+    for field in fields:
+        try:
+            model = env[field.model]
+        except KeyError:
+            continue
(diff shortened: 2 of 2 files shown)


pedrobaeza:
It's not unrelated. I'm using that new feature in the new method. You have to learn to distinguish :stuck_out_tongue_closed_eyes:

openupgradelib/openupgrade.py
```diff
@@ -909,20 +909,23 @@ def write_value(ids, field, value):
...
+def logged_query(cr, query, args=None, skip_no_result=False):
     """
     Logs query and affected rows at level DEBUG.
 
     :param query: a query string suitable to pass to cursor.execute()
-    :param args: a list, tuple or dictionary passed as substitution values \
-to cursor.execute().
+    :param args: a list, tuple or dictionary passed as substitution values
+      to cursor.execute().
+    :param skip_no_result: If True, then logging details are only shown
```
legalsylvain: Good addition. But should not be True by default. I mean, a request without result has no interest. don't you think ?

pedrobaeza: And it's not... This is the documentation.

legalsylvain: OK, we'll talk about that point in another thread.

pedrobaeza: It's not needed, please check the keyword argument. It's by default `False`.

legalsylvain: Yep. I checked. I didn't explain my PoV correctly.

I just wanna say that by default, **it should be True** to only show relevant log. (but it changes the behaviour of the function, so let's go talking about that point on another thread)

pedrobaeza: Ah, got it! Yeah, I was conservative with previous behavior. We can discuss to see if we prefer to not show them by default.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -0,0 +1,355 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import logging
+import functools
+from psycopg2.extensions import AsIs
+from .openupgrade import logged_query
+from .openupgrade_tools import column_exists, table_exists
+
+logger = logging.getLogger('OpenUpgrade')
+logger.setLevel(logging.DEBUG)
```
Yajo: AFAIK this is called `_logger` as per convention.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -0,0 +1,355 @@
...
+      to perform as values. If a field is not present here, default operation
+      will be performed. See _adjust_merged_values_orm method doc for all the
+      available operators.
+    :param method: Specify how to perform operations. By default or specifying
+      'orm', operations will be performed with ORM, maybe slower, but safer, as
+      related and computed fields will be recomputed on changes, and all
+      constraints will be checked.
+    """
+    args = (env, model_name, record_ids, target_record_id)
+    args2 = tuple(list(args) + [field_spec])
```
Yajo: You can do `args + (field_spec,)`

openupgradelib/openupgrade_merge_records.py
pedrobaeza: Well, here in openupgradelib we use `logger`

openupgradelib/openupgrade_merge_records.py
pedrobaeza: Done

openupgradelib/openupgrade_merge_records.py
```diff
@@ -0,0 +1,355 @@
...
+        if field.type in ('char', 'text', 'html'):
+            op = op or 'merge'
+            if op == 'merge':
+                vals[field.name] = ' | '.join(l)
+        elif field.type in ('integer', 'float', 'monetary'):
+            op = op or 'sum'
+            if op == 'sum':
+                vals[field.name] = sum(l)
+            elif op == 'avg':
+                vals[field.name] = sum(l) / len(l)
```
mreficent: Should the result be rounded in case of type='integer'?

openupgradelib/openupgrade.py
StefanRijnhart: I think this could be a global setting (Odoo's debug mode?) so that developers can see all queries while at runtime you only see the effective queries.

StefanRijnhart:
Interesting! Just wondering why you are proposing this in openupgradelib. It seems useful beyond migrations. Did you consider putting this in a more general Odoo method lib?

pedrobaeza:
> Did you consider putting this in a more general Odoo method lib?

No, because this is a foundation library that can be used also by Odoo modules. My idea is to implement an record_merger module that depends on this library with the UI and so. I made this proposition last year also on Odoo: https://github.com/odoo/odoo/issues/17268. I will try to ask for it this year too.

openupgradelib/openupgrade_merge_records.py
pedrobaeza: It's handled automatically by ORM.

pedrobaeza:
If you don't mind, I'm going to merge this one as well the stock migration for continuing with the rest of the dependent modules.

PR #106, opened 2018-06-29
Fix conversion to attachment on large datasets

dreispt:
Selecting all binary data at once will exceed memory limits, and get the
Odoo process killed.
Fetching one by one is not enough, since executing the SELECT to
retrieve all the table data can also fail with a Postgresql "out of
memory" error.

I had a bottleneck on res_partner.image, with almost 5k images to convert.
With this change this upgrade step was successful, and took around 6 minutes to run.

### DIFF
--- a/openupgradelib/openupgrade_90.py
+++ b/openupgradelib/openupgrade_90.py
@@ -50,21 +50,31 @@ def convert_binary_field_to_attachment(env, field_spec):
                 "Converting to attachment field {} from model {} stored in "
                 "column {}".format(field, model_name, column)
             )
-            env.cr.execute(
-                """SELECT id, {} FROM {};""".format(column, model._table)
-            )
-            for row in env.cr.fetchall():
-                data = bytes(row[1])
-                if not data or data == 'None':
-                    continue
-                attachment_model.create({
-                    'name': field,
-                    'res_model': model_name,
-                    'res_field': field,
-                    'res_id': row[0],
-                    'type': 'binary',
-                    'datas': data,
-                })
+            last_id = 0
+            while True:
+                env.cr.execute(
+                    """SELECT id, {0} FROM {1} WHERE {0} IS NOT NULL AND id > {2}
+                    ORDER BY id LIMIT 500;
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I'm not sure about the performance of this. Instead of selecting one by one, I think you can read content by chunks of for example 500, which a balance between memory consumption and speed making a lot of independent querys

dreispt:
I had the same thought when writing the code, but when running it the performance penalty didn't seem relevant enough to be worth that complication: smaller tables still took few seconds to process, and the large table processed 5k  images in 6 minutes - doesn't seem unreasonable.

gdgellatly:
Just as a mental exercise I thought of an alternative algorithm to avoid the duplicate selects.  Pseudocode, probably missing something.

```
SELECT id's FROM table WHERE field not=NULL
for id in ids:
    ir_attachment.create('datas': '1')
UPDATE ir_attachment SET datas=field.name FROM model WHERE model.id=res_id AND res_model = model_name AND res_field = field
```
Then the binary data never even hits openupgrade.

dreispt:
I wrote an improvement, following Pedro's suggestion, but I haven't tested it yet: https://github.com/dreispt/openupgradelib/commit/cf38433b06437c3ff1b05f9114953d2ca74cc8e4

dreispt:
PR updated with the new code.
It now works on batches of 500 records.

pedrobaeza:
> It would still be much quicker to create the attachments using SQL in a single query

Not really, check https://stackoverflow.com/questions/8635818/multiple-insert-statements-vs-single-insert-with-multiple-values

dreispt:
🎉 My first OpenUpgrade related contrib!

StefanRijnhart:
@pedrobaeza It's more that you would use SQL instead of the ORM.

StefanRijnhart:
@dreispt thanks and congrats! Keep em coming!

pedrobaeza:
@StefanRijnhart right, I see now the `create` statement

PR #108, opened 2018-07-19
rename_columns: do not fail if table does not exist

sbidoul:
This helps with the migration of payment orders on 8.0 databases with an old version of account_banking_payment_transfer.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -381,7 +381,7 @@ def rename_columns(cr, column_spec):
             logger.info("table %s, column %s: renaming to %s",
                         table, old, new)
             cr.execute(
-                'ALTER TABLE "%s" RENAME "%s" TO "%s"' % (table, old, new,))
+                'ALTER TABLE IF EXISTS "%s" RENAME "%s" TO "%s"' % (table, old, new,))
             cr.execute('DROP INDEX IF EXISTS "%s_%s_index"' % (table, old))
 
 


pedrobaeza:
This has been this by design. If you want to be resilient to non existing columns, you should do it on your migration scripts, but not silently bypass it.

About what you mention of the source of the problem, it's a sane requirement to update to the latest code of N - 1 version before upgrading to version N, as this has been also the requirement when developing migration scripts. If not, you have to do complicated scripts that covers all the possible module versions.

For me, this is :-1:

StefanRijnhart:
Agree with Pedro

sbidoul:
> update to the latest code of N - 1 version before upgrading to version N

That does not sound like a realistic requirement to me. For the cases I have in mind it looks much simpler to make the migration script more robust.

So `def rename_columns(cr, column_spec, ignore_missing_tables=False)`?

pedrobaeza:
That option seems OK for openupgradelib, but I'm not going to use it for the reasons stated.

Are you really saying you have your 8.0 installation frozen in non latest code? That's for me something not maintainable, but well, if so, again I think it's better to update to latest code, and then proceed to the migration, although you don't update on 8.0 production database. Trying to make migration scripts that covers all situations and intermediate versions will make you rework all the past contributions, migration script, etc.

sbidoul:
> Are you really saying you have your 8.0 installation frozen in non latest code

Of course. The vast majority of our customers prefer stability.

> Trying to make migration scripts that covers all situations and intermediate versions

Migration scripts should be designed to be resilient has much as possible. We cannot expect people to install all intermediate versions of a module. I mean, if customer has module 8.0.1.0.0, he must be able to install 8.0.3.0.0 and not be obliged to install 8.0.2.0.0 in between. The reasoning is not that much different when migrating across Odoo series.

pedrobaeza:
> Of course. The vast majority of our customers prefer stability.

Stability for me is having latest code with all the possible patches. Quality checks are for avoiding regressions on so on, and when eventually happens, they are fixed quickly. I invite you try the other approach, I can assure you that you won't find surprises, and more in these days (it could be different in the past).

> Migration scripts should be designed to be resilient has much as possible. We cannot expect people to install all intermediate versions of a module. I mean, if customer has module 8.0.1.0.0, he must be able to install 8.0.3.0.0 and not be obliged to install 8.0.2.0.0 in between. The reasoning is not that much different when migrating across Odoo series.

We are not talking about installing each of the intermediate versions. Of course that's not feasible, but again at least pass one time for the latest version code. Inside same branch, several migration scripts for each version will make the rest when updating to the latest one. See for example `l10n_es` module on 8.0:

https://github.com/OCA/l10n-spain/tree/8.0/l10n_es/migrations

It received 3 versions with changes of the DB that requires a migration script, but having them inside that branch makes that updating from 8.0.4.0 to 8.0.5.1 was smooth. The rest of the changes are handled by the ORM itself. For example, if you have one field added in one version, and another field added in the next one, updating to latest one will add both fields at once.

pedrobaeza:
> Can I ask you if there were data model changes in account_banking_payment_transfer, how come it does not have a migration script in the 8.0 version of the module?

@StefanRijnhart I think this is because it's not needed, due to they are simply new fields that are added and filled through regular ORM, but it can be that it also requires that intermediate script for populating some data. 8.0 was still some kind of wild west in the migration area. Fortunately, now there are a lot companies that cares on migration things (inside the same or different versions), and OCA also added some guidelines for this, although they can't be tracked.

gdgellatly:
Sorry to hijack, but I had a question that relates to this that I've been wanting to ask and the right people are here to answer.

I have 1 site on a pinned version of Odoo 7.0 due to a regression introduced which Odoo refused to revert (essentially the regression was not as impactful as the bug).  However, the nature of the regression is in the webclient so I can upgrade to latest.

Now this site wishes to go to v11, with 80,000 lines of custom code.

To use OpenUpgrade, I get that I need to step each version, but can I just not step the custom code and do a v10 - 11 migration script when I get to that version?  Or just stub the modules at each version? Idk, but I really don't feel like porting all of that to 4 versions.

Any hints appreciated.

sbidoul:
@StefanRijnhart it was not needed indeed. The "problematic" table is bank_payment_line which as not populated for old payment orders in 8.0 because it was not really necessary to have it for historical data. Now the 9.0 migration script chokes if it is absent.

sbidoul:
@pedrobaeza rolling releases of OCA modules is not going to work (for our customers, I won't say for others), I'm sure.

sbidoul:
@gdgellatly your question is relevant. We discussed something similar in https://github.com/OCA/bank-payment/issues/494.

In a nutshell, I personally don't understand why we drop old migration scripts when moving to the next Odoo branch while we keep them inside the same Odoo series. I'm aware it's a hard problem and all possible pitfalls, but still I believe this creates more problems than it solves.

pedrobaeza:
@sbidoul the question that @gdgellatly makes is not the same as that thread. I elaborate later.

StefanRijnhart:
@gdgellatly develop your custom code for the target release only. If they are not present in the addons path in the intermediate releases, their upgrade will remain pending.

gdgellatly:
@StefanRijnhart @sbidoul @pedrobaeza Thanks, of course any more info welcome.  I'm starting on this next week, so I'll try and write something up for docs once I got it going because I think skipping versions with custom code is probably a common use case.

sbidoul:
@StefanRijnhart 

> include any model changes that were taken for granted

Perhaps pre-migration scripts should check which minimum version of the module must be installed for them to work correctly?

StefanRijnhart:
@sbidoul let's say that from 8 to 9 you have a prescript, the data model changes from Odoo and a postscript. The new fields or tables from 8 to 9 may be changed again from 9 to 10. The methods that are called in the postscript may not available anymore. The scripts that you created for the upgrade 9 would be useless in version 10

StefanRijnhart:
Instead of going through 8.0 -> 9.0 -> 10.0 -> 11.0 you'd effectively have to write scripts from 8.0 -> 9.0, 8.0 -> 10.0, 8.0 -> 11.0, 9.0 -> 10.0, 9.0 -> 11.0, 10.0 -> 11.0. Looks like more work to me!

dreispt:
@sbidoul I can see why each version only has the scripts to migrate from the previous one. I don't see a big issue to "checkout 9.0; migrate to 9.0; checkout 10.0; migrate to 10.0". And it surely simplifies the migration script creation and maintenance.

@gdgellatly I'm trying to get into Openupgrade also, and for the custom code I'm thinking on using a "model migration only" approach: you disable all Views and as much business logic methods as possible, leaving only the data model and maybe some essential CRUD logic. You then perform the multiple version migration, keeping all the data, and the reassess the Views and Business logic for the latest version only.

*PS: multiversion migration discussion it's own Issue, and should be discussed in the docs.*

sbidoul:
> The scripts that you created for the upgrade 9 would be useless in version 10

I'm aware of that. Most of them will work though, especially if they do not use the orm.

gdgellatly:
haha @dreispt you and me in same boat, having been anygrate users before.

StefanRijnhart:
@dreispt I never have my custom addons in my addons path during the migration of intermediate releases. It works quite well.

dreispt:
@StefanRijnhart Thanks for sharing that. Sorry for the possibly dumb question: is there any risk for the custom fields and tables to be lost during the core module upgrades?

StefanRijnhart:
@dreispt no, not that I know of, unless there is a clash with new fields in a later release.

pedrobaeza:
We agree then to not add this option and be controlled outside the call to the function, so closing.

PR #114, opened 2018-08-02
[ADD] delete_records_safely_by_xml_id

pedrobaeza:
This removes in the safest possible way the records whose XML-IDs are passed as argument.

:param xml_ids: List of XML-ID string identifiers of the records to remove.


mreficent:
I would like this method was named `delete_records_safely_by_xml_id` instead like the `delete_record_translations` method.

pedrobaeza:
Changed!

PR #115, opened 2018-08-14
Add convert_text_to_serialized

SimoRubi:
Added utility method to convert text into serialized field


StefanRijnhart:
I can see how this is useful in https://github.com/OCA/reporting-engine/pull/232/files, but I don't think it's generic enough for here.

PR #117, opened 2018-09-27
[RFR] No yaml import in Odoo 12

StefanRijnhart:


StefanRijnhart:
We have another year to think about the test scheme as we only load YML into the source version which is currently 11 and still supports YML import.

PR #119, opened 2018-10-02
[ADD] search and replace a single ID across the whole database

StefanRijnhart:


openupgradelib/openupgrade.py
```diff
@@ -196,21 +199,27 @@ def allow_pgcodes(cr, *codes):
...
+                yield
+    except (ProgrammingError, IntegrityError) as error:
+        if kwargs.get('query'):
+            logger.info(kwargs['query'])
         msg = "Code: {code}. Class: {class_}. Error: {error}.".format(
             code=error.pgcode,
             class_=errorcodes.lookup(error.pgcode[:2]),
             error=errorcodes.lookup(error.pgcode))
-        if error.pgcode not in codes and error.pgcode[:2] in codes:
+        if error.pgcode in codes or error.pgcode[:2] in codes:
```
StefanRijnhart: Fixing what looks like an obvious bug after a refactoring. @yajo?

StefanRijnhart:
I'm using this in the migration of the base module when forking the new system user from the admin user in Odoo 12.0

openupgradelib/openupgrade.py
```diff
@@ -166,11 +168,12 @@ def do_raise(error):
     'delete_record_translations',
     'disable_invalid_filters',
     'delete_records_safely_by_xml_id',
+    'search_and_replace_single_id',
 ]
 
 
 @contextmanager
-def allow_pgcodes(cr, *codes):
+def allow_pgcodes(cr, *codes, **kwargs):
```
StefanRijnhart: Python 3 will support a signature like (cr, *codes, query=None), but Python 2 does not.

openupgradelib/openupgrade.py
Yajo: Yes, so it seems, thanks!

openupgradelib/openupgrade.py
```diff
@@ -2011,3 +2020,100 @@ def chunked(records, single=True):
...
+            WHERE constraint_type = 'FOREIGN KEY'
+            AND ccu.table_name = %s and ccu.column_name = 'id'
+        """, (env[model]._table,))
+    for table, column in env.cr.fetchall():
+        if (table, column) in exclude_columns:
+            continue
+        query = env.cr.mogrify(
+            """ UPDATE {table} SET {column} = %(new_id)s
+                WHERE {column} = %(old_id)s""".format(
+                    table=AsIs(table), column=AsIs(column)), values)
```
Yajo: I think you're getting confused by the `AsIs` function. The point is that you don't need to format it twice, as it seems you are doing here. This should do the same (and is actually the real benefit):

```python
env.cr.execute(
    """UPDATE %(table)s SET %(column)s = %(new_id)s
       WHERE %(column)s = %(old_id)s""",
    dict(values, table=AsIs(table), column=AsIs(column),
)
```

openupgradelib/openupgrade.py
```diff
@@ -197,21 +200,27 @@ def allow_pgcodes(cr, *codes):
...
 
     try:
         with cr.savepoint():
-            yield
-    except ProgrammingError as error:
+            with core.tools.mute_logger('odoo.sql_db'):
+                yield
+    except (ProgrammingError, IntegrityError) as error:
+        if kwargs.get('query'):
+            logger.info(kwargs['query'])
```
Yajo: It seems more logical to me that you add this to the `msg` below.

openupgradelib/openupgrade.py
```diff
@@ -2011,3 +2020,100 @@ def chunked(records, single=True):
...
+        """, (env[model]._table,))
+    for table, column in env.cr.fetchall():
+        if (table, column) in exclude_columns:
+            continue
+        query = env.cr.mogrify(
+            """ UPDATE {table} SET {column} = %(new_id)s
+                WHERE {column} = %(old_id)s""".format(
+                    table=AsIs(table), column=AsIs(column)), values)
+        with allow_pgcodes(env.cr, errorcodes.UNIQUE_VIOLATION, query=query):
+            logged_query(env.cr, query)
```
Yajo: If `logged_query` already logs it, why do we need the `query` kwarg in `allow_pgcodes`? :thinking:

StefanRijnhart: Logged query logs afterwards, to mention the number of affected records. If an exception occurs, the offending query is not logged. As I'm writing this, it occurs to me that I should solve this in logged_query instead....

StefanRijnhart:
@pedrobaeza I will have a look at unifying the methods.

StefanRijnhart:
@pedrobaeza I have merged my method with yours. Will test later today or this weekend.

openupgradelib/openupgrade.py
StefanRijnhart: Indeed, thank you. I'll be addressing this in my refactoring.

StefanRijnhart:
Tested, fixed and improved @pedrobaeza

PR #121, opened 2018-10-17
[FIX] merge_records: missing argument in _adjust_merged_values_orm

mreficent:
Bug.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -434,8 +434,9 @@ def merge_records(env, model_name, record_ids, target_record_id,
         field_spec = {}
     if isinstance(record_ids, list):
         record_ids = tuple(record_ids)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Well, then it's advisable to add the logic for excluding that columns indeed.

mreficent:
What do you mean? The exclusion of columns was for other methods, not for this one. This method doesn't have to exclude anything. Just happens that all methods are called with the same arguments and so this one has to have that argument.

pedrobaeza:
I would say that you haven't to adjust values for the excluded columns.

StefanRijnhart:
Thanks for the fix! If it's not used, maybe don't pass it at all but change the composition of args2 not to include it.

mreficent:
Done. I think now it's ok for all tastes :)

pedrobaeza:
I'm afraid this is still not correct: https://travis-ci.org/OCA/OpenUpgrade/builds/442749754#L6799

PR #124, opened 2018-11-20
[FIX][DOC] _adjust_merged_values_orm: char default case

mreficent:
If you see in the code, the merges in the `char` case are different from `text` and `html` case.


mreficent:
What is this error for python 2.7?

mreficent:
I will merge this PR as the error is unrelated

mreficent:
@StefanRijnhart can you check the error? It seems that is related with trying something of openupgrade v12 with python 2.7

PR #126, opened 2018-11-20
[FIX][DOC] _adjust_merged_values_orm: integer default case

mreficent:
Another case.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -232,7 +232,7 @@ def _change_translations_sql(env, model_name, record_ids, target_record_id,
 
 def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
                               field_spec):
-    """"This method deals with the values on the records to be merged +
+    """This method deals with the values on the records to be merged +
     the target record, performing operations that makes sense on the meaning
     of the model.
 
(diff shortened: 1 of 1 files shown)


mreficent:
@pedrobaeza

openupgradelib/openupgrade_merge_records.py
```diff
@@ -242,18 +242,17 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
-      * Text and Html fields:
-        - 'merge' (default): content is concatenated with an ' | ' as separator
-        - other value: content on target record is preserved
+      * Char, Text and Html fields:
+        - 'merge' (default for Text and Html): content is concatenated
+           with an ' | ' as separator
+        - other value (default for Char): content on target record is preserved
       * Integer, Float and Monetary fields:
-        - 'sum' (default): Sum all the values of the records.
+        - 'sum' (default for Float and Monetary): Sum all records's values.
```
pedrobaeza: Saxon genitive is for persons: https://escuelavictoria.es/saxon-genitive-2/, so you should use previous form.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -242,18 +242,17 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
         - 'avg': Perform the arithmetic average of the values of the records.
         - 'max': Put the maximum of all the values.
         - 'min': Put the minimum of all the values.
-        - other value: content on target record is preserved
+        - other value (default for Integer): content on target record
+           is preserved
```
pedrobaeza: bad indentation

mreficent:
Now it's full correct? 😏

pedrobaeza:
I'm afraid not: when I said bad indentation, it was because that line had 3 spaces instead of 2, but you should preserve 2 spaces indentation on bullet lists, although docstring formatting is tricky, so maybe you can pass any kind of checker for being sure.

pedrobaeza:
I have added #128 for a new interesting feature.

Issue #125, opened 2018-11-20
merge_records: compatible with ondelete restrict


mreficent:
Suppose you are merging records of a model A with merge_records method. Then, if those records are used in another model B using a many2one with `ondelete=restrict`, then it breaks.

To solve this, the method should be adapted to include this case: before doing the deletion of the record of model A, it should change the records of model B that reference that record with the new id.

I would do it myself, but I am a bit confused on where the exact code change should be done.

pedrobaeza:
It shouldn't be any problem. Check the order in 

https://github.com/OCA/openupgradelib/blob/0ac601497c9849c3f257ca505bc00d28af368155/openupgradelib/openupgrade_merge_records.py#L443

First, you change references, and at the end is when you unlink old records. Have you found any real problem or you are only theorizing about this?

mreficent:
When testing https://github.com/OCA/OpenUpgrade/pull/1619 on a database with some duplicated `res.country.state` records, in the post-migration, I receive this error:

IntegrityError: update or delete on table "res_country_state" violates foreign key constraint "res_partner_state_id_fkey" on table "res_partner"
DETAIL:  Key (id)=(3924) is still referenced from table "res_partner".

I supposed that it because `res.partner` has `ondelete=restrict` in `country_id` and `state_id`.

pedrobaeza:
Are you using SQL or ORM. SQL is not fully implemented. That can be the reason. I see also a lot of other code in your PR that can be the source. Is that code really needed? I think it can be simplified if needed.

mreficent:
It's using ORM (that's what is marked as default).

pedrobaeza:
Then it should be another thing as stated. Can you share demo set and full stack trace?

PR #127, opened 2018-11-20
[FIX] merge_records (orm): take into account inactive records

mreficent:
Fixes https://github.com/OCA/openupgradelib/issues/125.

By default, the `search` method avoids the inactive records.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -83,7 +83,7 @@ def _change_many2one_refs_orm(env, model_name, record_ids, target_record_id,
     ])
     for field in fields:
         try:
-            model = env[field.model]
+            model = env[field.model].with_context(active_test=False)
         except KeyError:
             continue
         field_name = field.name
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_merge_records.py
```diff
@@ -92,7 +92,12 @@ def _change_many2one_refs_orm(env, model_name, record_ids, target_record_id,
             not field.store or (model._table, field_name) in exclude_columns
         ):
             continue  # Discard SQL views + invalid fields + non-stored fields
-        records = model.search([(field_name, 'in', record_ids)])
+        if 'active' in model._fields:
```
pedrobaeza: Do `model.with_context(active_test=False).search...` instead

pedrobaeza: And better to do it on the line of `model = env[field.model]`

pedrobaeza:
I have fixed Travis (or I think so) in https://github.com/OCA/openupgradelib/commit/d58affc2bf62f55458c7011575ebf1a47c082328, so rebase.

mreficent:
Ready to merge! :green_apple:

PR #129, opened 2018-11-23
[FIX] Merge records on sql rel tables

hveficent:
Merge records on SQL mode didn't take into consideration relational tables where is no id column_name.

Will this break another scenario?

I'm not so sure.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -51,28 +51,78 @@ def _change_foreign_key_refs(env, model_name, record_ids, target_record_id,
         except (ProgrammingError, IntegrityError) as error:
             if error.pgcode != UNIQUE_VIOLATION:
                 raise
-            # Fallback on setting each row separately
+            # Check if the table contains an id
             env.cr.execute(
-                """ SELECT id FROM %(table)s
-                    WHERE "%(column)s" in %(record_ids)s """, {
-                        'table': AsIs(table),
-                        'column': AsIs(column),
-                        'record_ids': record_ids})
-            for row in env.cr.fetchall():
-                try:
-                    with env.cr.savepoint():
-                        logged_query(
-                            env.cr,
-                            """
-                            UPDATE %(table)s
-                            SET "%(column)s" = %(target_record_id)s
-                            WHERE id = %(id)s """, {
-                                'id': row[0],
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_merge_records.py
```diff
@@ -51,28 +51,78 @@ def _change_foreign_key_refs(env, model_name, record_ids, target_record_id,
...
-                        raise
+                """ SELECT column_name
+                    FROM information_schema.columns
+                    WHERE table_name = '%(table)s' and column_name='id'
+                """, {
+                        'table': AsIs(table)
+                })
+            if not env.cr.fetchone():
+                # There's no id. So this is a many2many table. We need
+                # to remove the record.
```
lreficent: `remove` is misleading here you are not doing that in the code, you are actually replacing with the target record, wich I think is good.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -51,28 +51,78 @@ def _change_foreign_key_refs(env, model_name, record_ids, target_record_id,
...
+                env.cr.execute("""
+                    SELECT %(column)s FROM %(table)s
+                    WHERE "%(column)s" in %(record_ids)s
+                    """, {
+                            'table': AsIs(table),
+                            'column': AsIs(column),
+                            'record_ids': record_ids})
+                for col in env.cr.fetchall():
```
mreficent: col -> row

openupgradelib/openupgrade_merge_records.py
```diff
@@ -51,28 +51,78 @@ def _change_foreign_key_refs(env, model_name, record_ids, target_record_id,
...
+                        with env.cr.savepoint():
+                            logged_query(
+                                env.cr,
+                                """
+                                UPDATE %(table)s
+                                SET "%(column)s" = %(target_record_id)s
+                                WHERE id = %(id)s """, {
+                                    'id': row[0],
+                                    'table': AsIs(table), 'column':
+                                    AsIs(column),
```
mreficent: It would look nicer if you put instead:
```python
    'table': AsIs(table),
    'column': AsIs(column),
```

pedrobaeza:
I have refactored the code for being more compact and readable, and the method is working, as seen in https://github.com/OCA/OpenUpgrade/pull/1579, so I merge.

Issue #128, opened 2018-11-21
openupgrade_merge_records: Add new adjust value


pedrobaeza:
Another interesting missing merged value is to get first not null on many2one/binary fields. Example: you merge countries, and you have the flag image. The selected target record doesn't have image, but one of the rest of merged records has. If we put target, the image will be missing.

PR #130, opened 2018-11-26
[IMP] openupgrade_merge_records: add adjust values for binary & many2one

mreficent:
Implements https://github.com/OCA/openupgradelib/issues/128.


openupgradelib/openupgrade_merge_records.py
```diff
@@ -325,6 +329,12 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
             if op == 'merge':
                 o2m_changes += 1
                 l.write({field.inverse_name: target_record_id})
+        elif field.type in ('binary', 'many2one'):
+            op = op or 'merge'
+            if op == 'merge':
+                if not getattr(target_record, field.name) and l and not \
+                        vals.get(field.name):
+                    vals[field.name] = l
```
pedrobaeza: This can contain more than 1 value, so put `l[:1]`

openupgradelib/openupgrade_merge_records.py
```diff
@@ -265,7 +267,9 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
         - 'min': Put the minimum of all the values.
         - other value (default): content on target record is preserved
       * Many2one fields:
-        - any value: content on target record is preserved
+        - 'merge' (default): apply first not null value of the records if
+        value of target record is null.
```
pedrobaeza: Append `, preserve target value otherwise`.

PR #135, opened 2018-12-19
merge_records: translations + exception

pedrobaeza:
3 things:

* **[FIX] Handle better translatable fields**

  When merging records from a model that has translatable fields, before this patch, the translations for all the records are reassigned to the target record, provoking duplicated translations. This is even forgiven in v12.

  With this new approach, target translation has priority, and if not present, then the first found is assigned instead.

* **[IMP] merge_records: Put exception condition on records to be merged**

  For avoiding that the target record is also included in records to be merged.

* **[IMP] Check which records to be merged exist**

  This way, if you pass unexisting records to the method, it won't crash.

  In addition, if no records to be merged after this purge, then just return w/o error.

Fixes #13 

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -202,21 +202,35 @@ def _change_translations_orm(env, model_name, record_ids, target_record_id,
                              exclude_columns):
     if ('ir_translation', 'res_id') in exclude_columns:
         return
-    records = env['ir.translation'].search([
+    translation_obj = env['ir.translation']
+    groups = translation_obj.read_group([
         ('type', '=', 'model'),
         ('res_id', 'in', record_ids),
-        ('name', 'like', '%s,%%' % model_name)])
-    if records:
-        records.write({'res_id': target_record_id})
-        logger.debug(
-            "Changed %s translations of model 'ir.translation'",
-            len(records))
+        ('name', 'like', '%s,%%' % model_name),
+    ], ['name', 'lang'], ['name', 'lang'], lazy=False)
+    for group in groups:
+        target_translation = translation_obj.search([
+            ('type', '=', 'model'),
+            ('res_id', '=', target_record_id),
+            ('name', '=', group['name']),
+            ('lang', '=', group['lang']),
+        ])
+        records = translation_obj.search(group['__domain'])
+        if not target_translation and records:
+            # There is no target translation, we pick one for being the new one
+            records[:1].res_id = target_record_id
+            records = records[1:]
+        if records:
+            records.unlink()
+            logger.debug("Deleted %s extra translations for %s.",
+                         len(records), group['name'])
 
 
 def _change_translations_sql(env, model_name, record_ids, target_record_id,
                              exclude_columns):
     if ('ir_translation', 'res_id') in exclude_columns:
         return
+    # TODO: To be handled with the same approach as in ORM method
     logged_query(
         env.cr,
         """
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -455,7 +455,9 @@ def merge_records(env, model_name, record_ids, target_record_id,
     args0 = (env, model_name, record_ids, target_record_id)
     args = args0 + (exclude_columns, )
     args2 = args0 + (field_spec, )
-
+    if target_record_id in record_ids:
+        raise Exception("You can't put the target record in the list or "
+                        "records to be merged.")
     _change_generic(*args, method=method)
     if method == 'orm':
         _change_many2one_refs_orm(*args)
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -458,6 +458,10 @@ def merge_records(env, model_name, record_ids, target_record_id,
     if target_record_id in record_ids:
         raise Exception("You can't put the target record in the list or "
                         "records to be merged.")
+    # Check which records to be merged exist
+    record_ids = env[model_name].browse(record_ids).exists().ids
+    if not record_ids:
+        return
     _change_generic(*args, method=method)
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_merge_records.py
```diff
@@ -202,21 +202,38 @@ def _change_translations_orm(env, model_name, record_ids, target_record_id,
...
+        ])
+        records = translation_obj.search(group['__domain'])
+        if not records:
+            continue
+        if target_translation:
+            # A translation already exists - We delete the rest of translations
+            logger.debug("Deleted %s extra translations.'", len(records))
+            records.unlink()
+        else:
+            # There is no target translation, we pick one for being the new one
```
mreficent: Shouldn't you check if len(records) > 1? Because if len(records) = 1 then you shouldn't delete it.

pedrobaeza: It doesn't happen anything. `unlink` over an empty recordset does nothing, but no error.

mreficent: Well, its a bit silly "Deleted 0 extra translations." Maybe this generates a lot of spam

pedrobaeza: OK, changing directly in GitHub...

openupgradelib/openupgrade_merge_records.py
```diff
@@ -202,21 +202,40 @@ def _change_translations_orm(env, model_name, record_ids, target_record_id,
...
+            ('res_id', '=', target_record_id),
+            ('name', '=', group['name']),
+            ('lang', '=', group['lang']),
+        ])
+        records = translation_obj.search(group['__domain'])
+        if not records:
+            continue
+        if target_translation:
+            # A translation already exists - We delete the rest of translations
+            logger.debug("Deleted %s extra translations.'", len(records))
```
mreficent: extra `'`

openupgradelib/openupgrade_merge_records.py
```diff
@@ -202,21 +202,40 @@ def _change_translations_orm(env, model_name, record_ids, target_record_id,
...
+            records.unlink()
+        else:
+            # There is no target translation, we pick one for being the new one
+            records[:1].res_id = target_record_id
+            if len(records) > 1:
+                records = records[1:]
+                logger.debug("Deleted %s extra translations.'", len(records))
```
mreficent: extra `'`

pedrobaeza:
Changed!

openupgradelib/openupgrade_merge_records.py
```diff
@@ -455,7 +455,13 @@ def merge_records(env, model_name, record_ids, target_record_id,
     args0 = (env, model_name, record_ids, target_record_id)
     args = args0 + (exclude_columns, )
     args2 = args0 + (field_spec, )
-
+    if target_record_id in record_ids:
+        raise Exception("You can't put the target record in the list or "
```
mreficent: .ValidationError,.UserError? :S

pedrobaeza: It should be a general Exception, as this is a library, not ORM code.

mreficent: ok

pedrobaeza:
I have added more changes, but trying simultaneously if they work, so I tell you when finished.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -202,21 +202,35 @@ def _change_translations_orm(env, model_name, record_ids, target_record_id,
...
-            len(records))
+        ('name', 'like', '%s,%%' % model_name),
+    ], ['name', 'lang'], ['name', 'lang'], lazy=False)
+    for group in groups:
+        target_translation = translation_obj.search([
+            ('type', '=', 'model'),
+            ('res_id', '=', target_record_id),
+            ('name', '=', group['name']),
+            ('lang', '=', group['lang']),
+        ])
```
mreficent: limit=1?

openupgradelib/openupgrade_merge_records.py
```diff
@@ -202,21 +202,35 @@ def _change_translations_orm(env, model_name, record_ids, target_record_id,
...
-        ('name', 'like', '%s,%%' % model_name)])
-    if records:
-        records.write({'res_id': target_record_id})
-        logger.debug(
-            "Changed %s translations of model 'ir.translation'",
-            len(records))
+        ('name', 'like', '%s,%%' % model_name),
+    ], ['name', 'lang'], ['name', 'lang'], lazy=False)
+    for group in groups:
+        target_translation = translation_obj.search([
```
mreficent: with_context(active_test=False)

openupgradelib/openupgrade_merge_records.py
pedrobaeza: As we are not reading values after this search, it's the same limiting or not. And in v12, this will be directly limited by the constraint, and for the rest, it's weird to have more than 1 values. We can leave it as is.

mreficent: Put perfomance-wise it's better reading one than more than one or not? :S

mreficent: Put perfomance-wise it's better reading one than more than one or not? :S

pedrobaeza: As said, we are not reading anything, so this is going to have the same performance. Even more, limiting will add an extra operation on PG side.

mreficent: just nitpicking

pedrobaeza: nitpicking is when there's a valid comment, but not critical. This is not the case, as you can check reading my first comment, so don't persist in this case.

pedrobaeza:
Tested and working without problem. Merging then. Thanks for your review, Miquel (although nitpickings, hehe).

Issue #138, opened 2019-02-09
merge_records fails hard if some value of a list is null


rvalyi:
example:
```
2019-02-08 23:25:15,944 144 ERROR openupgrade-to-11 odoo.modules.registry: Failed to load registry                                                                                                           
Traceback (most recent call last):                                                                                                                                                                           
  File "/odoo/src/odoo/modules/registry.py", line 85, in new                                                                                                                                                 
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)                                                                                                                               
  File "/odoo/src/odoo/modules/loading.py", line 463, in load_modules                                                                                                                                        
    migrations.migrate_module(package, 'end')                                                                                                                                                                
  File "/odoo/src/odoo/modules/migration.py", line 187, in migrate_module                                                                                                                                    
    migrate(self.cr, installed_version)                                                                                                                                                                      
  File "/src/openupgradelib/openupgradelib/openupgrade.py", line 1445, in wrapped_function                                                                                                                   
    if use_env2 else cr, version)                                                                                                                                                                            
  File "/odoo/src/addons/stock/migrations/11.0.1.1/end-migration.py", line 31, in migrate                                                                                                                    
    merge_quants(env)                                                                                                                                                                                        
  File "/odoo/src/addons/stock/migrations/11.0.1.1/end-migration.py", line 25, in merge_quants                                                                                                               
    env, 'stock.quant', quants[1:].ids, quants[0].id, QUANT_MERGE_OPS,                                                                                                                                       
  File "/src/openupgradelib/openupgradelib/openupgrade_merge_records.py", line 505, in merge_records
    _adjust_merged_values_orm(*args2)
  File "/src/openupgradelib/openupgradelib/openupgrade_merge_records.py", line 354, in _adjust_merged_values_orm
    vals[field.name] = min(l)
TypeError: unorderable types: bool() < str()
2019-02-08 23:25:15,945 144 CRITICAL openupgrade-to-11 odoo.service.server: Failed to initialize database `openupgrade-to-11`.
Traceback (most recent call last):
  File "/odoo/src/odoo/service/server.py", line 932, in preload_registries
    registry = Registry.new(dbname, update_module=update_module)
  File "/odoo/src/odoo/modules/registry.py", line 85, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/odoo/src/odoo/modules/loading.py", line 463, in load_modules
    migrations.migrate_module(package, 'end')
[...]

Issue #133, opened 2018-12-14
map_values cannot properly map the 3 states of a Postgresql Boolean column


rvalyi:
In Postgres, a Boolean can have 3 states:
true, false and unknown: https://www.postgresql.org/docs/10/datatype-boolean.html

However in in the code of the map_values function we have:
```
    for old, new in mapping:
        new = "'%s'" % new

        if old is True:
            old = 'NOT NULL'
            op = 'IS'
        elif old is False:
            old = 'NULL'
            op = 'IS'
        else:
            old = "'%s'" % old
            op = '='

        values = {
            'table': table,
            'source': source_column,
            'target': target_column,
            'old': old,
            'new': new,
            'op': op,
        }

        if write == 'sql':
            query = """UPDATE %(table)s
                       SET %(target)s = %(new)s
                       WHERE %(source)s %(op)s %(old)s""" % values
        else:
            query = """SELECT id FROM %(table)s
                       WHERE %(source)s %(op)s %(old)s""" % values
        logged_query(cr, query, values)
```

So if we want to do a mapping such as defined here:

```
def map_product_attribute_create_variant(cr):
    openupgrade.map_values(
        cr,
        openupgrade.get_legacy_name('create_variant'),
        'create_variant',
        [(False, 'no_variant'),
         (None, 'no_variant'),
         (True, 'always'),
         ],
        table='product_attribute', write='sql')
```
( see https://github.com/Eficent/OpenUpgrade/blob/12.0-mig-product/addons/product/migrations/12.0.1.2/post-migration.py the branch has errors which I am fixing but it can be used as reference to understand)

We cannot because:

1. False will be transformed to IS NULL (which different of False for a Boolean Postgres type)
2. We have no way to use the NULL value as the source to be mapped

cc @mreficent (I used you branch to discover this issue, I don't know how you worked around it)

rvalyi:
@mreficent BTW with your original post-migration script I get this error with the None mapping:
```
  File "/odoo/src/addons/product/migrations/12.0.1.2/post-migration.py", line 15, in map_product_attribute_create_variant
    table='product_attribute', write='sql')
  File "/src/openupgradelib/openupgradelib/openupgrade.py", line 1219, in map_values
    logged_query(cr, query, values)
  File "/src/openupgradelib/openupgradelib/openupgrade.py", line 929, in logged_query
    cr.execute(query, args)
  File "/odoo/src/odoo/sql_db.py", line 148, in wrapper
[...]

PR #139, opened 2019-03-15
[FIX] openupgradelib: TyperError when trying to join and use min() function in bool with strings values

zaoral:
When try to merge two records of a model that have a text fields: one of records has a content but the other one is empty this error raised:

```python
TypeError: sequence item 0: expected str instance, bool found
```

The line `vals[field.name] = ' | '.join(l)` in openupgradelib._adjust_merged_values_orm() is trying to join something like `['text content', False]` and that generate the TypeError exception. Here is part of the code

```python
        l = all_records.mapped(field.name)
        if field.type in ('char', 'text', 'html'):
            if not op:
                op = 'other' if field.type == 'char' else 'merge'
            if op == 'merge':
                vals[field.name] = ' | '.join(l)
```
[source here](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_merge_records.py#L331)

This can be reproduced by trying to merge two records of Task Stages (project.task.type) model.
There we have the  'description' field which is of type text.

* Stage A with description field empty (do not set) 
* Stage B with description field with actually content 'example text'

We can also reproduce it easily by doing the next test in a python 3.5 environment

```python
>>> l = ['example text', False]
>>> ' | '.join(l)
  File "<input>", line 1, in <module>
    ' | '.join(l)
TypeError: sequence item 1: expected str instance, bool found
```

I believe this change also fix the error reported in #138. There the line that throws the error is because we have a list of strings and bool values that can not be manage by the min() python method (NOTE: This errors only raise in python 3.5 version, if we run the same python test in a python 2.7 environment then the min() will throw no error, the result will be `False`.)
 
```python
>>> l = ['example text', False]
>>> min(l)
  File "<input>", line 1, in <module>
    min(l)
TypeError: unorderable types: bool() < str()
```

Please consider this PR and let me know if need to add something else to the PR

Best Regards

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -328,6 +328,7 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
             if not op:
                 op = 'other' if field.type == 'char' else 'merge'
             if op == 'merge':
+                l = filter(lambda x: x is not False, l)
                 vals[field.name] = ' | '.join(l)
         elif field.type in ('integer', 'float', 'monetary'):
             if not op:
@@ -347,6 +348,8 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
             elif op == 'or':
                 vals[field.name] = functools.reduce(lambda x, y: x | y, l)
         elif field.type in ('date', 'datetime'):
+            if op:
+                l = filter(lambda x: x is not False, l)
             op = op or 'other'
             if op == 'max':
                 vals[field.name] = max(l)
@@ -355,16 +358,18 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
         elif field.type == 'many2many':
             op = op or 'merge'
             if op == 'merge':
+                l = filter(lambda x: x is not False, l)
                 vals[field.name] = [(4, x.id) for x in l]
         elif field.type == 'one2many':
             op = op or 'merge'
             if op == 'merge':
+                l = filter(lambda x: x is not False, l)
                 o2m_changes += 1
                 l.write({field.inverse_name: target_record_id})
         elif field.type in ('binary', 'many2one'):
             op = op or 'merge'
             if op == 'merge':
-                l = [x for x in l if x]
+                l = filter(lambda x: x, l)
                 if not getattr(target_record, field.name) and l and not \
                         vals.get(field.name):
                     vals[field.name] = l[:1]


pedrobaeza:
Why this?

What happen if a model doesn't have `name` field?

zaoral:
@pedrobaeza 
This PR aims to solve the error reported in https://github.com/OCA/openupgradelib/issues/138

pedrobaeza:
Well, check my other question. This doesn't solve at all the problem.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -323,7 +323,7 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
         if not field.store or field.compute or field.related:
             continue  # don't do anything on these cases
         op = field_spec.get(field.name, False)
-        l = all_records.mapped(field.name)
+        l = all_records.filtered('name').mapped(field.name)
```
pedrobaeza: In any case, this should be `filtered(field.name)`, not `filtered('name')`

mreficent:
I will do the correct fix.

pedrobaeza:
Yes, I had that sensation that filtering is not the option, but amend the operations on each level.

openupgradelib/openupgrade_merge_records.py
zaoral: thanks!
changed

zaoral:
@mreficent  Excellent. Thank you!

I have updated this PR in order that it helps to solve the two cases operations describe in the description of the PR without affecting the case you expose,   

If you are going to do more changes fell free to update this branch, also le me know if there is something I can do to help.

Best Regards

mreficent:
Here https://github.com/OCA/openupgradelib/pull/140 are handled all cases plus a minor optimization for one2many cases.

pedrobaeza:
@mreficent, @zaoral deserves to have the credits for this (or co-credits), so please add an extra commit for missing things here or tell her what to do.

mreficent:
@pedrobaeza yeah, right, no problem with the credits. I was too fast that didn't thought on that.

mreficent:
I think that maybe `manymany` also _needs_ the filter.

zaoral:
Ready, the requested changes has been applied. @mreficent @pedrobaeza please review 

NOTE: Ohh well.. thanks for the credits! The intention is just contribute for the community

Best Regards

zaoral:
> I think that maybe `manymany` also _needs_ the filter.

I agreed, also can be applied in order to have the same change for ('binary', 'many2one') in this line "l = [x for x in l if x"

openupgradelib/openupgrade_merge_records.py
```diff
@@ -347,24 +348,27 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
+            l = filter(lambda x: x is not False, l)
             op = op or 'merge'
             if op == 'merge':
                 o2m_changes += 1
                 l.write({field.inverse_name: target_record_id})
         elif field.type in ('binary', 'many2one'):
             op = op or 'merge'
             if op == 'merge':
-                l = [x for x in l if x]
+                l = filter(lambda x: x is not False, l)
```
mreficent: after thinking a bit, I think here should be `filter(lambda x: x, l)` instead, because here you don't want empty records.

rvalyi:
not sure if it is related but I also reported the following issue with OpenUpgrade, might be related https://github.com/OCA/openupgradelib/issues/133

openupgradelib/openupgrade_merge_records.py
zaoral: Then, Do you think is good Idea then separate binary and many2one section one if for each type? Will be something like

* at many2one use the filter(lambda x: x, l)
* at binary use the filter(lambda x: x is not False, l)

mreficent: I think it's no needed to separate them because you won't have never a 0 binary file, but feel free to separate them if you want

zaoral: well, is seems not necessary then.

pedrobaeza: Then everything is correct now in this PR, @mreficent @zaoral ?

zaoral:
@mreficent @pedrobaeza 

The suggested changes were applied.
I think we are prepare to merge

Best Regards

openupgradelib/openupgrade_merge_records.py
mreficent: No, I think here we should put the filter(lambda x: x, l) which is more strict

mreficent: please @zaoral

zaoral: @mreficent  Ready, the change you requested have been applied and pushed.

mreficent: Good, thanks

openupgradelib/openupgrade_merge_records.py
```diff
@@ -325,6 +325,7 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
         op = field_spec.get(field.name, False)
         l = all_records.mapped(field.name)
         if field.type in ('char', 'text', 'html'):
+            l = filter(lambda x: x is not False, l)
```
pedrobaeza: Do this inside `if op == 'merge'` for not losing performance if not needed

openupgradelib/openupgrade_merge_records.py
```diff
@@ -347,24 +348,27 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
             elif op == 'or':
                 vals[field.name] = functools.reduce(lambda x, y: x | y, l)
         elif field.type in ('date', 'datetime'):
+            l = filter(lambda x: x is not False, l)
```
pedrobaeza: You can check before if the `op` is going to apply any operation for performance reason

openupgradelib/openupgrade_merge_records.py
```diff
@@ -347,24 +348,27 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
             op = op or 'other'
             if op == 'max':
                 vals[field.name] = max(l)
             elif op == 'min':
                 vals[field.name] = min(l)
         elif field.type == 'many2many':
+            l = filter(lambda x: x is not False, l)
```
pedrobaeza: Do it inside `if`

openupgradelib/openupgrade_merge_records.py
```diff
@@ -347,24 +348,27 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
             op = op or 'merge'
             if op == 'merge':
                 vals[field.name] = [(4, x.id) for x in l]
         elif field.type == 'one2many':
+            l = filter(lambda x: x is not False, l)
```
pedrobaeza: Do it inside `if`

zaoral:
@pedrobaeza 

Thank you for your review.
I have updated the PR with the requested changes.
Please review and let me know if everything is ok.

Best Regards

pedrobaeza:
Thanks all

pedrobaeza:
I had to fix this for many2one/one2many:

94c6852ca23b85a8394cc6109ffdd48e4121402a

mreficent:
Thanks @pedrobaeza , that part was tricky. Good in separating the binary case! :)

PR #140, opened 2019-03-15
[FIX] openupgradelib: handle null columns

mreficent:
Operations like ''.join('', False) or min('', False) are wrong. We need in some cases filter the null value by checking if is strictly False or not. Remember: it's not the same `x = False` than `x is False` (because 0 or [] are untrue but are not strictly a False)


pedrobaeza:
@mreficent, @zaoral has update its PR #139 with the proposed solution, so please review that one instead.

PR #141, opened 2019-03-27
[FIX] openupgradelib: adapt StringIO to python3

mreficent:
In Odoo doc, it has:

```rst
Python 3 reorganised, moved or removed a number of modules in the standard
library:

* ``StringIO`` and ``cStringIO`` were removed, you can use ``io.BytesIO`` and
  ``io.StringIO`` to replace them in a cross-version manner (``io.BytesIO``
  for binary data, ``io.StringIO`` for text/unicode data).
* ``urllib``, ``urllib2`` and ``urlparse`` were redistributed across
  ``urllib.parse`` and ``urllib.request``.
```

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -24,10 +24,6 @@ import os
 import inspect
 import uuid
 import logging
-try:
-    from StringIO import StringIO
-except ImportError:
-    from io import StringIO
 from contextlib import contextmanager
 try:
     from psycopg2 import errorcodes, ProgrammingError, IntegrityError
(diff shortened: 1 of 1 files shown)


mreficent:
Well, `StringIO` and `io.StringIO` are different packages. And for python3 versions (v11, v12) we should assure we are using the later one.

pedrobaeza:
As said, previous code handled that already. That's why the try and ImportError.

mreficent:
So, in python3, in this `try` import, you always will go in the `except` part?

pedrobaeza:
Yes, that's the idea.

mreficent:
ok

PR #145, opened 2019-04-24
[FIX] rename_models

mreficent:
In v11, the `model_name` field of `ir_act_server` became stored.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -545,17 +545,24 @@ def rename_models(cr, model_spec):
                     old, new)
         _old = old.replace('.', '_')
         _new = new.replace('.', '_')
-        cr.execute('UPDATE ir_model SET model = %s '
-                   'WHERE model = %s', (new, old,))
-        cr.execute('UPDATE ir_model_data SET model = %s '
-                   'WHERE model = %s', (new, old,))
-        cr.execute(
-            'UPDATE ir_model_data SET name=%s '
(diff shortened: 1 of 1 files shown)


mreficent:
But in the pre-migration of `base` in v11, the column `model_name` will not be created yet, right? 🤔

astirpe:
Thanks!
Same issue affects `mail.template`: the field `model` (Varchar) is not updated with the value of `model_id.name`.

I guess there will be other similar cases in other modules.

mreficent:
Well, I will not check for other cases. If you want to check, then check it, and I will update accordingly 😉

pedrobaeza:
This shouldn't take into account general data model, not specific modules models. `mail` can be an exception as it's a very base module, but you should always check any way if the model/table/column exists. You can't assume it exists.

The rest of the changes should be done on specific migration scripts for that module.

mreficent:
Just used table_exists and column_exists everywhere XD

mreficent:
he he he but you could have done that before  🙄

pedrobaeza:
Well, we are touching now this, so it's when it raises.

mreficent:
Ok. But should I put the `logged_query` in all of them or just in some specific ones? I think that putting on all of them is too much log.

pedrobaeza:
As said about the rule: in all of them that makes a massive change. Don't worry about being too much: log is something optional that you can even disable, and in this case the verbosity is thankful knowing the number of records modified for tracking possible problems.

mreficent:
done, logged all :)

PR #148, opened 2019-05-09
[IMP] Add XML conversion framework and Bootstrap 3 to 4 converter

Yajo:
In `openupgrade_tools`, I added some methods to perform transformation of XML strings based on given specs. The specs support some HTML-specific features, such as classes and styles. I provide also a handy shortcut for spec declaration.

I also added `openupgrade_120` with the spec for the Bootstrap 3 to 4 migration and a simple method that leverages the framework and spec to convert an XML string.

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade_120.py
+++ b/openupgradelib/openupgrade_120.py
@@ -1,0 +1,397 @@
+# Copyright 2019 Tecnativa - Jairo Llopis
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+"""Tools specific for migrating Odoo 11.0 to 12.0.
+
+Important changes come from the update from Bootstrap 3 to 4.
+To understand these changes, these sources have been consulted, among others
+specific to some rules that can be consulted in inline comments:
+
+* http://upgrade-bootstrap.bootply.com/
+* https://bit.ly/2xmUHmo
+* https://getbootstrap.com/docs/4.3/migration/#stable-changes
+* https://github.com/odoo/odoo/wiki/Tips-and-tricks:-BS3-to-BS4
+
+Don't expect perfection. But patches are welcome.
+"""
+
+from itertools import product
+from psycopg2.extensions import AsIs
+
+from .openupgrade_tools import (
+    convert_html_fragment,
+    convert_html_replacement_class_shortcut as _r,
+)
+
+
+_COLS = range(1, 13)
+_CONTEXTS = (
+    # (BS3 context, BS4 context)
+    ("default", "secondary"),
+    ("primary", "primary"),
+    ("active", "active"),
+    ("disabled", "disabled"),
+    ("success", "success"),
+    ("info", "info"),
+    ("warning", "warning"),
+    ("danger", "danger"),
+)
+_TIERS = (
+    # (BS3 tier, BS4 tier)
+    ("-lg", "-xl"),
+    ("-md", "-lg"),
+    ("-sm", "-md"),
+    ("-xs", ""),
+)
+_BS3_VISIBLES = ("block", "inline", "inline-block")
+
+# These replacements are from standard Bootstrap 3 to 4
+_BS4_REPLACEMENTS = (
+    # Convert columns and modifiers among tiers
+    *(_r("col%s-%s" % (t3, col), "col%s-%s" % (t4, col))
+      for (t3, t4), col in product(_TIERS, _COLS)),
+    *(_r("col%s-offset-%s" % (t3, col), "offset%s-%s" % (t4, col))
+      for (t3, t4), col in product(_TIERS, _COLS)),
+    *(_r("col%s-pull-%s" % (t3, col), "order-first")
+      for (t3, t4), col in product(_TIERS, _COLS)),
+    *(_r("col%s-push-%s" % (t3, col), "order-last")
+      for (t3, t4), col in product(_TIERS, _COLS)),
+
+    # Typography
+    _r(selector="blockquote", class_add="blockquote"),
+    _r(selector="blockquote > small", class_add="blockquote-footer"),
+    _r("blockquote-reverse", "blockquote text-right"),
+    _r(selector=".list-inline > li", class_add="list-linline-item"),
+
+    # .page-header dropped. See https://stackoverflow.com/a/49708022/1468388
+    _r("page-header", "pb-2 mt-4 mb-2 border-bottom"),
+
+    # <dl> & co. See https://stackoverflow.com/a/56020841/1468388
+    _r("dl-horizontal", "row"),
+    _r(selector=".dl-horizontal > dt",
+       class_add="col-sm-3 text-sm-right"),
+    _r(selector=".dl-horizontal > dd", class_add="col-sm-9"),
+
+    # Images
+    _r("img-circle", "rounded-circle"),
+    _r("img-responsive", "img-fluid d-block"),
+    _r("img-rounded", "rounded"),
+
+    # Tables
+    _r("table-condensed", "table-sm"),
+    *(_r("%s" % c3, "table-%s" % c4, selector=".table .%s" % c3)
+      for (c3, c4) in _CONTEXTS),
+
+    # Forms
+    _r("control-label", "col-form-label"),
+    _r("form-group-lg", "form-control-lg"),
+    _r("form-group-sm", "form-control-sm"),
+    _r("input-lg", "form-control-lg"),
+    _r("input-sm", "form-control-sm"),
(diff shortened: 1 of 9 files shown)


Yajo:
Please notice this is WIP. I haven't tested yet. I have to open a PR to OpenUpgrade and start testing.

openupgradelib/openupgrade_120.py
```diff
@@ -0,0 +1,208 @@
...
+
+    # Labels and badges
+    _r("label", "badge"),
+    _r("badge-default", "badge-secondary"),
+    *(_r("label-%s" % c3, "badge-%s" % c4) for (c3, c4) in CONTEXTS),
+
+    # Convert panels, thumbnails and wells to cards
+    _r("panel", "card"),
+    _r("panel-body", "card-body"),
+    _r("panel-default panel-group"),
```
mreficent: You need to put `", "` here.

**Remark:** Also, as an overall comment for all this code, you should also consider what Odoo did when it changed to BS4. I mean, here in openupgradelib we should not have a pure BS3 to BS4 converter, here we should have a Odoo BS3 to Odoo BS4 converter instead.

As an example of what I mean, Odoo didn't do:
```
_r("panel-default", "panel-group"),
```
What Odoo did was:
```
_r("panel-collapse", ""),
_r("panel-default", ""),
_r("panel-group", "accordion"),
```

pedrobaeza:
Added method `convert_field_bootstrap_3to4` for converting an entire field.

pedrobaeza:
Example in https://github.com/OCA/event/pull/146/commits/415224a3b7ecb850c1ee2220fa461002bda9a795

openupgradelib/openupgrade_120.py
```diff
@@ -0,0 +1,259 @@
...
+    :param model_name: Name of the model that contains the field.
+    :param field_name: Name of the field that contains the BS3 HTML content.
+    :param domain: Optional domain for filtering records in the model
+    :param method: 'orm' (default) for using ORM; 'sql' for avoiding problems
+        with extra triggers in ORM.
+    """
+    if domain is None:
+        domain = []
+    Model = env[model_name]
+    recs = Model.with_context(active_test=False).search(domain)
```
Yajo: You're using ORM to search before checking if you have to use ORM. What if this is being executed in a pre script where the model is new and still doesn't exist in the `env`?

IMHO it would be better to provide a list of ids instead of a domain, so they can be browsed fine equally with or without orm.

If you agree, I can patch it. I still have local WIP changes.

pedrobaeza: That can't happen, as you don't need to convert a non-existing field: you only convert fields of an existing model that are in BS3. Any way, you can add the SQL part for being totally coherent with the param `method`.

Yajo: I mean: imagine a module adds model X with field `html_description`. You can't find it through `Model = env[model_name]` because model X still doesn't exist in the pre stage. It would fail also when using raw SQL mode.

pedrobaeza: The same explanation: you don't need to convert a non existing field, but as said, just unfold this search to be done in SQL when using `method='sql'`.

Yajo: I found easier to split in 2 methods.

pedrobaeza: Well, 3 then: the main one and the 2 submethods, but I prefer this if. It's better for following the flow IMO.

pedrobaeza:
As said, please keep the unique external entry point and let the other 2 methods as private, and DON'T remove the extra checks for avoiding:

- unneeded conversions
- unneeded writes

Yajo:
Done as you please.

openupgradelib/openupgrade_120.py
mreficent: for `_r("panel-default", "")` and `_r("panel-group", "accordion")`, you can find examples here:
https://github.com/odoo/odoo/commit/59237ea81988bdcd0cc746422a7c94bf2eadaaa2#diff-24b0433302053618c5b742f3eded1d67

As for the `panel-collapse`, I found that in general the change is `_r("panel-collapse", "collapse")` but in some cases is `_r("panel-collapse", "")`. The same goes to wells, sometimes  is `_r("well", "card card-body")`, but sometimes is  `_r("well", "")`.

Yajo: I took that into account, although the goal here is not mainly to format widgets as the new widgets, but to format them as close as possible to how they were before.

If a user wants to update the look and feel of his website, usually he should drag and drop a new snippet and modify its content, and remove the old one.

Thanks!

pedrobaeza: @Yajo but you should have something into account: old snippets must be editable the same with new version, so if a snippet in v11 was for example image gallery, on v12 it should still be recognized by web_editor as image gallery. Take that into account.

Yajo:
This is ready to review now.

A sample here:

| v11 | v12 |
| --- | --- |
| Screenshot_2019-05-20 Home | Screenshot_2019-05-20 Home Website localhost |

All done using standard snippets.

openupgradelib/openupgrade_120.py
mreficent: @pedrobaeza exactly, that was the intention of my remark, good explanation 😉

@Yajo still you should add, at least, the `", "` in that indicated line :)

Yajo: > old snippets must be editable the same with new version

Yes, they are, I just meant that it's more important to keep the visual changes minimal. Users visit their pages more often than they modify them.

I'll double-check this manually however. Set this as WIP again in the mean time, please.



> you should add, at least, the `", "` in that indicated line :)

Well, with that I'd be replacing the `panel-default` class for a `panel-group`, but quoting from the [BS migration docs](https://getbootstrap.com/docs/4.3/migration/#panels):

> - `.panel-default` removed and no replacement.
> - `.panel-group` removed and no replacement. `.card-group` is not a replacement, it is different.

So, my intention was to just drop both. I have to make a change, but it's different; thanks for the pointer anyways. :blush:

Yajo: Thanks for your comments, there were indeed some usability glitches, they are polished now. :wink:  I'm gonna upload some gifs, down in the main thread.

Yajo:
This is ready to review, and has been much polished. Let me share some gifs for all blocks in v11 when migrated to v12:

<details>
<summary>Gimme those gifs! :framed_picture: </summary>

### Title
title

### Text image / image text
text image

### Text block
text block

### Slider block
slider block

### References
references

### Parallax and quotes slider
parallax and quotes slider

### Image galleries
image galleries

### Image and text
image text

### Features grid (includes inner content snippets)
features grid with inner content

### Features block
features block

### Cover
cover

### Comparisons
comparisons

### Company team
company team

### Big picture
big picture block

### Big message
big message

### Accordion
accordion

### Three columns
3 cols

</details>

pedrobaeza:
Please fix:
```
/home/travis/build/OCA/openupgradelib/openupgradelib/openupgrade_120.py:52:5: E999 SyntaxError: invalid syntax
/home/travis/build/OCA/openupgradelib/openupgradelib/openupgrade_120.py:305:80: E501 line too long (87 > 79 characters)
```

Yajo:
> `/home/travis/build/OCA/openupgradelib/openupgradelib/openupgrade_120.py:52:5: E999 SyntaxError: invalid syntax`

This one is a bug in the linter. The file `openupgrade_120` is python3-only. I can imagine no situations where one would need to import this file in a python 2 runtime, so this should be taken care of in the linter itself.

mreficent:
🤔 Maybe you could add in the beginning of the `openupgrade_120` file:
```python
if sys.version_info[0] < 3:
    return
```
or something similar.

Yajo:
Thanks @mreficent but I think it wouldn't fix it.

I cannot `return` from outside a function, and anyways the linter would find the syntax error; also I wouldn't like to add `# noqa: E999` to each line, because there are a lot of lines that would fail (all those starting with `*_r(`.

I tried a different approach: linting under python 3. Let's see...

Yajo:
So, should we merge? I guess this is still not perfect, but as migrations get done we can polish it...

pedrobaeza:
Yes, let's go! Please rebase and adapt pending PRs in OpenUpgrade

pedrobaeza:
@Yajo please squash your last 3 commits into one.

Yajo:
I squashed into 2: 1st one with linter changes and 2nd one with actual new features.

pedrobaeza:
Please adjust depending PRs

PR #152, opened 2019-06-18
Support multilang updates in Bootstrap 3 to 4 migration

### LINKED ISSUE #151: convert_field_bootstrap_3to4: Should convert also translations of the field if any
The method must check if it's a translatable field (and what kind of translatable field - although in the end, it will mean more or only one `ir.translation` records -), and perform the conversion also on that translations. Example, field `website_description` in `product.template`.

Yajo:
Fix #151.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ sudo: false
 python:
   - "3.6"  # But switching to 2.7 in the script
 
+services:
+  - postgresql
+
 addons:
     apt:
         packages:
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -148,6 +148,7 @@ __all__ = [
     'chunked',
     'drop_columns',
     'delete_model_workflow',
+    'update_field_multilang',
     'update_workflow_workitems',
     'warn_possible_dataloss',
     'set_defaults',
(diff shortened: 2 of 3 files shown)


pedrobaeza:
@moylop260 can you help to solve the problem I have with Travis? I don't see where `pytest` is instructed to be installed, and it's provoking a version conflict with `six`, but I also don't see where this one is installed.

Yajo:
AFAICS [the main branches test on Ubuntu 14.04](https://travis-ci.org/OCA/openupgradelib/jobs/542363869#L19) while [this PR is being tested in 16.04](https://travis-ci.org/OCA/openupgradelib/jobs/547225726#L19). I imagine Travis started switching the base OS depending on the commit date, because I rebuilt today the master branch and it went for 14.04 also.

That's the main difference I can notice, and it results in different versions of several packages, including pip itself, which probably makes those warnings appear. We're actually gonna have a hard time testing after 2020, where py2 support ends.

Nevermind, it seems that [in Ubuntu 16.04 postgres is not ready out of the box unless you list it in services](https://docs.travis-ci.com/user/reference/xenial/#databases-and-services), so I did that. Let's see.

Yajo:
Green! :green_heart:

PR #160, opened 2019-07-23
[IMP] Improvements to website migration

Yajo:
Without this patch, if an XML node had a `style="height:333px"` and the rules set it to `style="height:222px"`, there was no way to avoid that.

Now, if you *really* want to replace a style or attribute, then just add it to the list of styles/attributes to remove. Otherwise, it will only be added if it was missing before.

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade_tools.py
+++ b/openupgradelib/openupgrade_tools.py
@@ -104,6 +104,18 @@ def convert_xml_node(node,
                      wrap=""):
     """Apply conversions to an XML node.
 
+    All parameters except :param:`node` can be a callable that return the
+    expected type as specified in each of them below.
+
+    The callable would be called with these **keyword-only** arguments:
+
+    * ``attrs``: A ``dict`` of the original attributes in the node.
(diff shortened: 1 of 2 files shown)


Yajo:
I'll add into this PR several changes needed to polish the v12 website migration.

Please somebody set as WIP.

openupgradelib/openupgrade_tools.py
```diff
@@ -142,14 +160,35 @@ def convert_xml_node(node,
...
     class_rm = set(class_rm.split())
     style_add = style_add or {}
+    # Obtain attributes, classes and styles
+    classes = set(node.attrib.get("class", "").split())
+    styles = node.attrib.get("style", "").split(";")
+    styles = {key.strip(): val.strip() for key, val in
+              (style.split(":", 1) for style in styles if ":" in style)}
+    # Convert incoming callable arguments into values
+    originals = {
+        "attrs": node.attrib.copy(),
```
mreficent: AttributeError: 'lxml.etree._Attrib' object has no attribute 'copy'

PR #162, opened 2019-07-25
[IMP] update_module_names: rename instead of merge case

mreficent:
Do a rename instead of merge if new_name doesn't exist yet. It may happens that you do a merge from an installed module to an uninstalled module, but the later is a new module (and thus is not yet in the ir.module.module list) and it is not installed.

Example:

In v11, you have customer_activity_statement and customer_outstanding_statement.
In v12, you have partner_statement, that both previous modules merge into.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1002,7 +1002,10 @@ def update_module_names(cr, namespec, merge_modules=False):
         of just a renaming.
     """
     for (old_name, new_name) in namespec:
-        if merge_modules:
+        query = "SELECT id FROM ir_module_module WHERE name = %s"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Travis is pretty stuck due to a massive change tonight. I'm cancelling builds, but there are still a lot pending. Commit example: https://github.com/OCA/server-auth/commit/be9c36ba19ed0ff9923d96a24d77d6285fcdffc6

sbidoul:
@pedrobaeza yeah I just noticed it too... not sure yet why it happened.

pedrobaeza:
Please rebase and check the problem.

mreficent:
I think the error is unrelated with this PR.

PR #166, opened 2019-08-13
[ADD] copy_fields_multilang method

Yajo:
This method allows to copy data from some model's fields to other (or the same) model's equivalent fields, and transfers translated values.

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -53,6 +53,7 @@ except ImportError:
             while self._cms:
                 self._cms.pop().__exit__(exc_type, exc_value, traceback)
 
+from psycopg2 import sql
 from psycopg2.extensions import AsIs
 from lxml import etree
 from . import openupgrade_tools
@@ -139,6 +140,7 @@ __all__ = [
     'load_data',
     'add_fields',
     'copy_columns',
+    'copy_fields_multilang',
     'rename_columns',
     'rename_fields',
     'rename_tables',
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -376,6 +377,108 @@ def copy_columns(cr, column_spec):
...
+        If empty, it will default to :param:`destination_table`.
+
+    :param str source_table:
+        Name of the source table where the data will be copied from.
+        If empty, it will default to :param:`destination_table`.
+        It must be ``env[source_model]._table``.
+
+    :param list source_columns:
+        List of column names in the :param:`source_table` that will
+        provide the copied data.
```
pedrobaeza: Missed "If empty, it will default to :param:`destination_columns`."

Yajo: Fixed, thanks.

openupgradelib/openupgrade.py
```diff
@@ -376,6 +377,109 @@ def copy_columns(cr, column_spec):
...
+    if source_columns is None:
+        source_columns = destination_columns
+    cols_len = len(destination_columns)
+    assert len(source_columns) == cols_len > 0
+    # Basic copy
+    query = (
+        "UPDATE %(dest_t)s AS dt SET " +
+        ", ".join("dt.%(dest_col_{0})s = st.%(src_col_{0})s".format(n)
+                  for n in range(cols_len)) +
+        "FROM %(src_t)s AS st WHERE dt.%(rel_col)s = st.id"
```
pedrobaeza: Missing initial space

Yajo: Fixed, thanks.

openupgradelib/openupgrade.py
```diff
@@ -969,7 +1081,7 @@ def write_value(ids, field, value):
                 write_value(ids, field, value)
 
 
-def logged_query(cr, query, args=None, skip_no_result=False):
+def logged_query(cr, query, args=(), skip_no_result=False):
```
pedrobaeza: Don't change `args=None` to `args=()` or you will have the problem with mutable variables:

http://effbot.org/zone/default-values.htm

openupgradelib/openupgrade.py
```diff
@@ -979,17 +1091,15 @@ def logged_query(cr, query, args=None, skip_no_result=False):
...
     try:
         cr.execute(query, args)
     except (ProgrammingError, IntegrityError):
-        logger.error('Error running %s' % cr.mogrify(query, args))
+        logger.error('Error running %s' % cr._obj.query.decode())
         raise
     if not skip_no_result or cr.rowcount:
-        logger.debug('Running %s', query % args)
-        logger.debug('%s rows affected', cr.rowcount)
+        logger.debug('%s rows affected after running %s',
```
pedrobaeza: Why are you changing this base method that is working since a lot and with your changes can have side effects across all the versions that exists?

openupgradelib/openupgrade.py
Yajo: But tuples are immutable... :thinking:

pedrobaeza: And then any overwriting without check will do this to fail while trying to append. This expects to be `None` or a list, not a tuple. Why don't you simply keep as it was?

openupgradelib/openupgrade.py
Yajo: Yes, let me explain.

I found the [`psycopg2.sql`](http://initd.org/psycopg/docs/sql.html) module with lots of great features for a SQL-composing library such as openupgradelib, and I'm using it in the new method being added.

However, you can't just do `query % args` as it was being done before, because these are sql-specific objects, not raw strings, and don't support that. BTW, that was misleading because the logged query was not exactly the query that was being sent to the database. [As explained in psycopg2 docs](http://initd.org/psycopg/docs/usage.html?highlight=never#sql-injection):

> Warning
> Never, **never, NEVER** use Python string concatenation (+) or string parameters interpolation (%) to pass variables to a SQL query string. Not even at gunpoint. 

A few lines above there was another logger that did it more accurately: `logger.error('Error running %s' % cr.mogrify(query, args))`. This returns the same string that is sent to Postgres server, so it's a better candidate for debugging. The problem with that one is that under Python 3 it's a `bytes()` object, which means that the log is not properly formatted (basically it prints a raw `\n` instead of a real newline character, making it very hard to read by humans), and that's why I'm using `.decode()` at the end, which works fine both in Python 2 and 3.

About why use `cr._obj.query` instead of `cr.mogrify(query, args))`, it's because it's a property already found in all cursors.

I checked it works from v8 to v12, but TBH didn't check that in v7 or lower.

Maybe I could use the mogrify + decode version, which should work fine always, regardless of wether that hidden `cr._obj.query` attribute exists or not.

Yajo: Oh and I forgot it: now that I was changing it, I didn't see a reason to use 2 log lines, so I reduced it to 1. Not a big change, though.

pedrobaeza: You should make use to explain all of this in advance in the commit message, so it's kept for later references. OK to the change, but please include all these explanations in the commit message.

About `mogrify` you can simply check the existence of the field with `hasattr` and if not, use it as fallback.

pedrobaeza:
Please check Travis

Yajo:
Ready to merge.

PR #168, opened 2019-08-28
[IMP] Renaming index instead of dropping it while renaming columns

pnajman-modoolar:
[IMP] For small tables it's perfectly acceptable to drop index and then let Odoo recreate it, but with large tables this scenario is rather costly.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -513,7 +513,10 @@ def rename_columns(cr, column_spec):
                         table, old, new)
             cr.execute(
                 'ALTER TABLE "%s" RENAME "%s" TO "%s"' % (table, old, new,))
-            cr.execute('DROP INDEX IF EXISTS "%s_%s_index"' % (table, old))
+            cr.execute(
+                'ALTER INDEX IF EXISTS "%s_%s_index" RENAME TO "%s_%s_index"' %
+                (table, old, table, new)
+            )
 
 
 def rename_fields(env, field_spec, no_deep=False):


pedrobaeza:
One of the reasons for removing the index is for cleaning a bit the database, so I would keep `drop` when the renaming is to `None`.

openupgradelib/openupgrade.py
```diff
@@ -507,16 +507,24 @@ def rename_columns(cr, column_spec):
...
             logger.info("table %s, column %s: renaming to %s",
                         table, old, new)
             cr.execute(
                 'ALTER TABLE "%s" RENAME "%s" TO "%s"' % (table, old, new,))
-            cr.execute(
-                'ALTER INDEX IF EXISTS "%s_%s_index" RENAME TO "%s_%s_index"' %
-                (table, old, table, new)
-            )
+
+            if rename_index:
```
pedrobaeza: Why not simply `if new is not None:` here?

pedrobaeza:
@hbrunn @StefanRijnhart what do you think about this?

StefanRijnhart:
@pedrobaeza This was my idea ;-) Petar is on my team here and we had to rename a column of a large table. I asked him to propose this to save us the time of rebuilding the index.

StefanRijnhart:
I don't agree with dropping the index if the column is moved out of the way for the upgrade process using None. It may still be used to select records during the migration process, and keeping the index may very well benefit the (now often dramatic) performance of the migration process. After a migration, the database must be cleaned up with `database_cleanup` upon which the indexes are removed with their columns.

pedrobaeza:
Uhm, I didn't think about that possibility, Stefan. Not sure how many times we are going to search over that renamed columns, but it's definitively a possibility, and at the end, this doesn't cost except for space.

The only potential problem I see, is that we kept indexes of fields that are not indexed anymore in new version. This can cause bottleneck problems at DB level for populating such index on each creation. I think that's the original reason for dropping it.

Maybe we can detect such situation in `openupgrade_records`... cc @mreficent

hbrunn:
or in `database_cleanup`, I made something for creating indexes for lower versions, so we can do the reverse in higher versions

openupgradelib/openupgrade.py
pnajman-modoolar: Because at that point `new` can never be `None` due to the code on line 513 where `new` get value of the legacy name or am I missing something?

pedrobaeza: Good point! Let's see the general discussion about dropping or not.

StefanRijnhart:
Odoo drops indexes that are not present anymore by itself. In fact, if you upgrade individual modules on start-up it can drop indexes that are defined in a different dependency chain.

pedrobaeza:
Great, then there's no more to discuss. Let's rename the index always.

@pnajman-modoolar sorry for making you to change this, but as you see, an interesting discuss has arisen from this. Please revert to original state.

PR #169, opened 2019-08-28
[IMP] update_module_names: rename instead of merge special case

apps2grow:
If there is an uninstalled module with the new name: delete the uninstalled module, and rename the old module.

Example:
v11: addons_path includes account-financial-tools, but account_asset_management is not installed
v12: account_asset -> account_asset_management

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1136,7 +1136,8 @@ def update_module_names(cr, namespec, merge_modules=False):
         of just a renaming.
     """
     for (old_name, new_name) in namespec:
-        query = "SELECT id FROM ir_module_module WHERE name = %s"
+        query = ("SELECT id FROM ir_module_module WHERE name = %s "
+        "AND state in ('to upgrade', 'to remove', 'to install', 'installed')")
         cr.execute(query, [new_name])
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This has been working with that use case since a lot, so you may have another problem than this. Maybe you install an old version of the module and was properly uninstalled? It seems so if you need to delete ir_model_data. Please try to migrate on a fresh database.

apps2grow:
I just updated all repos and openupgradelib, with the latest code of https://github.com/OCA/account-financial-tools/pull/873
In odoo11, I created database 11A, installed account_asset, created asset category and asset.
I duplicated 11A to 11B.
I successfully migrated 11A to v12.
In odoo11 configuration addons_path, I added oca/account-financial-tools.
In database 11B, I updated the apps list.
I "successfully" migrated 11B to v12, but **I cannot see the asset profile and the asset.** This was the case last time I tried too, and the reason for this PR.

I am actually not using oca/account-financial-tools in v11, so this PR is not important for me now, but I would like to make you aware of the behaviour.

pedrobaeza:
I don't fully understand the difference between database 11A and 11B.

pedrobaeza:
Closing as no more feedback.

Issue #172, opened 2019-09-12
rename_columns: Misfiring of index rename when >63 chars


thomaspaulb:
Indexes named according to the 'table_field_index' pattern are being renamed [here](https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L517). 

However when this leads to a very long (>63 chars) index name eg. [this one](https://github.com/OCA/OpenUpgrade/blob/10.0/addons/product/migrations/10.0.1.2/pre-migration.py#L13) leading to the command:

```
ALTER INDEX IF EXISTS "product_attribute_line_product_attribute_value_rel_val_id_index" RENAME TO "product_attribute_line_product_attribute_value_rel_product_attribute_value_id_index"
```

Then Postgres's [maximum identifier length](https://github.com/postgres/postgres/blob/master/src/include/pg_config_manual.h#L29) kicks in and the target name is now abbreviated to `product_attribute_line_product_attribute_value_rel_product_attr`, which can already exist - in this case because of the earlier rename for [product_attribute_line_id](https://github.com/OCA/OpenUpgrade/blob/10.0/addons/product/migrations/10.0.1.2/pre-migration.py#L12).

kos94ok-3D:
Hi, @thomaspaulb.
Can close it?

thomaspaulb:
@kos94ok-3D Yes it was solved.

PR #173, opened 2019-09-12
[FIX] workaround for faulty index renames

thomaspaulb:
Workaround for https://github.com/OCA/openupgradelib/issues/172

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -513,10 +513,13 @@ def rename_columns(cr, column_spec):
                         table, old, new)
             cr.execute(
                 'ALTER TABLE "%s" RENAME "%s" TO "%s"' % (table, old, new,))
-            cr.execute(
-                'ALTER INDEX IF EXISTS "%s_%s_index" RENAME TO "%s_%s_index"' %
-                (table, old, table, new)
-            )
+            old_index_name = "%s_%s_index" % (table, old)
+            new_index_name = "%s_%s_index" % (table, new)
+            if len(new_index_name) <= 63:
+                cr.execute(
+                    'ALTER INDEX IF EXISTS "%s" RENAME TO "%s"' %
+                    (old_index_name, new_index_name)
+                )
 
 
 def rename_fields(env, field_spec, no_deep=False):


pedrobaeza:
OK, then I the best path is to detect if m2m, and not dropping it in that case, but yes on > 64 and not m2m. What do you think?

yelizariev:
Related PR in odoo: https://github.com/odoo/odoo/pull/7619

openupgradelib/openupgrade.py
```diff
@@ -513,10 +513,13 @@ def rename_columns(cr, column_spec):
                         table, old, new)
             cr.execute(
                 'ALTER TABLE "%s" RENAME "%s" TO "%s"' % (table, old, new,))
-            cr.execute(
-                'ALTER INDEX IF EXISTS "%s_%s_index" RENAME TO "%s_%s_index"' %
-                (table, old, table, new)
-            )
+            old_index_name = "%s_%s_index" % (table, old)
+            new_index_name = "%s_%s_index" % (table, new)
+            if len(new_index_name) <= 63:
```
pedrobaeza: Shouldn't this check as well in `len(old_index_name) <= 63`?

diggy128: A check wouldn't hurt, but since the index is there, if it were > 63 chars wouldn't it be already be truncated?

pedrobaeza: Yes, and that's why the renaming would fail then, as it won't find the previous index name we are providing (as it's not truncated).

extrememicro: What about something like this:

            max_len = 63 - len(new + '_index') # PostgreSQL's Max Identifier Length Is 63 Bytes (NAMEDATALEN - 1)
            renamed_table = (table[:max_len]) if len(table) > max_len else table
            cr.execute(
                'ALTER INDEX IF EXISTS "%s_%s_index" RENAME TO "%s_%s_index"' %
                (table, old, renamed_table, new)
            )

pedrobaeza:
Any news on this?

thomaspaulb:
Will make time for this soon

mgielissen:
This PR works.

pedrobaeza:
My comment is still missing

openupgradelib/openupgrade.py
Yajo: > it won't find the previous index name we are providing (as it's not truncated).

Yes, it finds it because it is truncated before searching for it too. I tested that.

pedrobaeza: Are you talking that is truncated in Postgres although you pass a name over 63 chars?

Yajo: Yes

pedrobaeza: Then I don't get why we need to shorten new index name and not the old one...

Yajo: It's not shortening it; it's just testing if it is longer that what is supported and, if so, skipping the rename. The shortening is done by postgres automatically.

pedrobaeza:
OK, let's merge this as is as it seems it doesn't fail in my theoretical case.

PR #174, opened 2019-09-13
[FIX] add_fields: Respect sql_type and build safe query

### LINKED ISSUE #171: Buggy add_fields
Look at this code we have:

https://github.com/OCA/openupgradelib/blob/614161f47bc7c1592fa9985650b4cbc7429983a2/openupgradelib/openupgrade.py#L2076-L2086

As you can see, the `sql_type` is *always* ignored. However, the docstring of the method says:

https://github.com/OCA/openupgradelib/blob/614161f47bc7c1592fa9985650b4cbc7429983a2/openupgradelib/openupgrade.py#L2048-L2055

So, it should be:

```diff
-        sql_type = sql_type_mapping.get(field_type)
+        sql_type = sql_type or sql_type_mapping.get(field_type)
```

However I'm wondering if fixing this bug would mean breaking existing migration scripts out there in the wild who didn't realize they were working because of this bug... :thinking: 

Also, since the pypi package is almost never released, we don't have this ability of adding a new major version and introducing backwards-incompatible fixes...

So what do we do here?

Yajo:
- Until now, `sql_type` was being ignored. Fix #171 and respect it.
- Use safe sql builder. If, i.e. one column was called "select", previous implementation with `AsIs()` would fail.

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2078,11 +2078,15 @@ def add_fields(env, field_spec):
         # Add SQL column
         if not table_name:
             table_name = env[model_name]._table
-        sql_type = sql_type_mapping.get(field_type)
+        sql_type = sql_type or sql_type_mapping.get(field_type)
         if sql_type:
             logged_query(
-                env.cr, """ALTER TABLE %s ADD COLUMN %s %s""",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you check the error?

Yajo:
Done.

PR #175, opened 2019-09-18
[IMP] rename_models: rename the models in the reference fields

MiquelRForgeFlow:
This new method renames the models in the reference fields.

Also fixes _change_reference_refs_sql method.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -863,6 +863,33 @@ def rename_models(cr, model_spec):
                     "new_string": "%s," % new,
                 },
             )
+        # Handle models that reference to this model using reference fields
+        cr.execute("""
+             SELECT model, name
+             FROM ir_model_fields
+             WHERE ttype='reference'
+             """)
+        rows = cr.fetchall()
+        for row in rows:
+            table = row[0].replace('.', '_')
+            if not table_exists(cr, table):
+                continue
+            column = row[1]
+            if not column_exists(cr, table, column):
+                continue
+            logged_query(
+                cr, """
+                 UPDATE %(table)s
+                 SET %(column)s = replace(
+                    %(column)s, '%(old)s,', '%(new)s,')
+                 WHERE %(column)s LIKE '%(old)s,%%'
+                 """ % {
+                    "table": table,
+                    "column": column,
+                    "old": old,
+                    "new": new,
+                }, skip_no_result=True,
+            )
         # Update export profiles references
         logged_query(
             cr, "UPDATE ir_exports SET resource = %s WHERE resource = %s",


pedrobaeza:
Why an extra method instead of being part of existing `rename_models`? All should be done in the same method IMO.

pedrobaeza:
Please put the other fix in a separate PR

mreficent:
> Why an extra method instead of being part of existing `rename_models`?

Because the need of the `env`. `rename_models` doesn't use `env` but the new method needs it.

Maybe we can expand `rename_models` signature like:

```python
def rename_models(cr, model_spec, env=False):
```

what do you think?

pedrobaeza:
But I think we can avoid the use of the env. Let's search directly on `ir_model_fields` table possible fields of ttype='reference', and then perform the SQL with the change.

mreficent:
But how I obtain the `table`?

pedrobaeza:
Yeah, good point... Now you know why I always want to pass `env` although no initial use. I don't like too much the option of adding that extra "optional" argument. Or at least not only. It's better to build a new environment for cr (we can do it the same as we do on `migrate` decorator) if not provided via that argument.

mreficent:
I think this method should remain separated and not merged into the `rename_models` method.
_Reason:_ `rename_models` is called in **pre**-migration and covers all cases without using the env. Instead, `rename_models_reference` should be called in **end**-migration, because it's only in end-migration when all models are in the env, and thus, only in end-migration will you obtain any model table, by the same way that merge_records always is called in end-migration.

MiquelRForgeFlow:
I have updated the PR. Now the `rename_models` method make use of it. And also it can be called independently in the end-migration with the environment.

MiquelRForgeFlow:
> instead of being part of existing rename_models? All should be done in the same method IMO.

Now it's handled this way. Very simple.

openupgradelib/openupgrade.py
```diff
@@ -863,6 +863,33 @@ def rename_models(cr, model_spec):
...
             )
+        # Handle models that reference to this model using reference fields
+        cr.execute("""
+             SELECT model, name
+             FROM ir_model_fields
+             WHERE ttype='reference'
+             """)
+        rows = cr.fetchall()
+        for row in rows:
+            table = row[0].replace('.', '_')
```
pedrobaeza, line 874: This is still not correct, as not all models has that table name. Maybe you can trace from the field to the table that hosts it.

MiquelRForgeFlow: IMO, I have three reasons to proceed this way, and I think you should accept:

 - I am doing the same as https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L783. If that line is acceptable in the same method, mine should also be acceptable.
 - Moreover, I am calling a "table_exists" to check if that table exists.
 - Moreover, nobody uses reference fields in custom modules. Only few models in odoo community have this kind of field and all of them their table is the same as their model name changing the `.` by `_`.

An, no, there is no way to trace the table from the field and model name. Only environment could help.

pedrobaeza: > I am doing the same as https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L783. If that line is acceptable in the same method, mine should also be acceptable.

Such replacement is not for table names, but for certain identifiers, so please don't mix things.

I'm only telling you that if there's a way to get the real table name, we should use it.

pedrobaeza: @StefanRijnhart do you see this reasonable? Maybe we should have a table mapping for known standard changes (like act_window and company).

MiquelRForgeFlow: But here I am not doing any table renames, I don't understand what are you saying :S

StefanRijnhart: Ah yes, forget about that. I thought we were talking about the table of the model that was being renamed. In this case the check on table_exists is fine by me.

MiquelRForgeFlow: But the `ir.actions.*` models don't have any reference field. And I doubt any custom modules added one. And if a custom module added one, they are records with external id, thus they will be updated during load.

Thus, I don't think such mapping is needed when the model in question is not affected.

StefanRijnhart: I tend to agree in this case. Maybe at some point in the future we can create a convenience method in the lib to provide a model's table based on the model name plus a lookup table for that rare list of exceptions?

pedrobaeza: OK, let's go with this.

PR #176, opened 2019-09-19
travis: Disable byte compiling when installing on Python 2.7

em230418:
This should fix https://travis-ci.org/OCA/openupgradelib/builds/585505577

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ script:
     - pip install -q -r requirements.txt
     - export PYTHONPATH=~/build/OCA/openupgradelib:$PYTHONPATH
     - coverage run setup.py test
-    - python setup.py install
+    - python -B setup.py install
     # Run functional tests from 6.1 to 11.0
     - set -e
     - git clone https://github.com/oca/openupgrade --depth 1 --no-single-branch -b 6.1 ~/openupgrade


pedrobaeza:
Do you think that's the real problem (although it's of course appreciated)?

See https://travis-ci.org/OCA/openupgradelib/builds/585505577#L9967 for other weird errors.

Also disabling that byte compiling has any side effect?

em230418:
> Do you think that's the real problem (although it's of course appreciated)?

The real problem that python3 code (openupgrade_120.py) is being compiled by python2. We could rewrite file, to make sure it runs under python2, but since odoo 11 and odoo 12 are running on python3, I don't think we need it.

I tried to take away this file while running `python setup.py install` by making this patch (assuming running on python2):
```diff
diff --git a/setup.py b/setup.py
index 3e6894b..1eeebbc 100755
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,8 @@
 import os
 import openupgradelib
 
+from setuptools.command.build_py import build_py as build_py_orig
+
 try:
     from setuptools import setup
 except ImportError:
@@ -25,12 +27,20 @@ test_requirements = [
     'mock',
 ]
 
+class build_py(build_py_orig, object):
+
+    def find_package_modules(self, package, package_dir):
+        modules = super(build_py, self).find_package_modules(package, package_dir)
+        return [(pkg, mod, file,) for (pkg, mod, file, ) in modules
+                if mod != 'openupgrade_120']
+
 setup(
     name='openupgradelib',
     version=openupgradelib.__version__,
     description=openupgradelib.__doc__,
     long_description=readme + '\n\n' + history,
     author=openupgradelib.__author__,
+    cmdclass={'build_py': build_py},
     author_email=openupgradelib.__email__,
     url='https://github.com/OCA/openupgradelib',
     packages=['openupgradelib'],
```

But it didn't work.

> See https://travis-ci.org/OCA/openupgradelib/builds/585505577#L9967 for other weird errors.

No idea, how to fix it.

> Also disabling that byte compiling has any side effect?

openupgradelib will run a little-bit slower.

pedrobaeza:
@moylop260 @Yajo can you help with this?

moylop260:
You are using python 3.6 from travis, but activating python2.7 from script.

Who you need do it as it?

What about using a build for each python version like pylint-odoo?
 - https://github.com/OCA/pylint-odoo/blob/a94d53e024612b4069b2ef02f194e142ed2c8e77/.travis.yml#L9-L19

pedrobaeza:
@moylop260 we want to publish this in pypi compatible for both P2 and P3

moylop260:
> @pedrobaeza 
We want to publish this in pypi compatible for both P2 and P3

Yes, like pypi package of pylint-odoo
It is compatible for py2.7, py3.5 and py3.7

You can see the builds:
https://travis-ci.org/OCA/pylint-odoo/builds/572406052

pedrobaeza:
Can you show us where is the code for publishing to Pypi (or Travis config part)?

pedrobaeza:
Is it also possible to not compile that file when doing P2 publishing?

moylop260:
Did you mean this one?
 - https://github.com/OCA/pylint-odoo/blob/a94d53e024612b4069b2ef02f194e142ed2c8e77/.travis.yml#L50

Yajo:
This has nothing to do with pylint-odoo. AFAIK it's trying to fix [this faiulre](https://travis-ci.org/OCA/openupgradelib/builds/585505577#L372-L376):

```
byte-compiling build/bdist.linux-x86_64/egg/openupgradelib/openupgrade_120.py to openupgrade_120.pyc
  File "build/bdist.linux-x86_64/egg/openupgradelib/openupgrade_120.py", line 55
    *(_r("col%s-%s" % (t3, col), "col%s-%s" % (t4, col))
    ^
SyntaxError: invalid syntax
```

The point here is that the file `openupgrade_120.py` (and `openupgrade_110.py` or `openupgrade_130.py` if they existed) don't need to provide python2 compatibility because they are only meant to be executed in a py3 interpreter. Actually `openupgrade_100.py` and lower also don't need py3 compatibility for the opposite reason.

So there should be a way to instruct python that certain files must be ignored if the python version is lower than 3, and others when it's bigger than 3. And it should be done at package level because otherwise the bytecompiling process fails without need.

lmignon:
@Yajo A solution should be to create a last version for py2. Once released, we could remove the code that should never be run on py3 and release a new version. What do you think about?

Yajo:
:thinking: That'd be great, although I'm not sure that's what we want.

The problem with this specific lib is that, no matter how much you modernize your instances, you always have a chance of a new customer coming from a partner that had them in OpenERP 6.1 and with some addons that still have no migration scripts, and you need to add them, and you want to use the latest features/fixes this lib offers. Even when the rest of the world will get rid of Python 2 in January 1st 2020 (I wish), this lib will still need to support old codebases in Python 2 possibly forever, because it needs to be able to migrate old databases.

So IMHO the best solution would be to tell `pip` to exclude certain files depending on the interpreter version. TBH I don't know if that's possible... You are the pip experts.

If that's not possible, then that "last" release should exist in a separate branch, so we can still open PRs to it with fixes. I wish it to be the last, but... well, we all know it... it won't be. :broken_heart:  :man_shrugging:

PR #179, opened 2019-09-30
[FIX] Avoid deprecated openerp warning in >= 13.0

mreficent:
Cause: https://github.com/odoo/odoo/commit/7c47eb185443af08a55b0045c6a78f74151dc083.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -60,12 +60,12 @@ from . import openupgrade_tools
 
 core = None
 # The order matters here. We can import odoo in 9.0, but then we get odoo.py
-try:  # < 10.0
-    import openerp as core
-    from openerp.modules import registry
-except ImportError:  # >= 10.0
+try:  # >= 10.0
     import odoo as core
     from odoo.modules import registry
+except ImportError:  # < 10.0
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why changing the order? This is very fast and only happens once, so no need to change this.

mreficent:
Well, then in v13 and above you will several always warnings.

pedrobaeza:
Which warning?

mreficent:
just try in local v13 of openupgrade and you will see.

pedrobaeza:
Put here that warning, please. And why in previous versions is not failing?

mreficent:
```yml
WARNING ? odoo.modules.module: openerp is a deprecated alias to odoo. Please consider using the latter as the former is going to be removed in the next version. 
DeprecationWarning
```

pedrobaeza:
And switching this, we will get any other warning in previous versions?

mreficent:
No, I think not. Old versions, as I know, they didn't have a warning for calling odoo instead of openerp. If they are very old and doesn't get the odoo call, then the import error will work and will call the openerp.

pedrobaeza:
OK, let's go for it then.

PR #182, opened 2019-10-14
[ADD] switch_noupdate

mreficent:
To be able to switch noupdate easily in several records.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -176,6 +176,7 @@ __all__ = [
     'delete_record_translations',
     'disable_invalid_filters',
     'delete_records_safely_by_xml_id',
+    'set_xml_ids_noupdate_value',
 ]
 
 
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -2258,3 +2259,16 @@ def chunked(records, single=True):
...
+
+def switch_noupdate(env, xml_ids, value):
+    for xml_id in xml_ids:
+        if '.' not in xml_id:
+            logger.error(
+                'Need the module reference to be specified in the IDs' % xml_id)
+        else:
+            env.cr.execute("""
+                UPDATE ir_model_data
+                SET noupdate = NOT COALESCE(noupdate, FALSE)
```
pedrobaeza: Must be `value`

mreficent: Then the name of the method is wrong. "switch" means change the noupdate **always**. If you want a method that only puts the passed value, then the name "switch" is wrong, as sometimes you only want to assure a value but you really are not changing it 😏

openupgradelib/openupgrade.py
```diff
@@ -2258,3 +2259,16 @@ def chunked(records, single=True):
                 yield record
             continue
         yield chunk
+
+
+def set_noupdate_value(env, xml_ids, value):
```
pedrobaeza:
```suggestion
def set_xmlid_noupdate_value(env, xml_ids, value):
```

openupgradelib/openupgrade.py
```diff
@@ -2258,3 +2259,20 @@ def chunked(records, single=True):
...
+    """Set the xml_ids noupdate values in a module.
+
+    :param module: module name
+    :param xml_ids: a tuple or list of xml record IDs
+    :param bool value: True or False.
+    """
+    if not isinstance(xml_ids, (list, tuple)):
+        do_raise("XML IDs %s must be a tuple or list!" % xml_ids)
+
+    env.cr.execute("""
```
pedrobaeza: openupgrade.logged_query

PR #183, opened 2019-10-23
[FIX] savepoint must raise

Yajo:
This is how it works in modern Odoo versions.

@Tecnativa TT18838

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1902,6 +1902,7 @@ def savepoint(cr):
             cr.execute('RELEASE SAVEPOINT "%s"' % name)
         except Exception:
             cr.execute('ROLLBACK TO SAVEPOINT "%s"' % name)
+            raise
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is not used since ages, and I'm pretty sure it was on purpose to not re-raise... any possible insight about this, @hbrunn @StefanRijnhart ?

Yajo:
Without raising, there's no way to know if the process succeeded or not, as the exception is "eaten" by the context manager.

A few lines above, it reuses `cr.savepoint()` for modern odoo versions that include that method, and [Odoo implementation has been re-raising since v8, where it was introduced](https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/openerp/sql_db.py#L412).

hbrunn:
I don't remember details, but this very much looks like an oversight from my side

pedrobaeza:
OK then, let's merge it.

PR #195, opened 2019-12-19
[ADD] remove_table_fks: New method

Yajo:
Helper to delete all foreign keys in a table automatically. Recommended to be executed in tables that are not going to be used anymore, but still you don't want to delete them.

@Tecnativa TT18838

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -142,6 +142,7 @@ __all__ = [
     'add_fields',
     'copy_columns',
     'copy_fields_multilang',
+    'remove_tables_fks',
     'rename_columns',
     'rename_fields',
     'rename_tables',
@@ -497,6 +498,45 @@ def copy_fields_multilang(cr, destination_model, destination_table,
                 "src_m": source_model,
             },
         )
+
+
+def remove_tables_fks(cr, tables):
+    """Remove foreign keys declared in ``tables``.
+
+    This is useful when a table is not going to be used anymore, but you still
+    don't want to delete it.
+
+    If you keep FKs in that table, it will still get modifications when other
+    tables are modified too; but if you're keeping that table as a log, that
+    is a problem. Also, if some of the FK has no index, it could slow down
+    deletion in other tables, even when this one has no more use.
+
+    .. HINT::
+        This method removes FKs that are *declared* in ``tables``,
+        **not** FKs that *point* to those tables.
+
+    :param [str, ...] tables:
+        List of tables where the FKs were declared, and where they will be
+        removed too. If a table doesn't exist, it is skipped.
+    """
+    drop_sql = sql.SQL("ALTER TABLE {} DROP CONSTRAINT {}")
+    for table in tables:
+        cr.execute(
+            """
+                SELECT constraint_name
+                FROM information_schema.table_constraints
+                WHERE constraint_type = 'FOREIGN KEY' AND table_name = %s
+            """,
+            (table,),
+        )
+        for constraint in (row[0] for row in cr.fetchall()):
+            logged_query(
+                cr,
+                drop_sql.format(
+                    sql.Identifier(table),
+                    sql.Identifier(constraint),
+                ),
+            )
 
 
 def rename_columns(cr, column_spec):


Yajo:
Usage example: https://github.com/OCA/OpenUpgrade/pull/2136

MiquelRForgeFlow:
Let me ask: is it a problem to maintain the foreign keys of an old table in a database? if yes, what problem/s?

pedrobaeza:
Yes, it's a problem that we have discovered removing sale_order_line records. There was an ondelete="cascade" in procurement_order table that provokes a slow down of the migration with no benefit.

Yajo:
@pedrobaeza AFAIK the query is good and what we *do* want is the FKs in the table that was obsoleted, not the ones pointing to it.

FKs pointing to it will be removed by normal odoo operation when the field disappears.

Check this:

```sql
SELECT constraint_name
            FROM information_schema.table_constraints
            WHERE constraint_type = 'FOREIGN KEY' AND table_name='procurement_order';
             constraint_name             
-----------------------------------------
 procurement_order_bom_id_fkey
 procurement_order_company_id_fkey
 procurement_order_create_uid_fkey
 procurement_order_group_id_fkey
 procurement_order_location_id_fkey
 procurement_order_move_dest_id_fkey
 procurement_order_move_id_fkey
 procurement_order_orderpoint_id_fkey
 procurement_order_partner_dest_id_fkey
 procurement_order_product_id_fkey
 procurement_order_product_uom_fkey
 procurement_order_product_uos_fkey
 procurement_order_production_id_fkey
 procurement_order_purchase_id_fkey
 procurement_order_purchase_line_id_fkey
 procurement_order_requisition_id_fkey
 procurement_order_rule_id_fkey
 procurement_order_warehouse_id_fkey
 procurement_order_write_uid_fkey
(19 rows)
```

Yajo:
Modified to accept multiple tables.

pedrobaeza:
Let's discuss it this by chat, as I don't see then the need of this.

pedrobaeza:
Seen directly and this is indeed the less bad of the options and there's only a remote possibility of having inconsistencies if reusing tables, and in that case, a script on the module that reuses the table is needed.

Please check any ways Travis.

StefanRijnhart:
For my taste, the method name as well as the docstring do not clearly disambiguate between FK constraints that refer to the given tables and FK constraints on values in the columns of the given tables. But the question is, maybe we should remove both given that the usage of this method indicates complete and utter deprecation of the data?

pedrobaeza:
We have checked in live that removing the FKs in other tables doesn't get any benefit, as that column remains untouched and in new records with NULL values FKs are not checked. Do you have something more in mind?

StefanRijnhart:
No, that's fine. I'd still like the method to be renamed to something like `remove_fk_constraints_from_tables` in that case.

Yajo:
> `remove_fk_constraints_from_tables`

That's grammatically equivalent to `remove_tables_fks`, so it's equally (un)specific.

Let me add that to the docstring.

Yajo:
Changed.

pedrobaeza:
Travis is failing

Yajo:
Thanks, fixed.

pedrobaeza:
Still red

Yajo:
We have some misconfiguration here and I cannot enter Sphinx docs. Well, nevermind. I removed the `:param:\`tables\`` call.

pedrobaeza:
I don't get the problem, but this continues red.

openupgradelib/openupgrade.py
```diff
@@ -499,6 +500,44 @@ def copy_fields_multilang(cr, destination_model, destination_table,
...
+    If you keep FKs in that table, it will still get modifications when other
+    tables are modified too; but if you're keeping that table as a log, that
+    is a problem. Also, if some of the FK has no index, it could slow down
+    deletion in other tables, even when this one has no more use.
+
+    .. HINT::
+        This method removes FKs that are *declared* in ``tables``,
+        **not** FKs that *point* to those tables.
+
+    :param *str tables:
```
StefanRijnhart: This star is taken as emphasis start causing /home/travis/build/OCA/openupgradelib/openupgradelib/openupgrade.py:docstring of openupgradelib.openupgrade.remove_tables_fks:15:Inline emphasis start-string without end-string.

openupgradelib/openupgrade.py
```diff
@@ -499,6 +500,44 @@ def copy_fields_multilang(cr, destination_model, destination_table,
         )
 
 
+def remove_tables_fks(cr, *tables):
```
pedrobaeza: I don't get why putting the argument with `*` and not simply pass an iterable

openupgradelib/openupgrade.py
Yajo: Fixed, thanks.

openupgradelib/openupgrade.py
Yajo: In case you want to just drop a table, you don't need to create a list.

Well, nevermind, I'll change it as I see that passing lists around is the common usage in this lib.

openupgradelib/openupgrade.py
```diff
@@ -499,6 +500,45 @@ def copy_fields_multilang(cr, destination_model, destination_table,
...
+    If you keep FKs in that table, it will still get modifications when other
+    tables are modified too; but if you're keeping that table as a log, that
+    is a problem. Also, if some of the FK has no index, it could slow down
+    deletion in other tables, even when this one has no more use.
+
+    .. HINT::
+        This method removes FKs that are *declared* in :param:`tables`,
+        **not** FKs that *point* to those tables.
+
+    :param [str, ...] tables:
```
pedrobaeza:
```
/home/travis/build/OCA/openupgradelib/openupgradelib/openupgrade.py:docstring of openupgradelib.openupgrade.remove_tables_fks:1:Unknown interpreted text role "param"
```

Yajo: Oh it seems [I cannot link to params](https://stackoverflow.com/a/11178709/1468388)... Fixed now.

PR #197, opened 2020-02-13
[ADD] new tool method to move data to a company dependent field

ernestotejeda:
This PR adds a new tool method to move data from a given field to a given company_dependet field, both from the same table. Useful in cases when from one version to another one, some field becomes a 'company dependent' field.

Cc @Tecnativa TT21084

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -179,6 +179,7 @@ __all__ = [
     'disable_invalid_filters',
     'delete_records_safely_by_xml_id',
     'set_xml_ids_noupdate_value',
+    'convert_to_company_dependent',
 ]
 
 
@@ -2367,3 +2368,92 @@ def set_xml_ids_noupdate_value(env, module, xml_ids, value):
         SET noupdate = %s
         WHERE module = %s AND name in %s
     """, (value, module, tuple(xml_ids),))
+
+
+def convert_to_company_dependent(
+    env,
+    model_name,
+    origin_field_name,
+    destination_field_name,
+    model_table_name=None,
+):
+    """ For each row in a given table, the value of a given field is
+    set in another 'company dependant' field of the same table.
+    Useful in cases when from one version to another one, some field in a
+    model becomes a 'company dependent' field.
+
+    This method must be executed in post-migration scripts after
+    the field is created, or in pre-migration if you have previously
+    executed add_fields openupgradelib method.
+
+    :param model_name: Name of the model.
+    :param origin_field_name: Name of the field from which the values
+      will be obtained.
+    :param destination_field_name: Name of the 'company dependent'
+      field where the values obtained from origin_field_name will be set.
+    :param model_table_name: Name of the table. Optional. If not provided
+      the table name is taken from the model (so the model must be
+      registered previously).
+    """
+    logger.debug("Converting {} in {} to company_dependent field {}.".format(
+        origin_field_name, model_name, destination_field_name))
+    if origin_field_name == destination_field_name:
+        do_raise("A field can't be converted to property without changing "
+                 "its name.")
+    cr = env.cr
+    mapping_type2field = {
+        'char': 'value_text',
+        'float': 'value_float',
+        'boolean': 'value_integer',
+        'integer': 'value_integer',
+        'text': 'value_text',
+        'binary': 'value_binary',
+        'many2one': 'value_reference',
+        'date': 'value_datetime',
+        'datetime': 'value_datetime',
+        'selection': 'value_text',
+    }
+    # Determine field id, field type and the model name of the relation
+    # in case of many2one.
+    cr.execute("SELECT id, relation, ttype FROM ir_model_fields "
+               "WHERE name=%s AND model=%s",
+               (destination_field_name, model_name))
+    destination_field_id, relation, d_field_type = cr.fetchone()
+    value_field_name = mapping_type2field.get(d_field_type)
+    field_select = sql.Identifier(origin_field_name)
+    args = {
+        'model_name': model_name,
+        'fields_id': destination_field_id,
+        'name': destination_field_name,
+        'type': d_field_type,
+    }
+    if d_field_type == 'many2one':
+        field_select = sql.SQL("%(relation)s || ',' || {}::TEXT").format(
+            sql.Identifier(origin_field_name))
+        args['relation'] = relation
+    elif d_field_type == 'boolean':
+        field_select = sql.SQL("CASE WHEN {} = true THEN 1 ELSE 0 END").format(
+            sql.Identifier(origin_field_name))
+    cr.execute("SELECT id FROM res_company")
+    company_ids = [x[0] for x in cr.fetchall()]
+    for company_id in company_ids:
+        args['company_id'] = company_id
+        logged_query(
+            cr, sql.SQL("""
+            INSERT INTO ir_property (
+                fields_id, company_id, res_id, name, type, {value_field_name}
+            )
+            SELECT
+                %(fields_id)s, %(company_id)s,
+                %(model_name)s || ',' || id::TEXT, %(name)s,
+                %(type)s, {field_select}
+            FROM {table_name} WHERE {origin_field_name} IS NOT NULL;
+            """).format(
+                value_field_name=sql.Identifier(value_field_name),
+                field_select=field_select,
+                origin_field_name=sql.Identifier(origin_field_name),
+                table_name=sql.Identifier(
+                    model_table_name or env[model_name]._table
+                )
+            ), args,
+        )


openupgradelib/openupgrade.py
```diff
@@ -2367,3 +2368,88 @@ def set_xml_ids_noupdate_value(env, module, xml_ids, value):
         SET noupdate = %s
         WHERE module = %s AND name in %s
     """, (value, module, tuple(xml_ids),))
+
+def move_to_company_dependent(
```
pedrobaeza: Better to call it `convert_to_company_dependent`

openupgradelib/openupgrade.py
```diff
@@ -2367,3 +2368,88 @@ def set_xml_ids_noupdate_value(env, module, xml_ids, value):
...
+        origin_field_name,
+        destination_field_name,
+    ):
+        """ For each row in a given table, the value of a given field is
+        set in another 'company dependant' field of the same table.
+        Useful in cases when from one version to another one, some field in a
+        model becomes a 'company dependent' field.
+
+        :param model_name: Name of the model.
+        :param model_table_name: Name of the table.
```
pedrobaeza: Put this argument as optional, and if not provided, you get it from `env[model_name]._table`.

openupgradelib/openupgrade.py
```diff
@@ -2367,3 +2368,88 @@ def set_xml_ids_noupdate_value(env, module, xml_ids, value):
...
+        :param origin_field_name: Name of the field from which the values
+                                  will be obtained.
+        :param company_dependent_field_name: Name of the 'company dependent'
+                                             field where the values obtained
+                                             from origin_field_name will be
+                                             set.
+        """
+        logger.debug(
+            "Moving data to a company_dependent field:\n"
```
pedrobaeza: Better to put it in only one line for not having very long logs:
```suggestion
            "Converting {} in {} to company_dependent field {}.".format(origin_field_name, model_name, destination_field_name)
```

openupgradelib/openupgrade.py
```diff
@@ -2367,3 +2368,88 @@ def set_xml_ids_noupdate_value(env, module, xml_ids, value):
...
+
+        :param model_name: Name of the model.
+        :param model_table_name: Name of the table.
+        :param origin_field_name: Name of the field from which the values
+                                  will be obtained.
+        :param company_dependent_field_name: Name of the 'company dependent'
+                                             field where the values obtained
+                                             from origin_field_name will be
+                                             set.
+        """
```
pedrobaeza: You must supply another optional argument for the field type. If not provided, then you should introspect the field type, but better to do it with ORM in this case.

openupgradelib/openupgrade.py
```diff
@@ -2367,3 +2368,88 @@ def set_xml_ids_noupdate_value(env, module, xml_ids, value):
...
+            'text': 'value_text',
+            'binary': 'value_binary',
+            'many2one': 'value_reference',
+            'date': 'value_datetime',
+            'datetime': 'value_datetime',
+            'selection': 'value_text',
+        }
+        # Determine field id and the model name of the relation
+        # in case of many2one.
+        cr.execute("""
```
pedrobaeza: Instead of this (and if not provided as argument), do `env[origin_field_name]._fields[model_name].type`

openupgradelib/openupgrade.py
```diff
@@ -2367,3 +2368,88 @@ def set_xml_ids_noupdate_value(env, module, xml_ids, value):
...
+        # Iterate values to insert into ir_property table
+        for model_id, value in values:
+            field_type = env[model_name]._fields[destination_field_name].type
+            # Determine extra field value: see mapping_type2field.
+            extra_field_value = value
+            extra_field = mapping_type2field.get(field_type)
+            if extra_field == 'value_reference':
+                extra_field_value = '%s,%d' % (relation, value)
+            # Insert into ir_property table using a sql query
+            logged_query(
```
pedrobaeza: You must do everything in a SQL sentence. Imagine if you have 2 million records to find that 2 million lines of log.

You can perform `INSERT INTO ... SELECT ...` in the same SQL sentence for inserting all rows at the same time:

https://www.w3schools.com/sql/sql_insert_into_select.asp

but you have to deal with all the conditional in SQL. You have `CASE` for that.

openupgradelib/openupgrade.py
```diff
@@ -2367,3 +2368,88 @@ def set_xml_ids_noupdate_value(env, module, xml_ids, value):
...
+                    extra_field=extra_field,
+                    fields_id=destination_field_id,
+                    company_id=env.user.company_id.id,
+                    res_id='{},{}'.format(model_name, model_id),
+                    name=destination_field_name,
+                    type=field_type,
+                    extra_field_value=extra_field_value,
+                )
+            )
+        logger.debug("Data moved successfully.")
```
pedrobaeza: No need to this log.

ernestotejeda:
Changes done.

But, I haven't set the optional argument for the Field type because anyway in these lines
https://github.com/OCA/openupgradelib/pull/197/files#diff-3d7675eb1386de028dab1f879bdc4681R2411-R2414
I have a select query to `ir_model_fields` table to obtain the ID of the destination field and the model to which it is related (in case be a many2one), so I can include in the select the `ttype` column to also obtain the Field type. What do you think? Do i need it anyway?

pedrobaeza:
OK, if the field ID is needed, then please mention in the docstring that this method must be executed in post-migration scripts after the field is created, or in pre-migration if you have previously executed `add_fields` openupgradelib method.

ernestotejeda:
Done

openupgradelib/openupgrade.py
```diff
@@ -2367,3 +2368,80 @@ def set_xml_ids_noupdate_value(env, module, xml_ids, value):
...
+        'selection': 'value_text',
+    }
+    # Determine field id, field type and the model name of the relation
+    # in case of many2one.
+    cr.execute(("SELECT id, relation, ttype FROM ir_model_fields "
+                "WHERE name='{}' AND model='{}'").format(
+                    destination_field_name, model_name))
+    destination_field_id, relation, d_field_type = cr.fetchone()
+    value_field_name = mapping_type2field.get(d_field_type)
+    field_select = origin_field_name
```
pedrobaeza: You have here the same injection problem, so this can be rewrite using a helper in psycopg2:
```python
from psycopg2 import sql
```
that allows to create fragments of SQL correctly prepared. Left parts of the SQL will be handled by the format method in SQL with `{}`, and right parts will continue being parsed by final `execute` passed as arguments in the query and using `%s` for marking them.

You have to also take into account all the possible existing companies in the DB.

I'm going to put whole code from this line as suggestion for not putting isolated comments:
```suggestion
    field_select = sql.Identifier(origin_field_name)
    args = {
        'model_name': model_name,
        'fields_id': destination_field_id,
        'name': destination_field_name,
        'type': d_field_type,
    }
    if d_field_type == 'many2one':
        field_select = sql.SQL("%(relation)s || ',' || {}::TEXT").format(
            sql.Identifier(origin_field_name))
        args['relation'] = relation
    elif d_field_type == 'boolean':
        field_select = sql.SQL("CASE WHEN {} = true THEN 1 ELSE 0 END").format(
            sql.Identifier(origin_field_name))
    cr.execute("SELECT id FROM res_company")
    company_ids = [x[0] for x in cr.fetchall()]
    for company_id in company_ids:
        args['company_id'] = company_id
        logged_query(
            cr, sql.SQL("""
            INSERT INTO ir_property (
                fields_id, company_id, res_id, name, type, {value_field_name}
            )
            SELECT
                %(fields_id)s, %(company_id)s,
                %(model_name)s || ',' || id::TEXT, %(name)s,
                %(type)s, {field_select}
            FROM {table_name} WHERE {origin_field_name} IS NOT NULL;
            """).format(
                value_field_name=sql.Identifier(value_field_name),
                field_select=field_select,
                origin_field_name=sql.Identifier(origin_field_name),
                table_name=sql.Identifier(
                    model_table_name or env[model_name]._table
            )), args,
        )
```

openupgradelib/openupgrade.py
```diff
@@ -2367,3 +2368,80 @@ def set_xml_ids_noupdate_value(env, module, xml_ids, value):
...
+    :param company_dependent_field_name: Name of the 'company dependent'
+                                         field where the values obtained
+                                         from origin_field_name will be
+                                         set.
+    :param model_table_name: Name of the table.
+    """
+    logger.debug("Converting {} in {} to company_dependent field {}.".format(
+        origin_field_name, model_name, destination_field_name))
+    cr = env.cr
+    mapping_type2field = {
```
pedrobaeza: Check if the source and destination field are different. If not, raise an error saying that a field can't be converted to property without changing its name.

ernestotejeda:
Right, thanks @pedrobaeza   , lesson learned. I also took a look at psycopg2 documentation to see those things. I also agree with taking into account all the companies.

PR #198, opened 2020-02-29
[FIX] merge_records: Handle duplicates in mail.followers

pedrobaeza:
If you have a model inheriting from mail.thread with followers, there's a chance that merged records contain the same followers, leading to an unique constraint error.

This commit handles this, detecting these duplicates before it happens, and removing extra remaining records.

@Tecnativa TT22543

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -5,6 +5,7 @@
 
 import logging
 import functools
+from psycopg2 import sql
 from psycopg2 import ProgrammingError, IntegrityError
 from psycopg2.errorcodes import UNIQUE_VIOLATION
 from psycopg2.extensions import AsIs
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_merge_records.py
```diff
@@ -418,26 +419,52 @@ def _change_generic(env, model_name, record_ids, target_record_id,
...
+                else:
+                    # We need to avoid duplicated results in this model
+                    target_duplicated = model.search([
+                        (model_column, '=', model_name),
+                        (res_id_column, '=', target_record_id),
+                        ('partner_id', 'in', records.mapped('partner_id').ids),
+                    ])
+                    duplicated = records.filtered(lambda x: (
+                        x.partner_id in target_duplicated.mapped('partner_id'))
+                    )
```
Yajo: :hand: Beware bottleneck ahead! Change to:
```suggestion
                    target_duplicated_partners = target_duplicated.mapped('partner_id')
                    duplicated = records.filtered(lambda x: (
                        x.partner_id in target_duplicated_pids)
                    )
```

pedrobaeza: I think with previous code it gets prefetched and no penalty happens, but no problem in changing it.

pedrobaeza: Changed

Yajo: It's not the prefetched, it's because you're ussing mapped instead of filtered, so that's exponential complexity.

PR #199, opened 2020-03-12
[FIX] let add_fields() work for Odoo <= v8

Yajo:
In those versions, [there is a required column called `select_level`][1] whose [default value is `"0"`][2].

I'm adding it to `add_fields()` so it doesn't crash when trying to use it in those versions.

The crash looks like this:

```
2020-03-12 03:48:38,493 1 ERROR prod openerp.sql_db: bad query:
            INSERT INTO ir_model_fields (
                model_id, model, name, field_description, ttype, state
            ) VALUES (
                117, 'product.template', 'pack', 'OU', 'bool', 'base'
            ) RETURNING id
  File "/opt/odoo/custom/src/odoo/openerp/sql_db.py", line 247, in execute
IntegrityError: null value in column "select_level" violates not-null constraint
DETAIL:  Failing row contains (9885, product.template, 117, pack, null, null, OU, bool, base, null, f, null, f, null, null, null, null, null, null, null, null, null, null, null, null, null).

2020-03-12 03:48:38,493 1 ERROR prod OpenUpgrade: Error after 0:00:00.000372 running
            INSERT INTO ir_model_fields (
                model_id, model, name, field_description, ttype, state
            ) VALUES (
                117, 'product.template', 'pack', 'OU', 'bool', 'base'
            ) RETURNING id
2020-03-12 03:48:38,493 1 ERROR prod OpenUpgrade: product_pack: error in migration script product_pack/migrations/8.0.1.0.0/pre-migration.py: IntegrityError('null value in column "select_level" violates not-null constraint\nDETAIL:  Failing row contains (9885, product.template, 117, pack, null, null, OU, bool, base, null, f, null, f, null, null, null, null, null, null, null, null, null, null, null, null, null).\n',)
2020-03-12 03:48:38,493 1 ERROR prod OpenUpgrade: null value in column "select_level" violates not-null constraint
```

[1]: https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/openerp/addons/base/ir/ir_model.py#L261
[2]: https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/openerp/addons/base/ir/ir_model.py#L286

@Tecnativa TT22023 TT18838

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2125,7 +2125,7 @@ def add_fields(env, field_spec):
     It's intended for being run in pre-migration scripts for pre-populating
     fields that are going to be declared later in the module.
 
-    NOTE: This only works in >=v9 and is not needed in >=v12, as now Odoo
+    NOTE: This is not needed in >=v12, as now Odoo
     always add the XML-ID entry:
     https://github.com/odoo/odoo/blob/9201f92a4f29a53a014b462469f27b32dca8fc5a/
     odoo/addons/base/models/ir_model.py#L794-L802, but you can still call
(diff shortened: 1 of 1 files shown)


Yajo:
But if the column exists as a remnant and it is still required, wouldn't it make sense to check its existence? It would crash anyway without this part.

pedrobaeza:
According latest moves we have done in OpenUpgrade, that constraint should be removed when the column is not in use anymore. I haven't seen any way any problem derived from that column, so it should be done already.

Yajo:
Done.

pedrobaeza:
Sorry for the nitpicking, but you have converted a 5 line log to a very longer log. Please keep the number of lines and also extra empty lines. Take into account that this is logged.

Yajo:
You shouldn't care so much about those meaningless details. :roll_eyes:  Done, anyway.

pedrobaeza:
Yes, I care because I'm the one reading that logs after each migration, and that upsets me, specially in a so low level function.

PR #202, opened 2020-03-22
[FIX] rename references in ir.filters when renaming models

boskowski:
Rename references in user-defined filters (ir.filters) when renaming model.


boskowski:
It's a Selection, where the key is the model name in dot notation. I've no clue why travis fails.

boskowski:
My fault, the UPDATE was inserted at a wrong position. https://github.com/OCA/openupgradelib/commit/91d21a2d035c8c8e80209d948e3088d277274fe7 should fix this.

PR #204, opened 2020-04-18
merge_records: Allow to run it by SQL on base pre-migration

pedrobaeza:
Removing env.cr.savepoint that requires that base model is loaded in registry

@Tecnativa


kos94ok-3D:
Please add patch khaeusler/openupgradelib@ee6b948 writed by @khaeusler

pedrobaeza:
Added with some minor tweak, but indeed it's the correct way

PR #206, opened 2020-05-01
[ADD] new progress decorator based on progressbar module

ypapouin:
This is a first attempt to create a progress decorator.
It was not tested with OpenERP/Odoo Cursor class, I made my tests with the psycopg2._ext.cursor class.

It was discussed here:
https://github.com/OCA/OpenUpgrade/issues/1897

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -24,6 +24,7 @@ import os
 import inspect
 import uuid
 import logging as _logging_module
+import progressbar
 from datetime import datetime
 try:
     from StringIO import StringIO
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Not sure about linking the bar with a row fetch, because progress bar will be associated usually with ORM operations, not with DB fetches. I think a decorator that links the call to a method and a total number would be better, so each time that method is called, you increase the counter.

legalsylvain:
I think that both (orm operation & DB fetches) are valid and should be covered. (Maybe with two decorators, if a single one is not simple to implement.)

pedrobaeza:
Supersede by #233

PR #207, opened 2020-06-17
[FIX] rename_fields: Don't rename only by name + reverse order

pedrobaeza:
Doing that, you can rename exports from other models that match the field name.

I don't know how can this be so many time there without being aware of the problem...

TT24303

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -618,10 +618,15 @@ def rename_fields(env, field_spec, no_deep=False):
         # Rename appearances on export profiles
         # TODO: Rename when the field is part of a submodel (ex. m2one.field)
         cr.execute("""
-            UPDATE ir_exports_line
+            UPDATE ir_exports_line iel
             SET name = %s
-            WHERE name = %s
-            """, (old_field, new_field),
+            FROM ir_exports ie,
+                ir_model im
+            WHERE iel.name = %s
+                AND ie.id = iel.export_id
+                AND im.id = ie.model_id
+                AND im.model = %s
+            """, (new_field, old_field, model),
         )
         # Rename appearances on filters
         # Example of replaced domain: [['field', '=', self], ...]
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -611,8 +611,8 @@ def rename_fields(env, field_spec, no_deep=False):
             WHERE name = %s
                 AND type = 'model'
             """, (
-                "%s,%s" % (model, old_field),
                 "%s,%s" % (model, new_field),
+                "%s,%s" % (model, old_field),
             ),
         )
         # Rename appearances on export profiles


openupgradelib/openupgrade.py
```diff
@@ -618,10 +618,15 @@ def rename_fields(env, field_spec, no_deep=False):
...
             SET name = %s
-            WHERE name = %s
-            """, (old_field, new_field),
+            FROM ir_exports ie,
+                ir_model im
+            WHERE iel.name = %s
+                AND ie.id = iel.export_id
+                AND im.id = ie.model_id
+                AND im.model = %s
+            """, (old_field, new_field, model),
```
MiquelRForgeFlow, line 629:
```suggestion
            """, (new_field, old_field, model),
```
🤔

MiquelRForgeFlow: the same for ir_translation query just above in the code

pedrobaeza: It has been this way all the time? :scream:

pedrobaeza:
Fixed

MiquelRForgeFlow:
some error in travis

pedrobaeza:
Travis is log is cut and I don't see any problem...

pedrobaeza:
Let's try to merge it...

/ocabot merge nobump

MiquelRForgeFlow:
hahaha, but then the travis will be red in main branch :S

pedrobaeza:
That's not the idea. I'm merging through bot, so it won't merge if not green. I'm trying if it was a temporary glitch.

pedrobaeza:
No way... @Yajo maybe you can see what is the problem. The log is cut suddenly and no error that provokes the CI failure is seen.

pedrobaeza:
The problem is on main branch as well, but I have been able to see the log:

https://api.travis-ci.org/v3/job/680075547/log.txt

so I have removed coveralls reference.

/ocabot merge nobump

MiquelRForgeFlow:
Needs rebase

pedrobaeza:
Still failing, but Travis is not indicating the problem. I'm going to merge manually as it seems Travis log is seen full on main branch. I will fix it later when knowing the problem.

StefanRijnhart:
Can we revert this? Running into 
```
2020-06-17 21:12:52,274 3188 INFO migtest OpenUpgrade: table res_partner_bank, column bank: renaming to bank_id
2020-06-17 21:12:52,280 3188 INFO migtest openerp.sql_db: Programming error: column ie.model_id does not exist
LINE 8:                 AND im.id = ie.model_id
                                    ^
, in query 
            UPDATE ir_exports_line iel
            SET name = %s
            FROM ir_exports ie,
                ir_model im
            WHERE iel.name = %s
                AND ie.id = iel.export_id
                AND im.id = ie.model_id
                AND im.model = %s
            
```
On Odoo 9.0

StefanRijnhart:
In fact, the field is not from Odoo in any edition but from OCA module base_export_manager.

PR #208, opened 2020-06-18
Revert "[FIX] rename_fields: Don't rename only by name + reverse order"

StefanRijnhart:
Reverts OCA/openupgradelib#207

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -611,22 +611,17 @@ def rename_fields(env, field_spec, no_deep=False):
             WHERE name = %s
                 AND type = 'model'
             """, (
-                "%s,%s" % (model, new_field),
                 "%s,%s" % (model, old_field),
+                "%s,%s" % (model, new_field),
             ),
         )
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
Fasttracking this as it breaks everything.

pedrobaeza:
Why do you revert so quick? The code is correct and it's fixing a problem. I'm investigating the problem in Travis.

StefanRijnhart:
See my comment on the original PR. ie.model_id is not from core Odoo but from base_export_manager. You need to file a new PR that checks if the column exists.

pedrobaeza:
OK, now I see the problem. Why Travis is not offering this information?

StefanRijnhart:
Because nobody writes tests for Openupgradelib even if it is supported? This would have come out today because PRs on Openupgrade would have failed using this version. But because my nightly migration failed today I found out early.

pedrobaeza:
Well, we know the problem here with tests, but the question is not that one. Travis was erroring, but the log was cut, so no clue on what's going on.

PR #209, opened 2020-06-18
[FIX] rename_fields: reverse order + consider base_export_manager

MiquelRForgeFlow:


pedrobaeza:
The no existing `base_export_manager` version is not valid. Isn't any field that we can populate for putting the proper model?

MiquelRForgeFlow:
ready

openupgradelib/openupgrade.py
```diff
@@ -611,18 +611,34 @@ def rename_fields(env, field_spec, no_deep=False):
...
+                FROM ir_exports ie,
+                    ir_model im
+                WHERE iel.name = %s
+                    AND ie.id = iel.export_id
+                    AND im.id = ie.model_id
+                    AND im.model = %s
+                """, (new_field, old_field, model),
+            )
+        elif column_exists(cr, "ir_exports", "resource"):
+            cr.execute("""
```
pedrobaeza, line 633: Why not using always this SQL then?

MiquelRForgeFlow: As I see, base_export_manager makes required=False the resource, and let's people choose between using model or resource.

I think it's ok now.

pedrobaeza:
Well, that's a bad design then, as `resource` must be always filled. Please get rid off that code and let the standard one, and let's fix it in `base_export_manager` (through a sync or similar). Don't remove the TODO as well, as it's not yet fulfill

MiquelRForgeFlow:
done

PR #211, opened 2020-07-15
[FIX] Image floating snippet replacements

chienandalu:
- A class was added to `s_image_floating` classes but no styles on that class were added by the core.
- The `float-left` and `float-right` classes added the opossite expected margin class.
- A p-0 is added to the image as well so we have the block as the original one.

cc @Tecnativa TT24615


chienandalu:
please @Tardo @Yajo take a look

openupgradelib/openupgrade_120.py
```diff
@@ -225,25 +225,25 @@
...
     _r(selector=".o_image_floating.o_margin_m", class_add="mb8"),
     _r(selector=".o_image_floating.o_margin_l",
        style_add={"margin-bottom": "12px"}),
     _r(selector=".o_image_floating.o_margin_xl", class_add="mb24"),
     _r(selector=".o_image_floating .o_footer", class_rm="o_footer"),
-    _r("s_image_floating", "col-5"),
+    _r(class_rm="s_image_floating"),
     _r("o_image_floating o_margin_s o_margin_m o_margin_l o_margin_xl",
-       "col-5", selector=".o_image_floating"),
+       "col-5 p-0", selector=".o_image_floating"),
```
Tardo, line 246: This is the expected behaviour?  Shouldn't be removed as with 's_image_floating'?

chienandalu: No, this one has related styles that have to be replaced: https://github.com/odoo/odoo/blob/11.0/addons/website/static/src/less/website.snippets.less#L88-L90

PR #213, opened 2020-08-20
[FIX] rename_fields: Rename attachments for binary field

pedrobaeza:
If the field is Binary with attachment=True, the attachments are not renamed, and the content was lost.

Here we apply it unconditionally for every rename, as if the field is not binary or without attachments, the result of this will be 0 rows changed.

@Tecnativa TT23524

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -615,6 +615,14 @@ def rename_fields(env, field_spec, no_deep=False):
                 "%s,%s" % (model, old_field),
             ),
         )
+        # Rename possible attachments (if field is Binary with attachment=True)
+        cr.execute("""
+            UPDATE ir_attachment
+            SET res_field = %s
+            WHERE res_model = %s
+                AND res_field = %s
+            """, (new_field, model, old_field)
+        )
         # Rename appearances on export profiles
         # TODO: Rename when the field is part of a submodel (ex. m2one.field)
         cr.execute("""


Yajo (approved):

StefanRijnhart (approved):

StefanRijnhart:
Does ocabot work on repos like this?

/ocabot merge minor

pedrobaeza:
The version bump is not going to happen, but the merge can be done. I'm not sure if this is going to be merged though, as there was an error in Travis and it's a weird one.

StefanRijnhart:
It just hangs there doesn't it, somewhere in update_module_names.

pedrobaeza:
Yeah, it's very weird.

We were planning to redo the CI at my colleague's return from holidays to use GitHub new triggers (removing Travis here) and make independent each version build. We have been already testing new GitHub CI tools in for example https://github.com/Tecnativa/doodba-copier-template, and it works very good.

pedrobaeza:
Merging this manually as well while we fix CI, as it adds an important fix.

StefanRijnhart:
That's fine. Have you managed to get more insight into the Travis troubles?

pedrobaeza:
Unfortunately no. The roadmap we will work next week is to switch to GitHub CI.

StefanRijnhart:
At least that's something to look forward to!

pedrobaeza:
@StefanRijnhart it turns out that the problem for having broken Travis is https://github.com/OCA/openupgradelib/commit/55531aa4c9580342f3a6def0826175a8da841294, so now we know that although the error is not shown in logs, there is something.

StefanRijnhart:
Excellent, thanks for the update!

PR #214, opened 2020-09-16
[FIX] _delete_records_sql was deleting wrong ids

MiquelRForgeFlow:
Strange we didn't see this before.


pedrobaeza:
Because AFAIK it's not used in OpenUpgrade (it uses ORM operation). Well, maybe it was introduced when merging the base user in v12. Anyway, indeed is a great failure. Merging manually as CI is broken, but very important to fix this ASAP.

legalsylvain (commented):
thanks !

PR #215, opened 2020-09-24
[FIX] Decorated functions should keep name and docstring.

NL66278:
There are errors with the migrate decorator in Odoo 13.0 and possibly other releases, because the migration
scripts do not have a migrate attribute. This commit recifies that situation.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1,6 +1,8 @@
-# Copyright (C) 2011-2013 Therp BV (<http://therp.nl>)
+# -*- coding: utf-8 -*-
+# Copyright 2011-2020 Therp BV <https://therp.nl>.
+# Copyright 2016-2020 Tecnativa - Pedro M. Baeza.
 # Copyright Odoo Community Association (OCA)
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 import sys
 import os
@@ -8,6 +10,7 @@ import inspect
 import uuid
 import logging as _logging_module
 from datetime import datetime
+from functools import wraps
 try:
     from StringIO import StringIO
 except ImportError:
@@ -1772,10 +1775,11 @@ def logging(args_details=False, step=False):
 
     """
     def wrap(func):
+
+        @wraps(func)
         def wrapped_function(*args, **kwargs):
             to_log = True
             msg = "Executing method %s" % func.__name__
-
             # Count calls
             if step:
                 # Compute unique name
@@ -1790,18 +1794,16 @@ def logging(args_details=False, step=False):
                         msg += " ; Logging Step : %d" % step
                 else:
                     to_log = False
-
             # Log Args
             if args_details and to_log:
                 if args:
                     msg += " ; args : %s" % str(args)
                 if kwargs:
                     msg += " ; kwargs : %s" % str(kwargs)
-
             if to_log:
                 logger.info(msg)
-
             return func(*args, **kwargs)
+
         return wrapped_function
     return wrap
 
@@ -1832,6 +1834,8 @@ def migrate(no_version=False, use_env=None, uid=None, context=None):
     logging purposes.
     """
     def wrap(func):
+
+        @wraps(func)
         def wrapped_function(cr, version):
             stage = 'unknown'
             module = 'unknown'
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -1778,29 +1760,35 @@ def wrapped_function(cr, version):
                     # Python3: fetch pyfile from locals, not fp
                     filename = frame.locals.get(
                         'pyfile') or frame.locals['fp'].name
-                except Exception as e:
+                except Exception as exc:
```
pedrobaeza, line 1763: Please don't rename this variable for not increasing diff

NL66278: pylint is loadly complaining about one-letter variable-name, as well as about using `%` formatter to create logging messages.

NL66278: I try not to create to big a diff, therefor only changed the functions that had to change anyway.

pedrobaeza: But that complaint is only a warning AFAIK.

openupgradelib/openupgrade.py
```diff
@@ -1,26 +1,10 @@
...
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2011-2020 Therp BV <https://therp.nl>.
```
pedrobaeza, line 2: This copyright is incomplete, although I understand the abbreviation. Please add at least:

```
Copyrigh 2016-2020 Tecnativa - Pedro M. Baeza
```

I will study later other contributors.

NL66278: I will add you. From git blame I see a lot op people worked on this. Especially Jairo Lopez. Also there were a lot of contributions from Stefan after he left Therp, then other people contributed as well.

Note that I only changed the format of the copyright and the enddate.

NL66278:
@Yajo There is a syntax error in a part of code that I do not quite understand or see how to correct. Git blames you. Can you have a look? Thanx in advance: https://travis-ci.org/github/OCA/openupgradelib/builds/729920133#L349 The code is here: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_120.py#L55

Yajo:
Yes, that syntax error happens when reading the `openupgrade_120.py` file from python 2. But that's not supposed to happen, because v12 already runs on py3, so there's no need for backwards compatibility in that precise file. You can ignore it.

openupgradelib/openupgrade.py
StefanRijnhart: Can we just assign copyright to OCA and update AUTHORS.rst?

StefanRijnhart: Here is the updated list of contributors.

```
* Alexandre Fayolle
* Alexis de Lattre
* Arthur Vuillard
* Benoit Guillot
* Bogdan Stanciu
* Christoph Giesel
* Daniel Reis
* David Arnold
* David Béal
* David Vidal
* Dimitrios T. Tanis
* Dmytro Katyukha
* Ernesto Tejeda
* Florent Xicluna
* Florian Dacosta
* Graeme Gellatly
* Hector Villarreal Ortega
* Hervé Martinet
* Holger Brunn
* Ivàn Todorovich
* Jairo Llopis
* Jordi Riera
* Juan Pablo Arias
* Julien Coux
* Katherine Zaoral
* Kay Häusler
* Manuel Vázquez Acosta
* Miku Laitinen
* Miquel Raich
* Moisés López
* Paulius Sladkevičius
* Pedro M. Baeza
* Petar Najman
* Robert Rübner
* Ronald Portier
* Sandy Carter
* Sebastien Alix
* Silvija Butko
* Stefan Rijnhart
* Stephane Le Cornec
* Stéphane Bidoul
* Sylvain LE GAL
* Tom Blauwendraat
* Yann Bongiovanni
* Yuri Quintana
```

openupgradelib/openupgrade.py
```diff
@@ -1704,7 +1689,6 @@ def wrap(func):
         def wrapped_function(*args, **kwargs):
```
Yajo, line 1689: IMHO it's better if you do:

```py
    @functools.wraps(func)
    def wrap(func):
        @functools.wraps(func)
        def wrapped_function(*args, **kwargs):
```

Instead of calling `update_wrapper` at the end.

openupgradelib/openupgrade.py
NL66278: Updated AUTHORS.rst by rebasing.

openupgradelib/openupgrade.py
NL66278: @Yajo why would that be better?

Yajo: it seems more idiomatic to me. Maybe just a matter of taste.

NL66278: @Yajo I changed it according to your suggestion. But actually only need one wraps per function.

Yajo (approved):

pedrobaeza:
Please fix Travis

NL66278:
@pedrobaeza The Travis errors seem to be quite unrelated to what is done in this PR. Not quite clear to me why these are occuring now.

pedrobaeza:
I have just restarted a build in main branch, and it's green:

https://travis-ci.com/github/OCA/openupgradelib/builds/219128575

so something here is provoking the error, but I can't say right now, as the log is not saying anything useful. Take into account that this lib is tested across all OpenERP/Odoo versions.

StefanRijnhart:
@NL66278 We don't have prehook here yet, so we still have the 79 character limit
```
/home/travis/build/OCA/openupgradelib/build/lib/openupgradelib/openupgrade.py:1869:80: E501 line too long (87 > 79 characters)
```

NL66278:
@StefanRijnhart Everything is within the 79 chars now.

StefanRijnhart (approved):
Thanks!

PR #216, opened 2020-10-05
[FIX] Create a method to decide what models with res_model and res_id  should be considered when merging records

JordiBForgeFlow:
Create a method to decide what models with res_model and res_id  should be considered when merging records. use the current version number to exclude models that were not existing in previous versions, or did not have a res_model field.

cc @AaronHForgeFlow @AdriaGForgeFlow

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -401,17 +401,26 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
         )
 
 
+def _change_generic_get_models_to_replace():
+    models = [
+        ('ir.attachment', 'res_id', 'res_model'),
+        ('mail.followers', 'res_id', 'res_model'),
+        ('mail.message', 'res_id', 'model'),
+        ('rating.rating', 'res_id', 'res_model'),
+    ]
+
+    if version_info[0] >= 11:
+        models += [('calendar.event', 'res_id', 'res_model'),
+                   ('mail.activity', 'res_id', 'res_model')]
+
+    return models
+
+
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
:+1: confirm that woorks for v10:

Before this fix. that method caused this error:
`Error: Invalid field 'res_model' in leaf "<osv.ExtendedLeaf: ('res_model', '=', 'product.product') on calendar_event (ctx: )>`

openupgradelib/openupgrade_merge_records.py
```diff
@@ -401,17 +401,26 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
 
 
+def _change_generic_get_models_to_replace():
+    models = [
+        ('ir.attachment', 'res_id', 'res_model'),
+        ('mail.followers', 'res_id', 'res_model'),
+        ('mail.message', 'res_id', 'model'),
+        ('rating.rating', 'res_id', 'res_model'),
+    ]
+
```
pedrobaeza, line 411: Don't add empty lines inside a method.

pedrobaeza (commented):
I created this for v11, so there's no guarantee of the rest of the things working as well for v10, but if you want to try and report back, go ahead.

legalsylvain (approved):
Lgtm. Thanks. (Not tested)

AdriaGForgeFlow (approved):

openupgradelib/openupgrade_merge_records.py
pedrobaeza: Please remove these empty lines and I'll merge.

PR #217, opened 2020-10-07
[FIX] merge_records: assure columns exist in _change_generic

MiquelRForgeFlow:
I think this fix is better and clearer than https://github.com/OCA/openupgradelib/pull/216.

Thus, if this one is merged, then https://github.com/OCA/openupgradelib/pull/216 should be closed.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -419,6 +419,9 @@ def _change_generic(env, model_name, record_ids, target_record_id,
         if (model._table, res_id_column) in exclude_columns:
             continue
         if method == 'orm':
+            if not model._fields.get(model_column) or \
+                    not model._fields.get(res_id_column):
+                continue
             records = model.search([
                 (model_column, '=', model_name),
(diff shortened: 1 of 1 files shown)


pedrobaeza:
What about non existing models?

MiquelRForgeFlow:
They are already filtered in https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_merge_records.py#L418.

pedrobaeza (approved):
@JordiBForgeFlow what do you think about this alternative?

AaronHForgeFlow (approved):
@pedrobaeza It seems to work from my side, tested in v10.

pedrobaeza:
Yes, I'm waiting for @JordiBForgeFlow blessing for closing his one.

JordiBForgeFlow:
@AaronHForgeFlow but we would need to check with https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_merge_records.py#L418 instead. We can close

pedrobaeza:
This one is not the one to close in theory, but yours.

PR #218, opened 2020-12-06
[ADD] Support for --upgrade-path

StefanRijnhart:
Adapting `load_data` to search for the file to load in the --upgrade-path if it cannot be found in the module directory tree.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1,23 +1,6 @@
-# -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#    This module copyright (C) 2011-2013 Therp BV (<http://therp.nl>)
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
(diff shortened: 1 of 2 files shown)


openupgradelib/openupgrade.py
```diff
@@ -2558,3 +2551,22 @@ def convert_to_company_dependent(
...
+def _load_apriori(path=None):
+    """ Load a 14.0+ Apriori object from apriori.json in the root of the
+    --upgrade-path """
+    if path is None:
+        from odoo import tools
+        path = tools.config['upgrade_path']
+    apriori_path = os.path.join(path, "apriori.json")
+    try:
+        with open(apriori_path) as json_file:
+            apriori_dict = json.load(json_file)
```
Yajo, line 2565:
```suggestion
            apriori_dict = yaml.safe_load(json_file)
```

:laughing: Well you know already probably, but just cross-linking https://github.com/legalsylvain/server-tools/pull/4#pullrequestreview-545971834 and https://github.com/legalsylvain/openupgrade-framework/pull/4#discussion_r537365871.

StefanRijnhart: Yes, the JSON was ugly. We're reverting to 'openupgrade_framework as an odoo module' and we can now import apriory as a Python module again.

legalsylvain (approved):
thanks.

Yajo (approved):

StefanRijnhart:
Thanks for the reviews. I noticed a slight logical error that I just fixed
```--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -278,7 +278,9 @@ def load_data(cr, module_name, filename, idref=None, mode='init'):
         if tools.config.get('upgrade_path'):
             pathname = os.path.join(
                 tools.config['upgrade_path'], module_name, filename)
-        fp = open(pathname)
+            fp = open(pathname)
+        else:
+            raise
 
     try:
         if ext == '.csv':
```

pedrobaeza (approved):

PR #219, opened 2020-12-17
[RFR] Merge records in case of an existing module category in rename_xmlids

StefanRijnhart:
As was implemented in the pre-migration script of 13.0/base: https://github.com/OCA/OpenUpgrade/blob/13.0/odoo/addons/base/migrations/13.0.1.3/pre-migration.py#L752

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -887,7 +887,7 @@ def rename_models(cr, model_spec):
     # TODO: signal where the model occurs in references to ir_model
 
 
-def rename_xmlids(cr, xmlids_spec):
+def rename_xmlids(cr, xmlids_spec, allow_merge=False):
     """
     Rename XML IDs. Typically called in the pre script.
     One usage example is when an ID changes module. In OpenERP 6 for example,
@@ -895,16 +895,57 @@ def rename_xmlids(cr, xmlids_spec):
     although they were still being defined in their respective module).
 
     :param xmlids_spec: a list of tuples (old module.xmlid, new module.xmlid).
+    :param allow_merge: if the new ID already exists, try to merge the records.
+        This is recommended when renaming module categories, which are
+        generated on the fly by the Odoo database initialization routine and
+        may resurface over a longer period of time. In general though, this
+        option should be avoided. Renaming to existing IDs is usually an
+        error, and because this method is usually called in the pre-stage,
+        the applied merge method is by SQL which is incomplete and can lead
+        to inconsistencies in the database.
     """
+    get_data_query = (
+        """SELECT res_id, model FROM ir_model_data
+        WHERE module=%s AND name=%s""")
     for (old, new) in xmlids_spec:
         if '.' not in old or '.' not in new:
             logger.error(
                 'Cannot rename XMLID %s to %s: need the module '
                 'reference to be specified in the IDs' % (old, new))
-        else:
-            query = ("UPDATE ir_model_data SET module = %s, name = %s "
-                     "WHERE module = %s and name = %s")
-            logged_query(cr, query, tuple(new.split('.') + old.split('.')))
+            continue
+        cr.execute(get_data_query, tuple(old.split('.')))
+        old_row = cr.fetchone()
+        if not old_row:
+            logger.info('XMLID %s not found when renaming to %s', old, new)
+            continue
+        if allow_merge:
+            cr.execute(get_data_query, tuple(new.split('.')))
+            new_row = cr.fetchone()
+            if new_row:
+                logger.info(
+                    'XMLID %s already exists when renaming from %s: Merging.',
+                    new, old)
+                if new_row[1] != old_row[1]:
+                    do_raise(
+                        "Cannot merge XMLIDs %s, %s because they don't belong "
+                        "to the same model (%s, %s)" % (
+                            old, new, old_row[1], new_row[1]))
+                table = old_row[1].replace('.', '_')
+                if not table_exists(cr, table):
+                    do_raise(
+                        "Cannot merge XMLIDs %s, %s because the table I "
+                        "guessed (%s) based on the model name (%s) does not "
+                        "exist." % (old, new, table, old_row[1]))
+                # Cannot import merge_records until after Odoo initialization
+                from .openupgrade_merge_records import merge_records
+                env = api.Environment(cr, SUPERUSER_ID, {})
+                merge_records(
+                    env, old_row[1], [old_row[0]], new_row[0],
+                    method="sql", model_table=table)
+                continue
+        query = ("UPDATE ir_model_data SET module = %s, name = %s "
+                 "WHERE module = %s and name = %s")
+        logged_query(cr, query, tuple(new.split('.') + old.split('.')))
 
 
 def add_xmlid(cr, module, xmlid, model, res_id, noupdate=False):


openupgradelib/openupgrade.py
```diff
@@ -896,15 +896,41 @@ def rename_xmlids(cr, xmlids_spec):
...
-            logged_query(cr, query, tuple(new.split('.') + old.split('.')))
+            continue
+
+        cr.execute(get_data_query, tuple(old.split('.')))
+        old_row = cr.fetchone()
+        if not old_row:
+            logger.warning('XMLID %s not found when renaming to %s', old, new)
+            continue
+
+        if old_row[1] == 'ir.module.category':
```
StefanRijnhart, line 916: Or, alternatively, add a parameter to the method to allow merging regardless of the model.

openupgradelib/openupgrade.py
```diff
@@ -896,15 +896,41 @@ def rename_xmlids(cr, xmlids_spec):
...
-        else:
-            query = ("UPDATE ir_model_data SET module = %s, name = %s "
-                     "WHERE module = %s and name = %s")
-            logged_query(cr, query, tuple(new.split('.') + old.split('.')))
+            continue
+
+        cr.execute(get_data_query, tuple(old.split('.')))
+        old_row = cr.fetchone()
+        if not old_row:
+            logger.warning('XMLID %s not found when renaming to %s', old, new)
```
MiquelRForgeFlow, line 913: I don't know if it's a good idea to log this.

openupgradelib/openupgrade.py
MiquelRForgeFlow: Oh, that would be even better. But then what about the model and model_table?

openupgradelib/openupgrade.py
StefanRijnhart: Could help to detect typos

openupgradelib/openupgrade.py
StefanRijnhart: Determine table by replacing dots. Should work in 99.9% of the time.

MiquelRForgeFlow: Maybe it would be better pass the table in the merge parameter you suggested. Imagine having xmlids you would like to merge and others that not, then you would have to create two lists and that would be a bit awful.

What do you think of the following?

```python
openupgrade.rename_xmlids(env.cr, [
    ('old_id1', 'new_id1'),  # no merge
    ('old_id2', 'new_id2', 'table2'), # merge
    ('old_id3', 'new_id3'), # no merge
    ...
])

def rename_xmlids(cr, xmlids_spec):
    for spec in xmlids_spec:
        if len(spec) == 3:
            old, new, table = spec
        else:
            old, new = spec
            table = None
        (...)
        if table:
            # proceed with the fetch new_row and merge code
        (...)
```

StefanRijnhart: Like I said, the current approach will work for 99.9% of the time, whereas the need to rename is only there for 0.01% of the time.

I pushed the new approach.

StefanRijnhart: Note that I took care to catch problems with the guessed table and a raise clear error message.

openupgradelib/openupgrade.py
```diff
@@ -887,24 +887,65 @@ def rename_models(cr, model_spec):
...
     Rename XML IDs. Typically called in the pre script.
     One usage example is when an ID changes module. In OpenERP 6 for example,
     a number of res_groups IDs moved to module base from other modules (
     although they were still being defined in their respective module).
 
     :param xmlids_spec: a list of tuples (old module.xmlid, new module.xmlid).
+    :param allow_merge: if the new ID already exists, try to merge the records.
+    This is recommended when renaming module categories, which are generated
+    on the fly by the Odoo database initialization routine and may resurface
+    over a longer period of time.
```
MiquelRForgeFlow, line 901: to greenify ci

```yml
Warning, treated as error:

/home/travis/build/OCA/openupgradelib/openupgradelib/openupgrade.py:docstring of openupgradelib.openupgrade.rename_xmlids:8:Field list ends without a blank line; unexpected unindent.
```

not sure if it's this:
```suggestion
    :param allow_merge: if the new ID already exists, try to merge the records.
     This is recommended when renaming module categories, which are generated
     on the fly by the Odoo database initialization routine and may resurface
     over a longer period of time.
```

or this:
```suggestion

    :param allow_merge: if the new ID already exists, try to merge the records.
    This is recommended when renaming module categories, which are generated
    on the fly by the Odoo database initialization routine and may resurface
    over a longer period of time.
```

or both

legalsylvain (approved):
LGTM. Thanks.

openupgradelib/openupgrade.py
StefanRijnhart: Thanks for suggesting this fix!

openupgradelib/openupgrade.py
pedrobaeza: I prefer this to be `info`, as there are legit cases where the XML-ID won't be found, and having it as warning may disturb. Example (real that happened on v12, but names may not be accurate): you have the XML-ID `sale.group_discount` that is now in v12 `account.group_discount`. On pre-migration of account, you do such renaming for having things in the proper place, but if you don't have `sale` account, the XML-ID won't exist.

openupgradelib/openupgrade.py
```diff
@@ -887,24 +887,65 @@ def rename_models(cr, model_spec):
...
+                    do_raise(
+                        "Cannot merge XMLIDs %s, %s because the table I "
+                        "guessed (%s) based on the model name (%s) does not "
+                        "exist." % (old, new, table, old_row[1]))
+                # Cannot import merge_records until after Odoo initialization
+                from .openupgrade_merge_records import merge_records
+                env = api.Environment(cr, SUPERUSER_ID, {})
+                merge_records(
+                    env, old_row[1], [old_row[0]], new_row[0],
+                    method="sql", model_table=table)
```
pedrobaeza, line 943: SQL merge method is not totally implemented, so it can give the false impression of doing everything. I see this very risky.

openupgradelib/openupgrade.py
```diff
@@ -887,24 +887,65 @@ def rename_models(cr, model_spec):
...
-                     "WHERE module = %s and name = %s")
-            logged_query(cr, query, tuple(new.split('.') + old.split('.')))
+            continue
+
+        cr.execute(get_data_query, tuple(old.split('.')))
+        old_row = cr.fetchone()
+        if not old_row:
+            logger.warning('XMLID %s not found when renaming to %s', old, new)
+            continue
+
```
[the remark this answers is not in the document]
StefanRijnhart: FIxed

openupgradelib/openupgrade.py
```diff
@@ -887,24 +887,65 @@ def rename_models(cr, model_spec):
...
+            if new_row:
+                logger.info(
+                    'XMLID %s already exists when renaming from %s: Merging.',
+                    new, old)
+                if new_row[1] != old_row[1]:
+                    do_raise(
+                        "Cannot merge XMLIDs %s, %s because they don't belong "
+                        "to the same model (%s, %s)" % (
+                            old, new, old_row[1], new_row[1]))
+                table = old_row[1].replace('.', '_')
```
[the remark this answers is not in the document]
StefanRijnhart: I know. But pasing `allow_merge=True` should be rare in itself. So I don't think it's a problem if does not cover all cases. Or is it safe to use `env[model]._table` in the pre-script of the base module?

openupgradelib/openupgrade.py
StefanRijnhart: Again, it's more or less to prevent us from having to copy a whole method to rename ir.module.category entries safely in every edition so I think it's safe enough. I can add some more disclaimers to the docstring if you like.

openupgradelib/openupgrade.py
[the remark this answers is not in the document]
hbrunn: [no](https://github.com/OCA/OCB/blob/14.0/odoo/modules/loading.py#L173-L198)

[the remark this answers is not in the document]
pedrobaeza: OK, then add such disclaimers for not being used massively in any XML-IDs merge.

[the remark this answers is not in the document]
StefanRijnhart: OK, disclaimer added.

pedrobaeza (approved):
Thank you!

PR #220, opened 2020-12-23
Add functions to autoupdate unmodified COW views

Yajo:
In Odoo v12+, with `website` installed, and with its COW system for view edits, when a user clicks enables or disables a view marked as `customize_show=True` by using the dropdown in the website editor, Odoo creates a COW website-specific view which is enabled or disabled as desired.

When updating, COW views are usually not updated. If they contain any logic, it can lead to template rendering problems easily.

These errors make sense on some circumstances, but if it was just because the user used the customize show feature, they are disturbing and meaningless. With these methods provided here, this problem can be reduced to only meaninful occurrences.

The 1s method will compare website-specific views with their agnostic counterparts. If bodies are equal, they will be marked.

The 2nd method takes all those marked views and resets their bodies to the new upstream, which possibly comes from the module and is updated for the new Odoo version.

@Tecnativa TT23136

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -163,6 +163,8 @@ __all__ = [
     'delete_records_safely_by_xml_id',
     'set_xml_ids_noupdate_value',
     'convert_to_company_dependent',
+    'cow_templates_mark_if_equal_to_upstream',
+    'cow_templates_replicate_upstream',
 ]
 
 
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -2606,3 +2608,86 @@ def convert_to_company_dependent(
...
+       pre-migration.
+    """
+    mark_colname = mark_colname or get_legacy_name("cow_equal_to_upstream")
+    mark_identifier = sql.Identifier(mark_colname)
+    logged_query(
+        cr,
+        sql.SQL("""
+            UPDATE ir_ui_view AS specific
+            SET arch_db = agnostic.arch_db
+            FROM ir_ui_view AS agnostic
```
pedrobaeza, line 2686:
```suggestion
            FROM ir_ui_view AS generic
```

openupgradelib/openupgrade.py
```diff
@@ -2606,3 +2608,86 @@ def convert_to_company_dependent(
...
+            WHERE
+                specific.{} AND
+                agnostic.website_id IS NULL AND
+                specific.key = agnostic.key
```
pedrobaeza, line 2690:
```suggestion
                specific.key = agnostic.key AND
                specific.type = 'qweb' AND
                specific.key IS NOT NULL AND
                generic.type = 'qweb'
```

pedrobaeza (changes requested):
Can't be done in a step?

Yajo:
It's not possible. 🤷🏼‍♂️ If you were to compare a specific view with its generic counterpart in post-update, maybe the generic one would have been updated, producing a false negative that would never update a view. The need to split in pre and post is technical.

pedrobaeza (approved):
Thanks for this!

openupgradelib/openupgrade.py
```diff
@@ -2606,3 +2608,89 @@ def convert_to_company_dependent(
...
+    if not column_exists(cr, "ir_ui_view", mark_colname):
+        logged_query(
+            cr,
+            sql.SQL("ALTER TABLE ir_ui_view ADD COLUMN {} BOOLEAN")
+            .format(mark_identifier))
+    # Map all qweb views
+    cr.execute("""
+        SELECT id, arch_db, key, website_id
+        WHERE type == 'qweb' AND key IS NOT NULL
+    """)
```
kos94ok-3D, line 2636: I don`t see `FROM` statement

Yajo: Oops... https://github.com/OCA/openupgradelib/pull/223

openupgradelib/openupgrade.py
```diff
@@ -2606,3 +2608,89 @@ def convert_to_company_dependent(
...
+        # ever touching the view, the COW system duplicates `arch_db`
+        # preserving even the same whitespace.
+        # In case this is proven false under some circumstances, we would need
+        # to do a smarter XML comparison here.
+        if arch_db == upstream_arch:
+            equal.append(id_)
+    # Mark equal views
+    logged_query(
+        cr,
+        sql.SQL("UPDATE ir_ui_view SET {} = TRUE WHERE ID IN %s")
```
MiquelRForgeFlow, line 2661: error in this query. see website PR

Yajo: See #225

PR #221, opened 2020-12-28
[FIX] add_xmlid: 14.0 compatibility

hildickethan:
When trying to use the `add_xmlid` function in a 14.0 repo it returned the following error:

Traceback error:
```
  File "/opt/odoo/src/repo/addons/x_module/hooks.py", line 17, in pre_init_hook
    openupgrade.add_xmlid(
  File "/opt/odoo/data/.cache/pypi/openupgradelib/openupgrade.py", line 985, in add_xmlid
    logged_query(
  File "/opt/odoo/data/.cache/pypi/openupgradelib/openupgrade.py", line 1310, in logged_query
    cr.execute(query, args)
  File "<decorator-gen-3>", line 2, in execute
  File "/opt/odoo/src/odoo/odoo/sql_db.py", line 101, in check
  File "/opt/odoo/src/odoo/odoo/sql_db.py", line 298, in execute
Exception

The above exception was the direct cause of the following exception:

  File "/opt/odoo/src/odoo/odoo/http.py", line 639, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/src/odoo/odoo/http.py", line 315, in _handle_exception
    raise exception.with_traceback(None) from new_cause
psycopg2.errors.UndefinedColumn: column "date_init" of relation "ir_model_data" does not exist
LINE 1: ... (create_uid, create_date, write_uid, write_date, date_init,...
```

SQL error:
```
odoo.sql_db: bad query: INSERT INTO ir_model_data (create_uid, create_date, write_uid, write_date, date_init, date_update, noupdate, name, module, model, res_id) VALUES (1, (now() at time zone 'UTC'), 1, (now() at time zone 'UTC'), (now() at time zone 'UTC'), (now() at time zone 'UTC'), true, 'category_x', 'x_module', 'product.category', 8)
ERROR: column "date_init" of relation "ir_model_data" does not exist
LINE 1: ... (create_uid, create_date, write_uid, write_date, date_init,...
```

It seems like this is caused, as stated by the error, by the lack of `date_init` and `date_update` existing as fields in `ir.model.data` in 14.0. Removed in this commit https://github.com/odoo/odoo/commit/148f3dabc91fbb59ae5abe6396ca614a708577b9.

My workaround is simply using a different SQL without these fields in the case of an Odoo version of 14.0 or higher and that worked fine for my use/test cases. If anyone has a deeper understanding of these models and can give some insight that would be appreciated!

I have tested this openupgradelib change in 14.0 and 13.0 and it worked as intended.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -982,16 +982,43 @@ def add_xmlid(cr, module, xmlid, model, res_id, noupdate=False):
     if already_exists:
         return False
     else:
+        query = "INSERT INTO ir_model_data ({fields}) VALUES ({values})"
+        fields = [
+            "create_uid",
+            "create_date",
+            "write_uid",
+            "write_date",
+            "noupdate",
+            "name",
+            "module",
+            "model",
+            "res_id",
+        ]
+        args = (
+            SUPERUSER_ID,
+            AsIs("(now() at time zone 'UTC')"),
+            SUPERUSER_ID,
+            AsIs("(now() at time zone 'UTC')"),
+            noupdate,
+            xmlid,
+            module,
+            model,
+            res_id,
+        )
+        if version_info[0] < 14:
+            fields += ["date_init", "date_update"]
+            args += (
+                AsIs("(now() at time zone 'UTC')"),
+                AsIs("(now() at time zone 'UTC')"),
+            )
         logged_query(
             cr,
-            "INSERT INTO ir_model_data (create_uid, create_date, "
-            "write_uid, write_date, date_init, date_update, noupdate, "
-            "name, module, model, res_id) "
-            "VALUES (%s, (now() at time zone 'UTC'), %s, "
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Thanks for the patch. For the sake of code clarity, it would be good to have a base query string and args, and only on version < 14, add the rest of query and args. Can you please do it?

hildickethan:
> Thanks for the patch. For the sake of code clarity, it would be good to have a base query string and args, and only on version < 14, add the rest of query and args. Can you please do it?

@pedrobaeza Updated! I re-tested on 13.0 and 14.0 and it adds the fields correctly in the lower version, let me know if there is a simpler solution as I may have overcomplicated it a bit

pedrobaeza (approved):

StefanRijnhart (approved):
Thanks!

PR #226, opened 2021-01-29
[FIX] merge_records: complete _change_translations_sql method

MiquelRForgeFlow:
This method was incomplete.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -248,26 +248,49 @@ def _change_translations_orm(env, model_name, record_ids, target_record_id,
             records = records[1:]
         if records:
             records.unlink()
-            logger.debug("Deleted %s extra translations for %s.",
-                         len(records), group['name'])
+            logger.debug("Deleted %s extra translations for %s (lang = %s).",
+                         len(records), group['name'], group['lang'])
 
 
 def _change_translations_sql(env, model_name, record_ids, target_record_id,
                              exclude_columns):
     if ('ir_translation', 'res_id') in exclude_columns:
         return
-    # TODO: To be handled with the same approach as in ORM method
     logged_query(
-        env.cr,
-        """
-        UPDATE ir_translation SET res_id = %(target_record_id)s
-        WHERE type = 'model' AND res_id in %(record_ids)s
-        AND name like %(model_name)s || ',%%'""",
+        env.cr, """
+        UPDATE ir_translation it
+        SET res_id = %(target_record_id)s
+        FROM (
+            SELECT min(it.id) as id, it.name, it.lang
+            FROM ir_translation it
+            LEFT JOIN ir_translation it2 ON (
+                it2.type = it.type AND it2.name = it.name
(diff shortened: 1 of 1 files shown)


HviorForgeFlow (approved):
Works fine! :+1:

pedrobaeza:
Why can't you do things in one query instead of a lot of select?

MiquelRForgeFlow:
What do you mean? I only did a verbatim copy of orm method, but with sql queries instead of using method search.

pedrobaeza:
I mean that you have put 3 SELECT query, and that seems to be a non optimized way to do SQL. Please try to join the final action to do (I understand `DELETE`) in only query.

MiquelRForgeFlow:
Done, fused the queries. But I prefer what it was done before, as it was more readable.

pedrobaeza:
But for logging purposes this is far better, and also for performance. Do you prefer to have 1 logged query or 200 (one per iteration of the loop)?

pedrobaeza:
And this is still not what I was asking for. You still do things in 2 steps. `UPDATE` or `DELETE` queries should include everything, not a previous `SELECT` and then the rest.

MiquelRForgeFlow:
I doubt clients have bbdd with 200 name/lang installed haha but ok, I will improve the way you say.

pedrobaeza:
It was just an example of general problem, but yes, I have customer with 7 languages. Anyway, I think you are not understanding yet the problem. Imagine this simplified case:
`SELECT id FROM res_partner WHERE customer`
`UPDATE res_partner SET name = name WHERE id in %s`
and get the logging with `UPDATE res_partner SET name = name WHERE id in (1, 2, 3, 4....21103)`
instead of a cleaner:
`UPDATE res_partner SET name = name WHERE customer`
The idea is the same for this, but with more difficult statements.

MiquelRForgeFlow:
Done, how it looks now?

openupgradelib/openupgrade_merge_records.py
```diff
@@ -248,26 +248,50 @@ def _change_translations_orm(env, model_name, record_ids, target_record_id,
...
         }, skip_no_result=True)
+    logged_query(
+        env.cr, """
+        DELETE FROM ir_translation it
+        USING (
+            SELECT it.id
+            FROM ir_translation it
+            WHERE it.type = 'model' AND it.res_id IN %(record_ids)s
+                AND it.name like %(model_name)s || ',%%'
+                AND it.res_id != %(target_record_id)s
```
pedrobaeza, line 287: Shouldn't this be?
```suggestion
                AND it.res_id IN %(record_ids)s
```

MiquelRForgeFlow: See two lines above, please hehe. This extra line is to discard one of those records

pedrobaeza: Ah, OK, then it's not needed, as `target_record_id` is not included in `record_ids`

MiquelRForgeFlow: Well explained. Updated!

PR #227, opened 2021-02-05
[IMP] Warn when an existing field is encountered in rename_fields

StefanRijnhart:
This will signal the cases when a field is renamed twice due to a programming error. Deleting the ir_model_field can be really damaging because for instance ir_property records are then deleted by cascade as we saw in https://github.com/OCA/OpenUpgrade/pull/2531

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -597,9 +597,29 @@ def rename_fields(env, field_spec, no_deep=False):
         # Delete possible existing field entry
         # Example: https://github.com/OCA/OpenUpgrade/issues/2339
         cr.execute(
-            """DELETE FROM ir_model_fields WHERE name = %s AND model = %s""",
+            """DELETE FROM ir_model_fields WHERE name = %s AND model = %s
+            RETURNING id""",
             (new_field, model),
         )
+        row = cr.fetchone()
+        if row:
+            cr.execute(
+                """
+                SELECT module||'.'||name
+                FROM ir_model_data WHERE model = 'ir.model.fields'
+                    AND res_id = %s
+                """, (row[0],))
+            xmlids = [xmlid for xmlid, in cr.fetchall()]
+            lg = logger.error if xmlids else logger.warning
+            lg(
+                "When renaming field %s of model %s to %s, an existing field "
+                "with the new name was encountered. This could be a "
+                "error in a migration script, or a remnant from a previous "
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Not sure the check over `ir.model.data` is a good indicator for highlighting errors, as you can see in https://github.com/OCA/OpenUpgrade/issues/2339 that a legitimate migration process can lead to such problem.

StefanRijnhart:
I just want Travis to break whenever this occurs. When migrating customers, the error can be reviewed and ignored or suppressed by removing the entry beforehand.

StefanRijnhart:
When migrating from 8.0 to 13, the ir_model_data entry should have been removed so a warning is logged instead of an error. So unless it's not working as I think it is, having the entry in ir_model_data should raise higher alarm levels.

pedrobaeza:
How is it removed if you migrate in batch without intermediate database cleanup?

StefanRijnhart:
Oh dear, you are right. The XMLIDs of fields removed after Odoo 8 are still in my version 12 migrated database. That seems like a bug to me because if I add a field in Odoo 12 and I remove it again the ir.model.data xmlid is removed by the unlinking triggered by _process_end, just like any other xmlid that is not marked as noupdate. I'll investigate.

StefanRijnhart:
Holy cow, it's because _process_end searches for noupdate=FALSE (https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/ir/ir_model.py#L1319-L1320) explicitely but ir_model_fields were (I think until Odoo 12) created with noupdate=NULL.

How about I run some tests where I run my current migration project to also remove items with noupdate=NULL to see what the impact is of that? That should fix https://github.com/OCA/OpenUpgrade/issues/2339 and then we can just update this PR to raise an exception if the field entry already exists.

pedrobaeza:
Well, it's a plan. The problem I see is that the current migrated databases contain that incorrect data, so we should be resilient to that.

StefanRijnhart:
We can update existing noupdate entries from null to false in the base init script of 12, 13 and 14 for that.

NL66278 (approved):
:+1: LGTM

StefanRijnhart:
Obsolete because we no longer delete ir.model.fields in openupgradelib anymore. The branches linked to https://github.com/OCA/OpenUpgrade/issues/2339 backport deleting `ir.model.fields` and `ir.model` from Odoo 13.0

PR #229, opened 2021-02-09
[RVT] rename fields: revert deletion of existing ir_model_fields

StefanRijnhart:
The deletion of ir_model_fields is harmful as it drops ir.property by cascade. This is currently what happens with product settings in OpenUpgrade 9.0. I'm working on an actual fix, but it would be good to prevent this unnoticed destruction of data in the mean time.


pedrobaeza:
OK, agree to revert meanwhile:

/ocabot merge nobump

StefanRijnhart:
Thank you!

PR #231, opened 2021-03-05
[TST] Test commit to see why travis is failing

NL66278:


NL66278:
@StefanRijnhart Thanx, I had to come back to this one. Anyway, saw that travis was failing no longer.

PR #232, opened 2021-03-16
[IMP] Defer invalidating filters

StefanRijnhart:
When bridging multiple versions during a migration, custom modules are
usually only available in the target version. If we run this method with
every version, filters that depend on fields from these modules will be
invalidated prematurely. Also, running the method to invalidate filters on
every version has a performance impact.

The target version can be provided with an environment variable

```
export OPENUPGRADE_TARGET_VERSION=13.0
```

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -110,6 +110,14 @@ if version_info[0] > 7:
 else:
     api = False
 
+# Setting the target version as an environment variable allows OpenUpgrade
+# to skip methods that are called in every version but really only need to
+# run in the target version. Make the target version available to OpenUpgrade
+# with `export OPENUPGRADE_TARGET_VERSION=13.0` (when migrating up to 13.0)
+target_version = os.environ.get("OPENUPGRADE_TARGET_VERSION")
+if target_version:
+    is_target_version = version_info[0] == int(float(target_version))
+
 
 # The server log level has not been set at this point
 # so to log at loglevel debug we need to set it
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -110,6 +110,15 @@ def do_raise(error):
...
     api = False
 
+# Setting the target version as an environment variable allows OpenUpgrade
+# to skip methods that are called in every version but really only need to
+# run in the target version. Make the target version available to OpenUpgrade
+# with `export OPENUPGRADE_TARGET_VERSION=13.0` (when migrating up to 13.0)
+target_version = os.environ.get("OPENUPGRADE_TARGET_VERSION")
+if target_version:
+    is_target_version = version_info[0] == int(float(
+        os.environ["OPENUPGRADE_TARGET_VERSION"]))
```
pedrobaeza, line 120:
```suggestion
        target_version))
```

pedrobaeza (approved):

openupgradelib/openupgrade.py
```diff
@@ -110,6 +110,14 @@ def do_raise(error):
 else:
     api = False
 
+# Setting the target version as an environment variable allows OpenUpgrade
+# to skip methods that are called in every version but really only need to
+# run in the target version. Make the target version available to OpenUpgrade
+# with `export OPENUPGRADE_TARGET_VERSION=13.0` (when migrating up to 13.0)
```
hbrunn, line 116: I'd like to see this paragraph somewhere in the docs too

hbrunn (changes requested):

pankk (approved):
Technical review.

openupgradelib/openupgrade.py
StefanRijnhart: Thanks for suggesting! I have now created https://github.com/OCA/OpenUpgrade/pull/2619 for that.

hbrunn (approved):

PR #235, opened 2021-04-20
[ADD] safe_unlink method

StefanRijnhart:


openupgradelib/openupgrade.py
```diff
@@ -2486,6 +2487,30 @@ def update_module_moved_fields(
...
 
 
+def safe_unlink(records, do_raise=False):
+    """Allow for errors to occur during unlinking of records.
+
+    Prevent broken database transactions, and by default, catch exceptions.
+
+    :param records: an iterable (not necessarily recordset) of records to
+    unlink.
+    :param do_raise: when set to True, don't catch the exception but only log
```
legalsylvain, line 2497: It should be "When set to False" don't you think ?

openupgradelib/openupgrade.py
```diff
@@ -2486,6 +2487,30 @@ def update_module_moved_fields(
...
+        logger.debug("Deleting record %s#%s", record._name, record.id)
+        if not record.exists():
+            continue
+        try:
+            with record.env.cr.savepoint():
+                record.unlink()
+        except Exception as e:
+            if do_raise:
+                raise
+            logger.error("Error deleting %s#%s: %s",
```
pedrobaeza, line 2510: I think it's better to not log ERROR, as it can be something expected (and that's why you build this method). I would go just for INFO.

openupgradelib/openupgrade.py
StefanRijnhart: No, that's the default value. But when set to True, it does raise.

openupgradelib/openupgrade.py
StefanRijnhart: OK, updated.

openupgradelib/openupgrade.py
```diff
@@ -2494,11 +2519,11 @@ def delete_records_safely_by_xml_id(env, xml_ids):
...
+        if not record:
+            continue
         try:
-            with env.cr.savepoint():
-                record = env.ref(xml_id, raise_if_not_found=False)
-                if record and record.exists():
-                    record.unlink()
+            safe_unlink(record, do_raise=True)
         except Exception as e:
             logger.error('Error deleting XML-ID %s: %s', xml_id, repr(e))
```
pedrobaeza, line 2528: As you are revisiting this, we can as well change this to INFO, as I found this annoying other times.

StefanRijnhart: Done

pedrobaeza (approved):

openupgradelib/openupgrade.py
legalsylvain: Yes.

So better to replace ``when set to True, don't catch the exception but only log  an error`` by  ``when set to True, don't catch the exception and raise the error``.
Do'nt you think ?

StefanRijnhart: Docstring updated.

legalsylvain (approved):

legalsylvain:
travis seems to fail for flake8 reason : 
https://travis-ci.com/github/OCA/openupgradelib/builds/223592703#L11345

``openupgradelib/openupgradelib/openupgrade.py:docstring of openupgradelib.openupgrade.safe_unlink:6:Field list ends without a blank line; unexpected unindent.``

pedrobaeza:
Uhm, Travis is now red. Do you know why?

StefanRijnhart:
Thanks for noting! I've now properly indented the wrapped lines in the docstring `:params:`

PR #237, opened 2021-04-28
[UPD] update_module_moved_fields: unneeded since v13

RaulOForgeFlow:
Added that this module is not needed since v13 in update_module_moved_fields docstring

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2456,6 +2456,9 @@ def update_module_moved_fields(
     """Update module for field definition in general tables that have been
     moved from one module to another.
 
+    No need to use this method: moving the XMLID is covered in
+    Odoo and OpenUpgrade natively.
+
     :param cr: Database cursor
     :param model: model name
     :param moved_fields: list of moved fields


pedrobaeza (approved):

pedrobaeza:
@StefanRijnhart do you ratify this?

StefanRijnhart:
Moving the field definition in ir_model_data was always taken care of by the OpenUpgrade framework except for the bug in OpenUpgrade 11 that was recently fixed in https://github.com/OCA/OpenUpgrade/pull/2537/files#diff-30ddfe4066ca987c45ebeb4bb2703fc4fa7477c4eeb2cdcec69b9c685787a8f2L831.

pedrobaeza:
Then you mean that we should advertise this as to never be used? What about translations?

StefanRijnhart:
Moving the translations over to another module is not covered indeed. It seems to me that the impact of having the *module* translations of moved fields refreshed is very limited.

pedrobaeza:
OK, I think we can put this note as the other versions are very old, and let it that way with the migration scripts that already use them in v11/v12, and let translations part as something uncovered, but with limited impact. Do you agree?

openupgradelib/openupgrade.py
```diff
@@ -2456,6 +2456,8 @@ def update_module_moved_fields(
     """Update module for field definition in general tables that have been
     moved from one module to another.
 
+    No need to use this method since v13
```
StefanRijnhart, line 2459:
```suggestion
    No need to use this method: moving the XMLID is covered in
    Odoo and OpenUpgrade natively.
```

MiquelRForgeFlow (approved):

StefanRijnhart (approved):

PR #239, opened 2021-05-18
[FIX] rename_xmlids: don't merge if hierarchy exists

MiquelRForgeFlow:
When we already have recursive hierarchy, doing a merge of a parent into one of their children let the awkward situation where the child points to itself, so we avoid it checking this condition.

Same as done in https://github.com/OCA/OpenUpgrade/pull/2730.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -30,7 +30,7 @@ def _change_foreign_key_refs(env, model_name, record_ids, target_record_id,
             JOIN information_schema.constraint_column_usage AS ccu
                 ON ccu.constraint_name = tc.constraint_name
                 AND ccu.table_schema = tc.table_schema
-            WHERE constraint_type = 'FOREIGN KEY'
+            WHERE tc.constraint_type = 'FOREIGN KEY'
             AND ccu.table_name = %s and ccu.column_name = 'id'
         """, (model_table,))
     for table, column in env.cr.fetchall():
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Lgtm. Thanks.

pedrobaeza:
I personally see this as a very special case for being handled in such generic rename method, but I let @hbrunn and @StefanRijnhart to give their opinion.

hbrunn (approved):

StefanRijnhart:
Wouldn't it be better to rewire the hierarchy? Is there a concrete case?

pedrobaeza:
This is coming from https://github.com/OCA/OpenUpgrade/pull/2730

StefanRijnhart:
OK, so this is the more generalized case. But the merging is executed by `merge_records`, so shouldn't the intervention be in there, and again, shouldn't the proper solution be to do the merge but prevent the circular reference?

kos94ok-3D:
I think better use?
image

pedrobaeza:
Yes, the field name shouldn't be hard-coded.

StefanRijnhart:
Maybe you will want to check by FK constraint instead of `_parent_name`, so that you catch Many2many references as well, such as res.group's implied_ids.

pedrobaeza (changes requested):
Pending comments

MiquelRForgeFlow:
@StefanRijnhart your comments are attended. The PR it's ready :)

openupgradelib/openupgrade_merge_records.py
```diff
@@ -682,6 +682,53 @@ def _delete_records_orm(env, model_name, record_ids, target_record_id):
...
+        if new_parent_row and new_parent_row[0] in record_ids:
+            # When we already have recursive hierarchy, doing a
+            # merge of a parent into one of their children let the
+            # awkward situation where the child points to itself,
+            # so we avoid it checking this condition
+            logger.info(
+                "Couldn't merge %s record(s) of model %s to record_id %s"
+                " to avoid recursion with field %s of table %s",
+                len(record_ids), model_name, target_record_id, origin, table)
+            return True
```
StefanRijnhart, line 728: Thanks for the refactoring. Now I'm just concerned with the semantics: I would expect a method called "check_no_recurrence" to return True if there is indeed "no recurrence" but I believe that is the other way around here, right?

MiquelRForgeFlow: hahaha Ok, I renamed it to "_check_recurrence"

StefanRijnhart (approved):
Thanks!

pedrobaeza:
I see this for detecting problems, but I think we should provide a way to auto-fix these ones, and maybe doing it through `adjust_values` (auto remove m2m references to itself for example) , or should we run it a time to see real problems?

MiquelRForgeFlow:
> or should we run it for some time to see real problems?

Yes, do that in the future, not now. At least until we know for sure we can do that auto remove without worries.

pedrobaeza (approved):

pedrobaeza:
I shouldn't merge this with such bad commit message, but there's no solution now. I will put at least a good description on the release message.

PR #241, opened 2021-07-08
[FIX] Missing filter by model when merging mail.followers using SQL

StefanRijnhart:
Regression of https://github.com/OCA/openupgradelib/pull/198


legalsylvain (commented):
thanks !

pedrobaeza (approved):

PR #244, opened 2021-09-22
Fix issue when running setup.py on python<3.8

sbidoul:


legalsylvain (approved):
Hi. 

I was just deploying on a test server, with a requirement.txt file that contains ``git+https://github.com/OCA/openupgradelib.git@master``

replacing by `` git+https://github.com/acsone/openupgradelib.git@fix-setup.py-old-python-sbi`` seems to fix the problem.

Context : ``Python 3.6.9``

thanks for the quick fix !

pedrobaeza:
Let's merge it then and create another tag.

PR #246, opened 2021-09-29
[IMP] _change_foreign_key_refs: Add extra_where params

pedrobaeza:
Adding the new argument `extra_where`, we are able to change a FK reference depending on an extra condition, so not all ocurrences are replaced.

This is useful for example when we have a unique record for all companies, and on new version, there should be one record per company. We perform the split, and we need to replace references for only a specific company (extra_where="AND company_id = X").

Retro-compatibility is kept as the argument is optional, and a safeguard has been introduced for skipping models where the condition can't be fulfilled if the field doesn't exist (so we are not able to say if the reference should be replaced or not).

@Tecnativa TT28115

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -7,7 +7,7 @@ import logging
 import functools
 from psycopg2 import sql
 from psycopg2 import ProgrammingError, IntegrityError
-from psycopg2.errorcodes import UNIQUE_VIOLATION
+from psycopg2.errorcodes import UNDEFINED_COLUMN, UNIQUE_VIOLATION
 from psycopg2.extensions import AsIs
 from .openupgrade import logged_query, version_info
 from .openupgrade_tools import column_exists, table_exists
@@ -17,7 +17,7 @@ logger.setLevel(logging.DEBUG)
 
 
 def _change_foreign_key_refs(env, model_name, record_ids, target_record_id,
-                             exclude_columns, model_table):
+                             exclude_columns, model_table, extra_where=None):
     # As found on https://stackoverflow.com/questions/1152260
     # /postgres-sql-to-list-table-foreign-keys
     # Adapted for specific Odoo structures like many2many tables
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Being used in https://github.com/OCA/OpenUpgrade/pull/2379

openupgradelib/openupgrade_merge_records.py
```diff
@@ -54,6 +58,9 @@ def _change_foreign_key_refs(env, model_name, record_ids, target_record_id,
             env.cr.execute('ROLLBACK TO SAVEPOINT sp1')
             if error.pgcode != UNIQUE_VIOLATION:
                 raise
+            elif error.pgcode == UNDEFINED_COLUMN:
+                # extra_where is introducing a bad column. Ignore this table thus
```
legalsylvain, line 62: Dont you think you should log a warning here ? It means that thé call to thé function was incorrect. Or did i missed something ?

pedrobaeza: It was incorrect, but with that exact code, so the only possible way to do that is through the extra_where (the rest of columns are got from the FK references, which are valid). I don't think we should hammer logs with that, and more on the use case we want to use it.

pedrobaeza:
It seems we have again a problem with Python libraries packaging

openupgradelib/openupgrade_merge_records.py
MiquelRForgeFlow: This line is too long, that's why CI is red.
Selection_203

MiquelRForgeFlow (approved):
LGTM

pedrobaeza:
@StefanRijnhart do you oppose to this change, as you were the original author?

openupgradelib/openupgrade_merge_records.py
```diff
@@ -54,6 +58,9 @@ def _change_foreign_key_refs(env, model_name, record_ids, target_record_id,
             env.cr.execute('ROLLBACK TO SAVEPOINT sp1')
             if error.pgcode != UNIQUE_VIOLATION:
                 raise
+            elif error.pgcode == UNDEFINED_COLUMN:
```
StefanRijnhart, line 61: Isn't the exception not raised already at this point?
Also, I would prefer that you check if `extra_where` is actually set before you assume it's causing the error here.

pedrobaeza: I don't get your first sentence, but yes, we can check if `extra_where` is set. Does this looks good to you:
```suggestion
            elif error.pgcode == UNDEFINED_COLUMN:
                if not extra_where:
                    raise
```

StefanRijnhart: Yes, agree with that.

What I mean with my first sentence, is that if `error.pgcode == UNDEFINED_COLUMN` is true, then `error.pgcode != UNIQUE_VIOLATION` is also true in which case line 60 raises before your code can take effect.

pedrobaeza: I get you. What about now?

StefanRijnhart (approved):

openupgradelib/openupgrade_merge_records.py
StefanRijnhart: Good now, thanks!

PR #251, opened 2021-11-10
[IMP] merge_records: be able to adjust values by sql

MiquelRForgeFlow:
And some other minor fixes.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -300,6 +300,103 @@ def _change_translations_sql(env, model_name, record_ids, target_record_id,
         })
 
 
+def apply_operations_by_field_type(
+        env, model_name, record_ids, target_record_id, field_spec, field_vals,
+        field_type, column, operation, method):
+    vals = {}
+    o2m_changes = 0
+    if method == 'orm':
+        model = env[model_name]
+        all_records = model.browse((target_record_id,) + tuple(record_ids))
+        target_record = model.browse(target_record_id)
+        first_value = getattr(target_record, column)
+        field = model._fields[column]
+    else:
+        first_value = field_vals[0]
+    if field_type in ('char', 'text', 'html'):
+        if not operation:
+            operation = 'other' if field_type == 'char' else 'merge'
+        if operation == 'first_not_null':
+            field_vals = [x for x in field_vals if x]
+            if field_vals:
+                vals[column] = field_vals[0]
+        elif operation == 'merge':
+            _list = filter(lambda x: x is not False, field_vals)
+            vals[column] = ' | '.join(_list)
+    elif field_type in ('integer', 'float', 'monetary'):
+        if not operation:
+            operation = 'other' if field_type == 'integer' else 'sum'
+        if operation == 'sum':
+            vals[column] = sum(field_vals)
+        elif operation == 'avg':
+            vals[column] = sum(field_vals) / len(field_vals)
+        elif operation == 'max':
+            vals[column] = max(field_vals)
+        elif operation == 'min':
+            vals[column] = min(field_vals)
+    elif field_type == 'boolean':
+        operation = operation or 'other'
+        if operation == 'and':
+            vals[column] = functools.reduce(lambda x, y: x & y, field_vals)
+        elif operation == 'or':
+            vals[column] = functools.reduce(lambda x, y: x | y, field_vals)
+    elif field_type in ('date', 'datetime'):
+        if operation:
+            field_vals = filter(lambda x: x is not False, field_vals)
+        operation = field_vals and operation or 'other'
+        if operation == 'max':
+            vals[column] = max(field_vals)
+        elif operation == 'min':
+            vals[column] = min(field_vals)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
You are still grouping several things with controversial ones included. Please split them for having the most obvious ones immediately merged. For the biggest one, this should be now opted-in, as existing code in OU relies on SQL merging without the values adjustment.

MiquelRForgeFlow:
De-grouped.

> this should be now opted-in, as existing code in OU relies on SQL merging without the values adjustment.

I have done:
Selection_033

So if you don't have any field_spec but you want to adjust values, you can pass `field_spec = {}` to do the trick.

pedrobaeza:
OK, seems a good compromise.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -612,7 +717,9 @@ def merge_records(env, model_name, record_ids, target_record_id,
         _change_foreign_key_refs(*args3)
         _change_reference_refs_sql(*args)
         _change_translations_sql(*args)
-        # TODO: Adjust values of the merged records through SQL
+        args4 = args3 + (field_spec,)
```
pedrobaeza, line 720: Do this inside the `if`

openupgradelib/openupgrade_merge_records.py
```diff
@@ -431,6 +431,111 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
+            op = op or 'merge'
+            if op == 'merge':
+                first_value = _list[0]
+                if _list and not first_value:
+                    vals[column] = _list[0]
+        # Curate values that haven't changed
+        if vals[column] != _list[0]:
+            new_vals[column] = vals[column]
+    if new_vals:
+        logged_query(
```
pedrobaeza, line 521: Not sure if this query should be logged, and if so, then don't double log after in debug mode. Isn't the exactly same detail?

openupgradelib/openupgrade_merge_records.py
```diff
@@ -431,6 +431,111 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
+        """, (model_table, model_name))
+    vals = {}
+    new_vals = {}
+    for column, column_type, field_type in env.cr.fetchall():
+        if (model_table, column) in exclude_columns:
+            continue
+        op = field_spec.get(column, False)
+        env.cr.execute(
+            sql.SQL(
+                """SELECT {column}
```
pedrobaeza, line 464: You are doing one SELECT per column. This is not performance-wise. Read all the columns in one shot.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -431,6 +431,111 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
+      to perform as values. If a field is not present here, default operation
+      will be performed.
+
+      Possible operations by field types same as _adjust_merged_values_orm.
+    """
+    if not column_exists(env.cr, model_table, 'id'):
+        return
+    env.cr.execute(
+        """
+        SELECT isc.column_name, isc.data_type, imf.ttype
```
pedrobaeza, line 450: You should get fields from `ir_model_fields`, not from here, as the existing columns may not be the real fields one (you can have more and perform operations in then incorrectly). The contrary may happen the same (non stored fields, m2m/o2m fields), but you should intersect both for having the proper one, and add the TODO to handle here m2m and o2m fields.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -431,6 +431,111 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
+                """SELECT {column}
+                FROM {table}
+                WHERE id IN %(record_ids)s"""
+            ).format(
+                table=sql.Identifier(model_table),
+                column=sql.Identifier(column),
+            ), {'record_ids': (target_record_id,) + tuple(record_ids)}
+        )
+        _list = [x[0] for x in env.cr.fetchall()]
+        if field_type in ('char', 'text', 'html'):
```
pedrobaeza, line 473: As this code is shared between both methods, I think you can extract it and pass by parameters the needed values to be handled, for saving code.

pedrobaeza (changes requested):

openupgradelib/openupgrade_merge_records.py
MiquelRForgeFlow: But there is a JOIN with ir_model_fields two lines below 🤔

pedrobaeza: OK, right, then please add the TODO for the m2m/o2m.

MiquelRForgeFlow:
Selection_088

MiquelRForgeFlow:
@StefanRijnhart any idea how to solve this package error?

StefanRijnhart:
https://github.com/odoo/odoo/issues/76144 suggests setuptools<58.0. Would that be an option?

pedrobaeza (approved):

PR #252, opened 2021-11-10
[FIX] _adjust_merged_values_orm: get target_record_id value first

MiquelRForgeFlow:
This way we avoid calling getattr().


pedrobaeza:
AFAIK, `getattr` was for avoiding incorrect or half-loaded fields that can happen on migration, so it's still needed.

MiquelRForgeFlow:
Ah, ok, I don't really get it, but I believe in you. Then should I close this PR?

pedrobaeza:
Well, there's a part that I like: put first the target record id, but I don't remember why I didn't do it that way, and if first not null exists already as operation and is not taking target record first.

pedrobaeza (approved):
It makes sense for first not null operation to get the target value first

PR #253, opened 2021-11-10
[IMP] *: adapt to Many2manyReference fields

MiquelRForgeFlow:
It was time.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -346,6 +346,8 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
       * One2many fields:
         - 'merge' (default): combine all the values
         - other value: content on target record is preserved
+      * Many2manyReference fields:
+        - any value: content on target record is preserved
       * Reference fields:
         - any value: content on target record is preserved
       * Selection fields:
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2291,9 +2291,9 @@ def add_fields(env, field_spec):
       * SQL table name: Put `False` if the model is already loaded in the
         registry and thus the SQL table name can be obtained that way.
       * field type: binary, boolean, char, date, datetime, float, html,
-        integer, many2many, many2one, monetary, one2many, reference,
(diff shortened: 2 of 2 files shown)


openupgradelib/openupgrade_merge_records.py
```diff
@@ -346,6 +346,8 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
       * One2many fields:
         - 'merge' (default): combine all the values
         - other value: content on target record is preserved
+      * Many2manyReference fields:
+        - any value: content on target record is preserved
```
pedrobaeza, line 350: Why not adding the same philosophy as `Many2one` (first not null)?

MiquelRForgeFlow: Because the same reason we don't do that for `reference` fields. I mean, this kind of field depends on another field (a char) by model_field parameter. Suppose you have:

 - target: M2OR field A = NULL, char field B = NULL
 - 2n record: M2OR field A = some_id, char field B = some_model
 
 If applied first not null, you will get:
- target: M2OR field A = some_id, char field B = NULL
 
which would be wrong.

MiquelRForgeFlow: I don't see how. I mean, suppose you have:

- target: M2OR field A = NULL, char field B = NULL
- 2n record: M2OR field A = some_id, char field B = some_model
- 3n record: M2OR field A = some_id_2, char field B = some_model
- 4n record: M2OR field A = some_id_3, char field B = some_model_2

Suppose I apply "first not null" and pass field_spec = {B: 'merge'}. Result:
- target: M2OR field A = some_id, char field B = 'some_model'+'some_model'+'some_model_2'

Still wrong.

MiquelRForgeFlow: If you merge https://github.com/OCA/openupgradelib/pull/254, I could make us of it to do want you hope.

pedrobaeza (approved):
Let's merge it for now as is.

MiquelRForgeFlow:
No, I put the "work in progress" tag to avoid the merge!

pedrobaeza:
You can do the other work in a later pull request.

PR #254, opened 2021-11-10
[IMP] _adjust_merged_values_orm: add first_not_null option for chars

MiquelRForgeFlow:
:)

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -315,6 +315,7 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
       * Char, Text and Html fields:
         - 'merge' (default for Text and Html): content is concatenated
           with an ' | ' as separator
+        - 'first_not_null': Put first not null value.
         - other value (default for Char): content on target record is preserved
       * Integer, Float and Monetary fields:
         - 'sum' (default for Float and Monetary): Sum all the values of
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_merge_records.py
```diff
@@ -365,6 +366,10 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
         if field.type in ('char', 'text', 'html'):
             if not op:
                 op = 'other' if field.type == 'char' else 'merge'
+            if op == 'first_not_null':
+                _list = filter(lambda x: x is not False, _list)
```
pedrobaeza, line 370: Is it possible to have `""` instead of False?

MiquelRForgeFlow: Yeah, `""` cases are excluded now.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -365,6 +366,10 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
         if field.type in ('char', 'text', 'html'):
             if not op:
                 op = 'other' if field.type == 'char' else 'merge'
+            if op == 'first_not_null':
+                _list = filter(lambda x: x, _list)
+                if _list:
+                    vals[field.name] = _list[0]
             if op == 'merge':
```
pedrobaeza, line 373: Just for a little code performance optimization
```suggestion
            elif op == 'merge':
```

MiquelRForgeFlow: done

pedrobaeza (approved):

PR #256, opened 2021-11-17
[IMP] merge_records: add 'merge' option for many2one_reference

MiquelRForgeFlow:
Firstly I thought on using new 'first_not_null' option in chars for its model_field, but later I understood that both fields are linked and thus I used the actual approach.

It's a bit corner case, and few models have many2one_reference for now, but maybe they increase in the future.

In v14, they are used in `ir.model.data`, `ir.attachment` and several mail models.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -348,7 +348,11 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
         - 'merge' (default): combine all the values
         - other value: content on target record is preserved
       * Many2manyReference fields:
-        - any value: content on target record is preserved
+        - 'merge' (default): if its model_field is in field_spec,
+        delete it from there. Apply first positive (on field and
+        corresponding model_field) of the records if value of target record
+        is not positive, preserve target value otherwise.
+        - other value: content on target record is preserved
       * Reference fields:
         - any value: content on target record is preserved
       * Selection fields:
@@ -421,6 +425,19 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
             if op == 'merge':
                 if not getattr(target_record, field.name) and _list:
                     vals[field.name] = _list[0]
+        elif field.type == 'many2one_reference':
+            if field.model_field not in model._fields:
+                continue
+            op = op or 'merge'
+            if op == 'merge':
+                if field.model_field in field_spec:
+                    del field_spec[field.model_field]
+                list_model_field = all_records.mapped(field.model_field)
+                zip_list = [(x, y) for x, y
+                            in zip(_list, list_model_field) if x and y]
+                if not getattr(target_record, field.name) and zip_list:
+                    vals[field.name] = zip_list[0][0]
+                    vals[field.model_field] = zip_list[0][1]
     # Curate values that haven't changed
     new_vals = {}
     for f in vals:


openupgradelib/openupgrade_merge_records.py
```diff
@@ -421,6 +425,23 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
             if op == 'merge':
                 if not getattr(target_record, field.name) and _list:
                     vals[field.name] = _list[0]
+        elif field.type == 'many2one_reference':
+            if field.model_field not in model._fields:
+                continue
+            model_field = [x for x in fields if x.name == field.model_field][0]
```
pedrobaeza, line 431: No need to twist the thing:
```suggestion
            model_field = model._fields[field.model_field]
```

openupgradelib/openupgrade_merge_records.py
```diff
@@ -421,6 +425,23 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
+            op = op or 'merge'
+            if op == 'merge':
+                if field.model_field not in field_spec:
```
pedrobaeza, line 434: It shouldn't matter if it's on field_spec or not, we should perform the operation IMO. In fact, we must remove the entry from `field_spec` if exists.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -421,6 +425,23 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
+                    list_model_field = all_records.mapped(field.model_field)
+                    zip_list = [(x, y) for x, y
+                                in zip(_list, list_model_field) if x and y]
+                    if not (
+                        getattr(target_record, field.name)
+                        and getattr(target_record, field.model_field)
```
pedrobaeza, line 440: We should remove this line IMO, as we are artificially restricting the update.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -421,6 +425,23 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
+                    ) and zip_list:
+                        vals[field.name] = zip_list[0][0]
+                        vals[field.model_field] = zip_list[0][1]
+                        fields.remove(model_field)
```
pedrobaeza, line 444: Why removing this?

openupgradelib/openupgrade_merge_records.py
MiquelRForgeFlow: Then we need https://github.com/OCA/openupgradelib/pull/252.

MiquelRForgeFlow: The same you could say about the `gettattr` of line https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade_merge_records.py#L422, that also could be removed, right?

pedrobaeza: It's not the same. I'm talking about removing this specific line, not the previous one. The first check is for not overwriting if there's already a value set. This second one is for forcing to have a value in the other field.

pedrobaeza: Anyways, this is not critical of course. Both fields should be filled in normal conditions and not filled on contrary. To have one filled and the other not is the anomaly.

MiquelRForgeFlow: Imagine you have target record with m2o_r = 0, model = 'ir.attachment', record 2 with m2o_r = 0, model = 'ir.attachment', and record 3 with m2o_r = 1, model = 'ir.attachment'. You want target_record having m2o_r = 1, model = 'ir.attachment'. That's why we need double check.

MiquelRForgeFlow: Yeah, you are right. Done.

MiquelRForgeFlow:
Ready

pedrobaeza (approved):

PR #257, opened 2021-11-17
[IMP] merge_records: add 'merge' option for Reference fields

MiquelRForgeFlow:
In v14, models with Reference fields are:

ir.server.object.lines
ir.ui.menu
rating.rating
theme.ir.ui.view

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -354,7 +354,9 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
         is not positive, preserve target value otherwise.
         - other value: content on target record is preserved
       * Reference fields:
-        - any value: content on target record is preserved
+        - 'merge' (default): apply first not null value of the records if
+        value of target record is null, preserve target value otherwise.
+        - other value: content on target record is preserved
       * Selection fields:
         - any value: content on target record is preserved
     """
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_merge_records.py
```diff
@@ -368,7 +370,12 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
         if not field.store or field.compute or field.related:
             continue  # don't do anything on these cases
         op = field_spec.get(field.name, False)
-        _list = all_records.mapped(field.name)
+        try:
```
pedrobaeza, line 373: You should make the try then only for that case, but not in general, or you will be masking too much.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -438,6 +445,11 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
                 if not getattr(target_record, field.name) and zip_list:
                     vals[field.name] = zip_list[0][0]
                     vals[field.model_field] = zip_list[0][1]
+        elif field.type == 'reference':
```
pedrobaeza, line 448: This is the exact same code as `many2one`, so include both in the same if:
`elif field.type in ('many2one', 'reference'):`

MiquelRForgeFlow:
Ready

openupgradelib/openupgrade_merge_records.py
```diff
@@ -368,7 +370,15 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
...
-        _list = all_records.mapped(field.name)
+        if field.type != 'reference':
+            _list = all_records.mapped(field.name)
+        else:
+            try:
+                _list = all_records.mapped(field.name)
+            except Exception:
+                # if two Reference fields records have distinct selection
+                # option, then it returns mixing apples and oranges TypeError
+                continue
```
pedrobaeza, line 381: I'm thinking that you should build a list instead:
```suggestion
            _list = [x[field.name] for x in all_records]
```

pedrobaeza (approved):

PR #261, opened 2021-12-03
[FIX] _adjust_merged_values_sql: error in query parameter order

MiquelRForgeFlow:
Stupid error.

Fast-tracking.


AaronHForgeFlow (approved):

legalsylvain:
please wait CI before merging.

MiquelRForgeFlow:
Yeah, but the CI doesn't help in this case because that method is not tested yet anywhere.

legalsylvain:
I don't know, but the CI was red, when merging, that is not, imo a good thing to do.

PR #266, opened 2021-12-13
[IMP] delete_records_safely_by_xml_id: restrict Exception

MiquelRForgeFlow:
This way, we can detect cases where the method is called wrong, for example passing the cr instead of the env.


pedrobaeza (approved):

MiquelRForgeFlow:
Not sure if ValidationError is needed there, but I prefer be safer.

PR #269, opened 2021-12-16
[FIX] merge_records: improve _change_reference_refs_sql

MiquelRForgeFlow:
This way, we can do merge_records of type sql in pre-migration safely.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -176,11 +176,11 @@ def _change_reference_refs_sql(env, model_name, record_ids, target_record_id,
     for row in rows:
         try:
             model = env[row[0]]
+            if not model._auto:  # Discard SQL views
+                continue
+            table = model._table
         except KeyError:
-            continue
-        if not model._auto:  # Discard SQL views
-            continue
-        table = model._table
+            table = row[0].replace(".", "_")  # guess table
         if not table_exists(cr, table):
             continue
         column = row[1]


pedrobaeza:
We shouldn't unconditionally replace `.` by `_`, or we will see surprises. Instead, please pass an extra dict {model: table} argument that is tested against when the model can't be loaded.

MiquelRForgeFlow:
But that replace is also used in https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L778. 😂

Also it doesn't makes sense to passing that dict, I mean, I don't expect people using merge_records passing a dict with the models that have reference fields :S

Imagine everywhere where merge records is used (in sql), passing always
```python
ref_dict = {
        'crm.lead': 'crm_lead',
        'ir.actions.server': 'ir_act_server',
        'ir.server.object.lines': 'ir_server_object_lines',
        'ir.property': 'ir_property',
        'ir.ui.menu': 'ir_ui_menu',
        'sms.template.preview': 'sms_template_preview',
        'theme.ir.ui.view': 'theme_ir_ui_view',
    }
```

That's horrible.

pedrobaeza:
If you look at the code of `rename_models`, no SQL operation is involved. The replacement of `.` by `_` is performed for getting the XML-ID identifiers and similar things.

You don't have to pass whole model to table mapping, just those that are being done on pre-migration. Please do it at your return.

MiquelRForgeFlow:
Ok. In fact, this is a remote case and maybe never happen.

PR #270, opened 2021-12-17
[IMP] merge_records: option to exclude all fields not in field_spec

MiquelRForgeFlow:
This way, we can be safe when passing a field_spec, but the model is one that may have many other fields added in custom modules.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -402,6 +402,8 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
     :param: field_spec: Dictionary with field names as keys and forced
       operation to perform as values. If a field is not present here, default
       operation will be performed.
+      Note: If you pass 'openupgrade_other_fields': 'preserve' in the dict,
+      the fields that are not specified in the dict will not be adjusted.
 
       Possible operations by field types:
 
@@ -460,6 +462,9 @@ def _adjust_merged_values_orm(env, model_name, record_ids, target_record_id,
     vals = {}
     o2m_changes = 0
     for field in fields:
+        if field_spec.get('openupgrade_other_fields', '') == 'preserve' \
+                and field.name not in field_spec:
+            continue
         if not field.store or field.compute or field.related:
             continue  # don't do anything on these cases
         op = field_spec.get(field.name, False)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I don't think this is sane. Default merge options are sane enough for the rest of the fields. Unfortunately, there's no good way to inject extra field_spec on merge operation, and it's a destructive operation, but I haven't found so far any problem on the models we have used in OU (stock.quant, product.attribute...)

MiquelRForgeFlow:
This should be used in product.product, which is a model which uses to have new fields in some custom modules. We have clients that add a lot of fields in product.product.

pedrobaeza:
It's the same risky to "preserve" the rest of the fields than applying the defaults. I prefer though if you are going to do such risky operation the second one.

MiquelRForgeFlow:
Then we differ. Imagine you have a custom module which adds a float fields like volume2. Then if we do merge_records by sql with a field_spec, and we don't put that field in the field_spec (i.e., like in https://github.com/OCA/OpenUpgrade/pull/2513), then for that field will apply default operation, which is sum, and then _voilá_, the volume2 for that field have been duplicated or something like that. That doesn't make sense. If we pass `"other_fields": "preserve"`, then we avoid this weird situation.

pedrobaeza:
And you have the contrary case: you have a many2one or a char filled on one of the records to merge, but not on the target record. With such option, you will lose that data. That's why I say it's the same bad one option or the other.

MiquelRForgeFlow:
I think it's much safer to lose that many2one (the char not, because default is do nothing), because hey, you have another product (in case of duplicated products) which don't have that many2one, so is less bad option, as you can later add that relation again. Instead, the value for the float field has been replaced and it's more difficult to restore the previous value, as you may have lost.

MiquelRForgeFlow:
Add also: this PR just add an option. You are free to chose not use that option, but don't block others from choosing to use that option.

pedrobaeza:
Several random things:

- We can choose another defaults. But I still think float must be sum. Your specific use case with the volume is very particular.
- Adding the option here means that is a convenient option, while I don't think it should be.
- Of course, such technique is something that I don't approve on OpenUpgrade.

MiquelRForgeFlow:
> Adding the option here means that is a convenient option, while I don't think it should be.

But that's just your taste, not an argument. I like tools that provide many options, just in case I need them some day.

pedrobaeza:
OK, but several things:

1. Document properly such option in the docstring.
2. Use a even weirder key, like `openupgrade_other_fields`, for avoiding collapsing with a real field name.

pedrobaeza (approved):

PR #271, opened 2021-12-18
[FIX] openupgrade_merge_records: Ensure a tuple is passed on SQL queries

pedrobaeza:
psycopg2 converts lists to `ARRAY` when transposing params in the query, so we have to make sure that the iterable passed on our methods are tuples for a proper conversion.

@Tecnativa


legalsylvain (approved):
Makes senses.

/ocabot merge nobump

PR #273, opened 2022-01-18
[IMP] update_module_names: set noupdate = FALSE in remaining occurrences

MiquelRForgeFlow:
There are models, like `ir.rule`, where its records usually are `noupdate = True`, and they may affect the databases if they are left in the database.

For example, this issue happens in stock (v11) because there are two `ir.rule`, one in 'procurement' module and the other in 'stock' module, that both have the same xmlid name (product_pulled_flow_comp_rule), and in v11, procurement is merged in stock.


pedrobaeza (approved):
Discussed internally and it totally makes sense

PR #274, opened 2022-01-18
[IMP] *: warn -> warning

MiquelRForgeFlow:
Logger.warn() is deprecated in last python versions, and we should try to avoid filling logs with useless DeprecationWarning's.

Selection_154


legalsylvain (approved):
thnaks.

pedrobaeza:
But is this supported in all expected Python versions? That's from 2.7

pedrobaeza:
OK then:

/ocabot merge nobump

PR #275, opened 2022-02-02
[FIX] rename_models: also rename models in res_id of ir.property

MiquelRForgeFlow:
Until today it didn't matter (and still doesn't matter, but maybe in the future, who knows...) because the usual models in that column are product.template, product.category, res.partner, product.product, ...

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -825,6 +825,16 @@ def rename_models(cr, model_spec):
             "UPDATE ir_filters SET model_id = %s "
             "WHERE model_id = %s", (new, old,),
         )
+        logged_query(
+            cr, """
+            UPDATE ir_property
+            SET res_id = replace(res_id, %(old_string)s, %(new_string)s)
+            WHERE res_id like %(old_pattern)s""", {
+                "old_pattern": "%s,%%" % old,
+                "old_string": "%s," % old,
+                "new_string": "%s," % new,
+            },
+        )
         # Handle properties that reference to this model
         logged_query(
             cr,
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -825,6 +825,17 @@ def rename_models(cr, model_spec):
...
             "WHERE model_id = %s", (new, old,),
         )
+        logged_query(
+            cr, """
+            UPDATE ir_property
+            SET res_id = regexp_replace(
+                res_id, %(old_pattern)s, %(new_pattern)s
+            )
+            WHERE res_id ~ %(old_pattern)s""", {
+                'old_pattern': r"^%s,[ ]*([0-9]*)" % old,
```
pedrobaeza, line 835: Why the `[ ]*`? It shouldn't be any space in between according ORM

Anyways, I think you don't need regular expressions, as the `WHERE... like` avoids the possibility of matching a substring instead the beginning of the field:
```python
logged_query(
    env.cr,
    """UPDATE ir_property
    SET res_id = replace(res_id, %(old_string)s, %(new_string)s)
    WHERE res_id like %(old_pattern)s""",
    {
        "old_pattern": "%s,%%" % old,
        "old_string": "%s," % old,
        "new_string": "%s," % new,
    },
)
```

pedrobaeza: Well, I think we can improve both :wink: (in separate commits). I suppose using regexp will be always slower.

MiquelRForgeFlow: Done!

pedrobaeza (approved):
Thanks for fine-tuning the function

PR #277, opened 2022-02-08
[IMP] upgrade_path: support upgrade paths (separated by commas)

royleviindoo:
Odoo CE has support the list of upgrade paths, we should be support it to PATH_TO_openupgrade_scripts_MODULE of any repository does not need to contain source code


royleviindoo:
@pedrobaeza what do you think about this PR?

pedrobaeza (approved):

Issue #278, opened 2022-02-10
Extend function "delete_records_safely_by_xml_id" to delete also child records


baguenth:
When applying function "delete_records_safely_by_xml_id" on views which are inherited by other views, we currently receive the following constraint error which lets the migration fail.

OpenUpgrade: Deleting record for XML-ID module.view 
OpenUpgrade: Deleting record ir.ui.view#1234
odoo.sql_db: bad query: DELETE FROM ir_ui_view WHERE id IN (1234)ERROR: update or delete on table "ir_ui_view" violates foreign key constraint "ir_ui_view_inherit_id_fkey" on table "ir_ui_view"DETAIL:  Key (id)=(1234) is still referenced from table "ir_ui_view". 
Error deleting XML-ID module.view : ForeignKeyViolation('update or delete on table "ir_ui_view" violates foreign key constraint "ir_ui_view_inherit_id_fkey" on table "ir_ui_view"\nDETAIL:  Key (id)=(1234) is still referenced from table "ir_ui_view".\n') 
OpenUpgrade: XML-ID module.view changed to noupdate. 

The current workaround is to first delete all the child records, which in fact is not a great solution since it is not really safe in terms of rollback capability etc.:

@openupgrade.migrate()
def migrate(env, version):
    parent_id = env.ref("module.view")
    env["ir.ui.view"].search([("inherit_id", "=", parent_id.id)]).unlink()
    openupgrade.delete_records_safely_by_xml_id(
        env,
        [
            "module.view",
        ],
    )

I understand, that deleting all the child records is may not what everybody wants, therefore it should be optional, e.g. by passing an additional parameter "delete_childs" to "delete_records_safely_by_xml_id".

PR #279, opened 2022-02-10
[IMP] delete_records_safely_by_xml_id: Added functionality to also delete child records

baguenth:
This brings the functionality to be able to also delete child records when using delete_records_safely_by_xml_id. 

The related issue / feature request is: https://github.com/OCA/openupgradelib/issues/278

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2559,7 +2559,7 @@ def safe_unlink(records, do_raise=False):
                         record._name, record.id, repr(e))
 
 
-def delete_records_safely_by_xml_id(env, xml_ids):
+def delete_records_safely_by_xml_id(env, xml_ids, delete_childs=False):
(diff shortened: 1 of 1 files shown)


funbaker:
Such a great PR. It's really annoying trying to migrate a View with inherited Views.

legalsylvain (approved):
Make sense and the implementation will not generate regression.

thanks !

openupgradelib/openupgrade.py
```diff
@@ -2583,7 +2585,12 @@ def delete_records_safely_by_xml_id(env, xml_ids):
             record = env.ref(xml_id, raise_if_not_found=False)
             if not record:
                 continue
-            safe_unlink(record, do_raise=True)
+            if delete_childs:
+                child_and_parent_records = env["ir.ui.view"].search(
```
MiquelRForgeFlow, line 2589: It's sad this improvement only applies to views, but ok.

PR #280, opened 2022-02-25
[FIX] disable_invalid_filters: <module 'time' (built-in)> can not be used in evaluation contexts

royleviindoo:
This error will save archive all filters
=> Prefer providing only the items necessary for your intended use as attributes of `odoo.tools.safe_eval`

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2239,13 +2239,17 @@ def disable_invalid_filters(env):
         columns = (
             getattr(model, '_columns', False) or getattr(model, '_fields')
         )
+
+        globaldict = {'uid': env.uid}
+        if version_info[0] < 14:
+            globaldict.update({'time': time})
         # DOMAIN
         try:
             with savepoint(env.cr):
                 # Strange artifact found in a filter
                 domain = f.domain.replace('%%', '%')
                 model.search(
-                    safe_eval(domain, {'time': time, 'uid': env.uid}),
+                    safe_eval(domain, globaldict),
                     limit=1,
                 )
         except Exception:
@@ -2256,7 +2260,7 @@ def disable_invalid_filters(env):
             continue
         # CONTEXT GROUP BY
         try:
-            context = safe_eval(f.context, {'time': time, 'uid': env.uid})
+            context = safe_eval(f.context, globaldict)
             assert(isinstance(context, dict))
         except Exception:
             logger.warning(


pedrobaeza:
This has been working perfectly during several versions. Is it a problem in v14? Why? If so, you must restrict it to such version, not on the rest.

royleviindoo:
> This has been working perfectly during several versions. Is it a problem in v14? Why? If so, you must restrict it to such version, not on the rest.

Yes, it is a problem in v14 (all filters is archived)
How to restrict it to such version?

pedrobaeza:
But I don't get why. Having an extra argument shouldn't be the reason for that. I have migrated to v14 already some DBs and don't have this problem.

royleviindoo:
> But I don't get why. Having an extra argument shouldn't be the reason for that. I have migrated to v14 already some DBs and don't have this problem.

Because if error occus, except clause is handled, so all filters is archived
You are checking all filters in db after migrated? You can check at menu Settings / Technical / User-defined Filters

royleviindoo:
> But I don't get why. Having an extra argument shouldn't be the reason for that. I have migrated to v14 already some DBs and don't have this problem.
> 

The following code generates that message, however, I see no need to use `time` in the safe_eval

https://github.com/odoo/odoo/blob/14.0/odoo/tools/safe_eval.py#L366-L372

royleviindoo:
@pedrobaeza I fixed to restrict it to version 14.0, please review, thank you

openupgradelib/openupgrade.py
```diff
@@ -2239,13 +2239,17 @@ def format_message(f):
         columns = (
             getattr(model, '_columns', False) or getattr(model, '_fields')
         )
+
+        localdict = {'uid': env.uid}
```
MiquelRForgeFlow, line 2243:
```suggestion
        globaldict = {'uid': env.uid}
```
Selection_245

royleviindoo: @MiquelRForgeFlow I updated, thank you

MiquelRForgeFlow (approved):
LGTM

royleviindoo:
@pedrobaeza please review this PR, thank you

all-dm:
We're migrating a db from v12 to v14 and found this problem too. It's something new in v14 that I didn't see before. Because of this we're getting around 408 user-defined filters archived, so would be great when this pull request can be reviewed and merged.

pedrobaeza (approved):
Checked that no other replacement is found, so good to go

PR #282, opened 2022-03-09
[FIX] rename_models: bug in mail template query

MiquelRForgeFlow:
Little bug.


legalsylvain (approved):
thanks.

legalsylvain:
fast tracking trivial patch.

PR #283, opened 2022-03-09
[ADD] merge_models method

MiquelRForgeFlow:
Based on https://github.com/OCA/OpenUpgrade/pull/3173.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -145,6 +145,7 @@ __all__ = [
     'rename_fields',
     'rename_tables',
     'rename_models',
+    'merge_models',
     'rename_xmlids',
     'add_xmlid',
     'chunked',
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -917,6 +918,104 @@ def rename_models(cr, model_spec):
     # TODO: signal where the model occurs in references to ir_model
 
 
+def merge_models(cr, old_model, new_model, ref_field):
+    """
+    Update model references for models that have merged to an existing model.
+    :param old_model: old model
+    :param new_model: destination model
+    :param ref_field: name of the field in new model that references old model.
```
StefanRijnhart, line 926: I would prefer if this mentioned that it refers to the `id` in the old model's table (if that is indeed the case).

openupgradelib/openupgrade.py
```diff
@@ -917,6 +918,104 @@ def rename_models(cr, model_spec):
...
+            ('mail_message_subtype', 'res_model', '', ''),
+            ('mail_activity', 'res_model', 'res_id', 'res_model_id'),
+            ('mail_activity_type', '', '', 'res_model_id'),
+            ('mail_template', 'model', '', 'model_id'),
+            ('mail_alias', '', '', 'alias_model_id'),
+            ('mail_alias', '', 'alias_parent_thread_id',
+             'alias_parent_model_id'),
+            # mail_followers: special case handled below
+        ]
+    for rename in renames:
```
StefanRijnhart, line 951: I think the code would be more clear if you replaced `rename` with a tuple, perhaps something like
```
for (table, model_name_column, res_id_column, model_id_column) in renames:
```

openupgradelib/openupgrade.py
```diff
@@ -917,6 +918,104 @@ def rename_models(cr, model_spec):
...
+            UPDATE {table} t
+            SET """ + query_1a + query_3 + query_2a + """
+            WHERE """ + query_1b + query_2b
+        logged_query(cr, sql.SQL(query).format(
+            model_table=sql.Identifier(table),
+            table=sql.Identifier(rename[0]),
+            field1=sql.Identifier(rename[1]),
+            field2=sql.Identifier(rename[2]),
+            field3=sql.Identifier(rename[3]),
+            field=sql.Identifier(ref_field)
```
StefanRijnhart, line 990: Here too, the query would be more clear to me I think if the `field1`, `field2`(etc) keys were renamed to something like `model_name_column` etc. What do you think?

openupgradelib/openupgrade.py
MiquelRForgeFlow: Ok, I like your suggestion. I will proceed with it.

MiquelRForgeFlow:
@StefanRijnhart done

leemannd (approved):

StephaneMangin (approved):

sebalix (approved):

PR #284, opened 2022-03-09
[IMP] rename_fields: consider mail templates

MiquelRForgeFlow:
Some clients have several mail templates without corresponding ir.model.data, so these ones are a pain to fix manually after migration. Let's handle it automatically.


pedrobaeza:
I don't have right now a proposition.

PR #285, opened 2022-03-16
[IMP] delete_record_translations: consider `__export__` translations

MiquelRForgeFlow:
- Independently of the module parameter, we should always consider `__export__` translations.
 - If module parameter is `__export__`, then we should also consider null module translations.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2210,18 +2210,24 @@ def delete_record_translations(cr, module, xml_ids):
     """
     if not isinstance(xml_ids, (list, tuple)):
         do_raise("XML IDs %s must be a tuple or list!" % (xml_ids))
-
+    if not isinstance(xml_ids, (unicode, str)):
+        do_raise("module parameter is required!")
+    module_condition = "(module = '__export__' OR module IS NULL)"
+    if not module:
+        module = "__export__"
+    elif module != "__export__":
+        module_condition = "(module = '__export__' OR module = '%s')" % module
     cr.execute("""
         SELECT model, res_id
         FROM ir_model_data
         WHERE module = %s AND name in %s
     """, (module, tuple(xml_ids),))
     for row in cr.fetchall():
-        query = ("""
+        query = """
             DELETE FROM ir_translation
-            WHERE module = %s AND name LIKE %s AND res_id = %s;
-        """)
-        logged_query(cr, query, (module, row[0] + ',%', row[1],))
+            WHERE {} AND name LIKE %s AND res_id = %s;
+        """.format(module_condition)
+        logged_query(cr, query, (row[0] + ',%', row[1],))
 
 
 def disable_invalid_filters(env):


pedrobaeza:
I don't get the use case.

MiquelRForgeFlow:
The use case is to be able to delete `__export__` translations in any case (because in `__export__` translations, the 'module' of the translation normally don't match with 'module' of the ir.model.data). What explanation do you need? :S

MiquelRForgeFlow:
Selection_269

pedrobaeza:
How have you got to have such translations in your system. I don't have any.

MiquelRForgeFlow:
Good question, I don't know. That's from a client database. They have a lot of `__export__` data in their database and I found they have also export translations. My two cents is that in some point in their history, they imported many data including translations.

pedrobaeza:
But then that shouldn't be part of the framework, but something specific of such database and handle it apart, at least IMO.

MiquelRForgeFlow:
But they imported the data using odoo I suppose. So if odoo lets to happen this, then this should be considered as framework, although being such a minor side case.

pedrobaeza:
And what are the steps to reproduce such behavior? Maybe importing translations as CSV? We need such steps for documenting also that thing.

MiquelRForgeFlow:
hehehe I found it. Check the following steps:

- Find a product template record (one which doesn't have external id nor translation in its name:

( you can do:

```sql
SELECT pt.id, pt.name, pt.company_id, imd.id, imd.module, it.id, it.module, it.lang
FROM product_template pt
LEFT JOIN ir_model_data imd ON (imd.model = 'product.template' and imd.res_id = pt.id)
LEFT JOIN ir_translation it ON (it.name = 'product.template,name' and it.res_id = pt.id)
WHERE imd.id IS NULL AND it.id IS NULL limit 1;
```
to obtain a good product template)
- then go to product template tree view and apply filter in search bar in order for that product to appear in the tree.
- then check the left square box for that product and go to actions -> export.
- when do you export in .csv, an external id (ir.model.data) have been created for that product (with module = `__export__`) (still no translations)
- then go to that product (form view) and click on Edit, and then click in the name box just on the language icon/letter. When click on that, a new tree view appear (in case of multi languages) or a wizard to create translation (in case of mono language).
- In the multi languages case, new translations have been automatically created for that field/record with module = `__export__`. I didn't check the monolanguage case but for sure should be the same, because the module value is taken from the corresponding ir model data.

pedrobaeza:
Thanks for the steps, but I still don't understand how this affects to a call to a method that always clean the translations for an already existing XML-ID that comes from modules. It's impossible to get translations for a record that already has XML-ID with `__export__` module. When you click on Action > Export, such records already contain the "good" XML-IDs.

MiquelRForgeFlow:
> It's impossible to get translations for a record that already has XML-ID with `__export__` module

My previous explanation step by step proved that it's really possible. If the field has `translate=True` (i.e, for example the case for the field name of products), when editing in form view, you can automatically add translations for that record. So if that record has any `__export__` xmlid generated previously by exporting, then the translation will have module = `__export__`.

PR #290, opened 2022-05-03
[IMP][15.0] openupgradelib: Since Odoo 15.0, Environment.manage() is useless

royleviindoo:
Reference:

https://github.com/odoo/odoo/blob/b35f127c86c1d271a76dae7186a96190e8558c73/odoo/api.py#L491


legalsylvain (approved):
LGTM at the first sight. (But not sure to understand what was doing that command). should be replaced ? or it is handled natively ?

pedrobaeza (approved):

PR #294, opened 2022-07-08
[IMP] rename_fields: make a complete renaming of a field with side features: filters

royleviindoo:


openupgradelib/openupgrade.py
```diff
@@ -679,7 +679,26 @@ def rename_fields(env, field_spec, no_deep=False):
...
+                            |'pivot_measures'|'pivot_row_groupby'|'pivot_column_groupby'
+                        ):([^\]]*)"""
+                    r"'%s(:day|:week|:month|:year){0,1}'(.*?\])$$"
+                ) % old_field,
+                'new_pattern': r"$$\1:\2'%s\3'\4$$" % new_field,
+            }, (model, ),
+        )
+        # TODO: Rename when the field is part of a domain in ir_ui_view_custom
+        cr.execute(r"""
+            UPDATE ir_ui_view_custom
```
pedrobaeza, line 692: AFAIK, this can't be done this way, as you don't know for which model the definition is, so a simply renaming by field name can't be done, as you would change an appearance of the same field name in another model.

royleviindoo: You're right, I will do another PR in the future

royleviindoo: Have you seen this PR merged yet?

openupgradelib/openupgrade.py
```diff
@@ -679,12 +679,15 @@ def rename_fields(env, field_spec, no_deep=False):
                 AND context ~ %(old_pattern)s
             """ % {
                 'old_pattern': (
-                    r"$$('group_by'|'col_group_by'):([^\]]*)"
+                    r"""$$('group_by'|'col_group_by'|'graph_groupbys'|'graph_measure'
```
pedrobaeza, line 682: `graph_measure` passes a single field, so it doesn't work in this regexp.

pedrobaeza (changes requested):
May it possible to have spaces between `:` and `[]`?

royleviindoo:
> May it possible to have spaces between `:` and `[]`?

Maybe, I added code to support it

openupgradelib/openupgrade.py
royleviindoo: You' right, I correct it, please re-review

duong77476 (approved):

MiquelRForgeFlow (approved):

pedrobaeza (approved):

PR #295, opened 2022-07-11
[IMP] rename_fields: replaced domain having both style of quotes

royleviindoo:


pedrobaeza:
Previous syntax is correct, and it's more resilient. What is the problem with it?

royleviindoo:
> Previous syntax is correct, and it's more resilient. What is the problem with it?

It doesn't work for me
Screenshot from 2022-07-12 07-59-11

pedrobaeza:
But the problem here is not `$$`, but the quotes. It's `"` instead of `'`. Was such domain manually put? If not, it should be a regexp replace having both style of quotes.

royleviindoo:
> But the problem here is not `$$`, but the quotes. It's `"` instead of `'`. Was such domain manually put? If not, it should be a regexp replace having both style of quotes.

Maybe, I'll check again

royleviindoo:
Please review again, thank you

MiquelRForgeFlow:
error

duong77476 (approved):

MiquelRForgeFlow (approved):

openupgradelib/openupgrade.py
```diff
@@ -657,12 +657,14 @@ def rename_fields(env, field_spec, no_deep=False):
         # TODO: Rename when the field is part of a submodel (ex. m2one.field)
         cr.execute("""
             UPDATE ir_filters
-            SET domain = replace(domain, %(old_pattern)s, %(new_pattern)s)
+            SET domain = regexp_replace(
+                domain, %(old_pattern)s, %(new_pattern)s
```
duong77476, line 661:
```suggestion
                domain, %(old_pattern)s, %(new_pattern)s , 'g'
```
@MiquelRForgeFlow  May be need this cc @royleviindoo

royleviindoo: @duong77476 thank you, I added

openupgradelib/openupgrade.py
```diff
@@ -657,12 +657,14 @@ def rename_fields(env, field_spec, no_deep=False):
         # TODO: Rename when the field is part of a submodel (ex. m2one.field)
         cr.execute("""
             UPDATE ir_filters
-            SET domain = replace(domain, %(old_pattern)s, %(new_pattern)s)
+            SET domain = regexp_replace(
+                domain, %(old_pattern)s, %(new_pattern)s, 'g'
```
pedrobaeza, line 661: Why this extra arg `g`?

duong77476: @pedrobaeza  As far as I know, the flag ‘g’ is used to indicate that replace all the occurrences that appear.
If no arg 'g' like this, it wont work in this case (i have muiltiple word which need to replace)
Screenshot from 2022-07-15 09-40-54

But with arg 'g', it work for me: 
Screenshot from 2022-07-15 09-42-55

pedrobaeza:
OK, thanks for the extra explanations. Then this may be applicable to other regexp replaces? And also the possibility of single or double quotes apply to the context replace as well? A final question: is there the possibility of having spaces in between the tuples of the domain?

pedrobaeza:
Why having #294 as well instead of everything together?

royleviindoo:
> OK, thanks for the extra explanations. Then this may be applicable to other regexp replaces? And also the possibility of single or double quotes apply to the context replace as well? A final question: is there the possibility of having spaces in between the tuples of the domain?

Maybe, but it doesn't matter, since we're replace for "field" in tuples of the domain

pedrobaeza:
OK, let's merge this one for clear the pending things. Please check the rest of the questions on the rest of the code.

duong77476:
> Why having #294 as well instead of everything together?

For my point of view, i prefer to separate things, and isn't it easier for you to review right?

pedrobaeza:
Yeah, no problem at all. It was just that for a moment, I didn't know what I was seeing, hehe. And thanks for all this work :smile:

PR #296, opened 2022-07-12
[FIX] disable_invalid_filters: datetime / context_today / relativedelta is not defined

royleviindoo:
Ref: https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/models/ir_filters.py#L39-L44

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2372,6 +2372,18 @@ def disable_invalid_filters(env):
         globaldict = {'uid': env.uid}
         if version_info[0] < 14:
             globaldict.update({'time': time})
+        if version_info[0] >= 13:
+            try:
+                from odoo.tools.safe_eval import datetime as safe_eval_datetime
+                from odoo.tools.safe_eval import dateutil
+            except ImportError:
+                import datetime as safe_eval_datetime
+                import dateutil
+            globaldict.update({
+                'datetime': safe_eval_datetime,
+                'context_today': safe_eval_datetime.datetime.now,
+                'relativedelta': dateutil.relativedelta.relativedelta,
+            })
         # DOMAIN
         try:
             with savepoint(env.cr):


openupgradelib/openupgrade.py
```diff
@@ -2372,6 +2372,12 @@ def format_message(f):
         globaldict = {'uid': env.uid}
         if version_info[0] < 14:
             globaldict.update({'time': time})
+        elif version_info[0] > 14:
```
pedrobaeza, line 2375:
```suggestion
        if version_info[0] >= 13:
```
https://github.com/odoo/odoo/blob/13.0/odoo/addons/base/models/ir_filters.py#L41

royleviindoo: Okey, done

pedrobaeza: Remove the else part also, as it should apply on v13 as well

royleviindoo: Sorry, I edited

openupgradelib/openupgrade.py
```diff
@@ -2372,6 +2372,15 @@ def format_message(f):
         globaldict = {'uid': env.uid}
         if version_info[0] < 14:
             globaldict.update({'time': time})
+        if version_info[0] >= 13:
```
MiquelRForgeFlow, line 2375:
```suggestion
        if version_info[0] > 12:
```

pedrobaeza: Current one is the same valid, as we are getting the main version only (`version_info[0]`)

pedrobaeza: On that mood, I prefer to see `13`, that is the significant version that contains already the change.

openupgradelib/openupgrade.py
```diff
@@ -2372,6 +2372,15 @@ def format_message(f):
...
+            try:
+                from odoo.tools.safe_eval import datetime as safe_eval_datetime
+            except ImportError:
+                import datetime as safe_eval_datetime
+            globaldict.update({
+                'datetime': safe_eval_datetime,
+                'context_today': safe_eval_datetime.datetime.now,
+            })
```
duong77476, line 2383:
```suggestion
            try:
                from odoo.tools.safe_eval import datetime as safe_eval_datetime
                from odoo.tools.safe_eval import dateutil
            except ImportError:
                import datetime as safe_eval_datetime
                import dateutil
            globaldict.update({
                'datetime': safe_eval_datetime,
                'context_today': safe_eval_datetime.datetime.now,
                'relativedelta': dateutil.relativedelta.relativedelta
            })
``` 
@pedrobaeza I recently check it can have a case when relativedelta also disable too, dont you think?

royleviindoo: I added, thank you

pedrobaeza (approved):

duong77476 (approved):

MiquelRForgeFlow (approved):

Issue #190, opened 2019-11-04
rename_fields: Rename field also in customized views domain


pedrobaeza:
Customized views - `ir.ui.view.custom` - (used on dashboards), can contain a field in the domain that being renamed through the `rename_fields` method, are not touched.

It would be good to have them renamed as well if they are first level domain field at least.

PR #298, opened 2022-07-15
[IMP] rename_field : rename field also on custom view like the one used dashboards

duong77476:
https://github.com/OCA/openupgradelib/issues/190


pedrobaeza:
`ir.ui.view.custom` may not exists for <=v9

duong77476:
I haved test 2 cases where we gonna use rename_fields function
- module mass_mailing :  for the field contact_ab_pc (v14) -> ab_testing_pc (v15)
- module utm : for the field is_website (v14) -> is_auto_campaign (v15)
It seem right to me, here the test video if necessary : https://drive.google.com/file/d/12YuCwWRkmN0BUyR0cOmSZYPB80OmFwUL/view?usp=sharing

MiquelRForgeFlow (approved):

royleviindoo (approved):

openupgradelib/openupgrade.py
```diff
@@ -582,6 +583,65 @@ def rename_columns(cr, column_spec):
                 )
 
 
+def rename_field_on_dashboard(env, model, old_field, new_field):
```
pedrobaeza, line 586: This should be a private method.

pedrobaeza (commented):
I doubt this is going to be performant enough

openupgradelib/openupgrade.py
duong77476: Maybe you 're right

pedrobaeza:
What about performance?

duong77476:
> What about performance?

I have a few changes, if you have time please review

openupgradelib/openupgrade.py
```diff
@@ -582,6 +583,65 @@ def rename_columns(cr, column_spec):
                 )
 
 
+def _rename_field_on_dashboard(env, model, old_field, new_field,
+                               dashboard_view_data):
+    for r in dashboard_view_data:
```
pedrobaeza, line 588: Do the search here and only if the view contains `old_field`.

duong77476: Thanks for the review and suggestion, I'm quite busy at the moment so i will comeback with this ASAP. In the meantime, feel free to add some modification to this PR, i don't mind.

Kind regard,

openupgradelib/openupgrade.py
```diff
@@ -700,6 +764,12 @@ def rename_fields(env, field_spec, no_deep=False):
                     'new_pattern': "$$'%s'$$" % new_field,
                 }, (model, ),
             )
+        # Perform rename field on a custom view used on dashboard
+        # TODO: Rename when the field is part of a submodel (ex. m2one.field)
+        if version_info[0] > 9:
+            _rename_field_on_dashboard(
```
MiquelRForgeFlow, line 770:
```suggestion
        if version_info[0] > 9:
            dashboard_view_data = env["ir.ui.view.custom"].search([])
            _rename_field_on_dashboard(
```
this way you avoid repeating the "if version_info[0] > 9" comparison

pedrobaeza (changes requested):
Pending the scope of the search for better performance

openupgradelib/openupgrade.py
gdgellatly: @pedrobaeza Forgive me, but is all you are suggesting is (in my simplistic ways)

```
if old_field not in r.arch:
    continue
```

pedrobaeza: More or less, for improving drastically the performance. The previous search shouldn't contain non matched results.

openupgradelib/openupgrade.py
```diff
@@ -582,6 +583,65 @@ def rename_columns(cr, column_spec):
                 )
 
 
+def _rename_field_on_dashboard(env, model, old_field, new_field,
+                               dashboard_view_data):
```
gdgellatly, line 587: This function needs a docstring - it is unclear what dashboard_view_data is.

However from an API perspective, my preference would be that this function was

```def _rename_field_on_dashboard(env, field_spec):
    if version_info[0] < 10:
        return
    # Variable used to change field which present in ir.ui.view.custom
    # like the one used on dashboard module
    dashboard_view_data = env["ir.ui.view.custom"].search([])
```

pedrobaeza: `dashboard_view_data` must disappear as argument and do the search for relevant data containing `old_field`.

openupgradelib/openupgrade.py
```diff
@@ -582,6 +583,65 @@ def rename_columns(cr, column_spec):
...
+                arch_context = re.sub(
+                    r"""('group_by'|'col_group_by'|'graph_groupbys'
+                        |'pivot_measures'|'pivot_row_groupby'|'pivot_column_groupby'
+                        ):([\s*][^\]]*)'%s(:day|:week|:month|:year)
+                        {0,1}'(.*?\])"""
+                    % old_field,
+                    r"\1:\2'%s\3'\4" % new_field,
+                    arch_context,
+                )
+
```
pedrobaeza, line 622: Don't add empty lines inside a method.

openupgradelib/openupgrade.py
gdgellatly: Yes sorry, bad markdown by me, I did write that.

openupgradelib/openupgrade.py
duong77476: Done i use `ilike` search for `ir.ui.view.custom` that contain the `old_field` , is that ok? , also i rename the method and handle some of your comment, please review if you have time

duong77476-viindoo:
@pedrobaeza Can this PR merged sir ? you have anything else to review ?

pedrobaeza:
My fear on this is that it affects a lot the performance. Have you performed a migration with and without this and measure the time for both?

duong77476-viindoo:
> My fear on this is that it affects a lot the performance. Have you performed a migration with and without this and measure the time for both?

I did, but i haven't measured only test the script (it still lacking some cases). Next time i will try to measure and let you know

pedrobaeza:
At the end, it's to see: before the change, the migration lasts 20 minutes; after the change, it lasts 21 minutes. That's acceptable. But if you check it and it takes 40 minutes, then the performance should be analyzed.

duong77476-viindoo:
> At the end, it's to see: before the change, the migration lasts 20 minutes; after the change, it lasts 21 minutes. That's acceptable. But if you check it and it takes 40 minutes, then the performance should be analyzed.

Yes, understood, will keep that in mind

PR #299, opened 2022-07-19
[ADD] delete_constraint_safely: method to delete constraints

MiquelRForgeFlow:
To be used on v14 onward.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -179,6 +179,7 @@ __all__ = [
     'disable_invalid_filters',
     'safe_unlink',
     'delete_records_safely_by_xml_id',
+    'delete_sql_constraint_safely',
     'set_xml_ids_noupdate_value',
     'convert_to_company_dependent',
     'cow_templates_mark_if_equal_to_upstream',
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Several comments:

- Isn't the same for previous versions?
- Can we deduce the constraint name from the field instead of having to put the exact name?

MiquelRForgeFlow:
> - Isn't the same for previous versions?

In v13 and previous versions, the constrains don't have an xmlid. So, to make the method compatible with previous versions, I could do:

```python
if version_info[0] > 13:
    delete_records_safely_by_xml_id(
        env, [module + ".constraint_" + table + "_" + name]
    )
```

but, is it worth? Previous versions are already done so I think this method won't be used in them.

> - Can we deduce the constraint name from the field instead of having to put the exact name?

No, the 'name' in the ir.model.constraint model is equivalent with what in this method is `table + "_" + name`. So a 'table' and a 'name' will be introduced as variables one way or the other. My way is a sum and your way would have to do a left trim. Thus, if I have the choice, I prefer the sum. Moreover, my "name" way is straightforward as is the same as it's declared:
Selection_493

pedrobaeza:
Ok, please document the arguments in the docstring with these details for future references.

MiquelRForgeFlow:
Ready

PR #302, opened 2022-08-21
[Add] Multiple insert with logging implemented

Xpshnik:
The logged_query function doesn't particularly fit for cases where the query is a multiple rows insertion statement, as one would have to call it for each row insertion separetely. 

Recently I had to populate a new table based on the rows from already existing table, and such function would come in handy, especially as it takes exactly the type of argument (sequence of tuples) that a fetch after select from another table would return. Pure SQL INSERT INTO SELECT is the perfect solution if you don't have to handle those values before inserting, but you may have to construct a new list of tuples with somehow modified values or additional values that couldn't be added to the tuples via SELECT. So it probably makes sense to have a function like this for such cases.

I used the cr.mogrify() + cr.execute() approach instead of cr.executemany() as the former is generally believed to perform better.


legalsylvain:
@OCA/openupgrade-maintainers an opinion on that proposal ?

hbrunn:
I'd find this more interesting if there was an OCA/Openupgrade migration script using this function. Code looks fine though

PR #304, opened 2022-09-12
[IMP] convert_field_to_html: be ablet to avoid spamming query for each row

MiquelRForgeFlow:
In v15, this method is used several times and some times in tables that usually have many rows.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2178,7 +2178,8 @@ def move_field_m2o(
                 {field_new_model: value})
 
 
-def convert_field_to_html(cr, table, field_name, html_field_name):
+def convert_field_to_html(
+        cr, table, field_name, html_field_name, verbose=True):
     """
     Convert field value to HTML value.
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Uhm, the operation is not the same right now...

You can add a `verbose=True` keyword argument and don't log if verbose=False.

MiquelRForgeFlow:
> Uhm, the operation is not the same right now...

What do you mean?

pedrobaeza:
You are not processing each content and pass `plaintext2html` over it.

MiquelRForgeFlow:
ah, shit. right!

openupgradelib/openupgrade.py
```diff
@@ -2195,12 +2195,18 @@ def convert_field_to_html(cr, table, field_name, html_field_name):
...
-            }, (plaintext2html(row[1]), row[0])
-        )
+        query = "UPDATE %(table)s SET %(field)s = %%s WHERE id = %%s" % {
+            'field': html_field_name,
+            'table': table,
+        }
+        if verbose:
+            logged_query(
+                cr, query, (plaintext2html(row[1]), row[0])
+            )
```
pedrobaeza, line 2205: Everything fits in one line:

```suggestion
            logged_query(cr, query, (plaintext2html(row[1]), row[0]))
```

openupgradelib/openupgrade.py
```diff
@@ -2195,12 +2195,18 @@ def convert_field_to_html(cr, table, field_name, html_field_name):
...
+        else:
+            cr.execute(
+                query, (plaintext2html(row[1]), row[0])
+            )
```
pedrobaeza, line 2209:
```suggestion
            cr.execute(query, (plaintext2html(row[1]), row[0]))
```

pedrobaeza (approved):

PR #305, opened 2022-09-12
[FIX] flake8

MiquelRForgeFlow:
Detected doing another PR.


openupgradelib/openupgrade.py
```diff
@@ -2428,7 +2428,7 @@ def format_message(f):
         # CONTEXT GROUP BY
         try:
             context = safe_eval(f.context, globaldict)
-            assert(isinstance(context, dict))
+            assert isinstance(context, dict)
```
pedrobaeza, line 2431: Is the same in Python 2?

MiquelRForgeFlow: yes

PR #310, opened 2022-10-31
[FIX] adapt common openupgradelib functions to V16. (ir_translation table has been removed)

legalsylvain:
- ``ir_translation`` table doesn't exist anymore in V16.
- a lot of openupgradelib function base queries on that table.

adding some ``if`` in the code.


pedrobaeza (approved):

pedrobaeza:
When we are sure there's no leftovers, I will release a new lib version.

PR #311, opened 2022-11-03
[FIX] added variable definition

cormaza:
I got this error on load_data method

UnboundLocalError: local variable 'fp' referenced before assignment


cormaza:
T

PR #312, opened 2022-11-28
[IMP] rename_models: Avoid mogrify 'IndexError: tuple index out of range' for % LIKE parameter

rousseldenis:


pedrobaeza:
Can you minimize the diff? I suppose you have isort in your pipeline, but here it's not applied. We can include pre-commit stuff though in this library, but it should be in a separate PR.

pedrobaeza:
If we don't put whole pre-commit thing, I wouldn't change anything of that, even in a separate commit here. It should be handled separately.

rousseldenis:
@pedrobaeza Done

pedrobaeza (approved):

PR #317, opened 2023-02-12
[IMP] Use pre-commit through GH workflow for formatting

pedrobaeza:
@Tecnativa


pedrobaeza:
@sbidoul @simahawk @etobella it seems GH workflows are not activated in this repository. And why I don't have access to Settings having it on OpenUpgrade?

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,37 @@
+name: pre-commit
+
+on:
+  pull_request:
+    branches:
+      - "13.0*"
+  push:
+    branches:
+      - "13.0"
+      - "13.0-ocabot-*"
```
legalsylvain, line 10: I think it's master here.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,115 @@
...
+            entry: found a en.po file
+            language: fail
+            files: '[a-zA-Z0-9_]*/i18n/en\.po$'
+    - repo: https://github.com/oca/maintainer-tools
+      rev: 4cd2b852214dead80822e93e6749b16f2785b2fe
+      hooks:
+          # update the NOT INSTALLABLE ADDONS section above
+          - id: oca-update-pre-commit-excluded-addons
+          - id: oca-fix-manifest-website
+            args: ["https://github.com/OCA/l10n-spain"]
```
legalsylvain, line 41: That value dosn't make sense. But in fact, i think that a lot of checks should be removed, because have no sense in openupgradelib library context.

.pylintrc
```diff
@@ -0,0 +1,123 @@
+
```
legalsylvain, line 1: Unncessary file, imo.

openupgradelib/openupgrade_tools.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
legalsylvain, line 1: This will work in python2 context ?

legalsylvain (commented):
Thanks for your work !
Some remarks inline.
Otherwise, lgtm.

.pylintrc
pedrobaeza: Not really. The rest of the checks except the manifest ones are correct for the Python code.

.pre-commit-config.yaml
pedrobaeza: Yes, removed.

.github/workflows/pre-commit.yml
pedrobaeza: Yes.

openupgradelib/openupgrade_tools.py
pedrobaeza: Well, I think there aren't Unicode characters, but I'm looking how to preserve that, as black removes it.

.github/workflows/pre-commit.yml
pedrobaeza: This was the key to execute the GH workflow!

openupgradelib/openupgrade_tools.py
sbidoul: I think you need to pin black to 21.12b0, which is the last version that supports python 2 code.

pedrobaeza: Tried, and it's still removing it. I'm digging more now.

pedrobaeza: Found. It's this one: https://github.com/OCA/openupgradelib/pull/317/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R71-R72

Removing it, it's not done.

.pylintrc
legalsylvain: Right ! Most of check are useless out of an odoo module context, but some are usefull, like sql-injection.

pedrobaeza:
Finally green! Merging and the next step will be to convert Travis tests to GH.

legalsylvain:
Hi @pedrobaeza.
> Finally green! Merging and the next step will be to convert Travis tests to GH.

do you have pending work on that topic ? Tests are run only for odoo from 6.1 to 11.0 that seems quite obsolete.

And travis makes me nervious on that PR https://github.com/OCA/openupgradelib/pull/318. Don't understand why it is failing.

Let me know ! 
(can be a good objective for a spanish code sprint workshop ;-))

PR #318, opened 2023-05-08
[FIX+ADD] openupgradelib documentation 

legalsylvain:
Ref : https://github.com/OCA/OpenUpgrade/issues/3862

- Add github workflow to check and publish automatically openupgradelib documentation on github page : on https://oca.github.io/openupgradelib/
- use the same way as for openupgrade documentation. (docs / docsource folder)
- include new API page to describe all the openupgradelib function. (copy from the old API page in the openupgrade documentation)

It replace the previous documentation that has not been updated since 2015. (here : https://openupgradelib.readthedocs.io/en/latest/)

Related to : https://github.com/OCA/OpenUpgrade/pull/3863

**Once accepted, before merge** : 
- [x] remove last test commit "TEMP - To remove". (usefull to generate https://legalsylvain.github.io/openupgradelib/)
- [x] remove last automatic commit "[UPD] HTML documentation", done by github_actions user.

**After merge** : set correctly github page in the setting configuration. (I don't have access but I guess @pedrobaeza / @StefanRijnhart have access)


legalsylvain:
Hi @OCA/openupgrade-maintainers. Someone has an idea about that error ? 

Log : https://app.travis-ci.com/github/OCA/openupgradelib/builds/262883178#L482
Travis line : https://github.com/OCA/openupgradelib/blob/master/.travis.yml#L35

At a first sight, it's look unrelated to the current PR.

hbrunn:
seems indeed unrelated, looks more like one of those libs isn't compatible with py3 and needs to be version pinned?

legalsylvain:
> seems indeed unrelated, looks more like one of those libs isn't compatible with py3 and needs to be version pinned?

yes indeed. I'm not very confortable with the technical ``.travis.yml`` file and I have not a lot of time for the time being to invesetigate.
I opened this issue : https://github.com/OCA/openupgradelib/issues/321
Any help welcome.

Thanks !

pedrobaeza:
I suppose this requires a manual rebase removing the generated docs + the temp commit.

legalsylvain:
Hi. Indeed ! 
- CI is green.
- I removed the temporary commits.

@pedrobaeza : could you merge and set up the repo ? 
thanks !

pedrobaeza:
Which setup I have to do?

legalsylvain:
Hi @pedrobaeza.

enable github pages. (I guess disabled for the time being, but I don't have access to this part).

Go to the repository > settings > Pages

Then : 

image

Once validated, a github action should be executed, and deploy the content of this new folder (https://github.com/OCA/openupgradelib/tree/master/docs) in https://oca.github.io/openupgradelib  

let me know once it is done. 

Thanks !

pedrobaeza:
Ouch, I don't have access to Settings either.

@sbidoul @simahawk can you give us access to that part? @hbrunn @StefanRijnhart maybe you can access such part?

hbrunn:
neither here not for openupgrade, seems they both should be owned by @OCA/openupgrade-maintainers ?

pedrobaeza:
I think so... I hope @OCA/board can fix this.

sbidoul:
Due to the way the GitHub teams are synced from the OCA database, only the OCA representative has access to the admin settings. In this case it is @StefanRijnhart.

sbidoul:
In the meantime I have configure gh pages as requested in https://github.com/OCA/openupgradelib/pull/318#issuecomment-1548445041

StefanRijnhart:
Does anything still need to be done by me? The docs seem to be up, but I don't see any actual method documentation, only the headers: https://oca.github.io/openupgradelib/API.html#general-methods

Relatedly, because of my large periods of absence from community projects maybe someone else wants to become the OCA representative?

PR #322, opened 2023-05-14
[FIX] pin CI to trusty for now

hbrunn:


pedrobaeza:
Maybe we can switch to GH actions? I remember I did a test and didn't work, but I didn't go further. I see this branch with documentation switch to GH actions: https://github.com/Tecnativa/openupgradelib/tree/gh-actions

hbrunn:
that should be the goal indeed, but today I'll be happy to just make it work again

.travis.yml
```diff
@@ -24,7 +25,7 @@ script:
     - pip install -q -r requirements.txt
     - export PYTHONPATH=~/build/OCA/openupgradelib:$PYTHONPATH
     - coverage run setup.py test
-    - python setup.py install
+    - PYTHONDONTWRITEBYTECODE=1 python setup.py install
```
hbrunn, line 28: @legalsylvain I think this should go to the documentation, as without this env var it will try to precompile all .py files, breaking on the py3 ones

legalsylvain: > wherever it fits, probably somewhere you'll describe how to setup an environment to actually run the migration?

Hi. Ok I see. for my part, I never used that configuration. The problem I see is that there are many ways to install Odoo (docker, virtualenv, docker based env like doodba (tecnativa), docky (akretion), etc...) **and so**, many ways to setup an openupgrade environment. I so don't have a clear point of view of how to write the documentation page "how to set up an Openupgrade environment".

I will ping you when I start this page, to define the scope of the documentation.

PR #328, opened 2023-06-07
[FIX] delete_record_translations: adapt to Odoo v16

MiquelRForgeFlow:
Needed.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2585,15 +2585,17 @@ def rename_property(cr, model, old_name, new_name):
     )
 
 
-def delete_record_translations(cr, module, xml_ids):
+def delete_record_translations(cr, module, xml_ids, field_list=None):
     """Cleanup translations of specific records in a module.
 
     :param module: module name
     :param xml_ids: a tuple or list of xml record IDs
+    :param field_list: optional list of field names whose translations will be deleted
     """
     if not isinstance(xml_ids, (list, tuple)):
         do_raise("XML IDs %s must be a tuple or list!" % (xml_ids,))
-
+    if not field_list:
+        field_list = []
     cr.execute(
         """
         SELECT model, res_id
@@ -2609,19 +2611,38 @@ def delete_record_translations(cr, module, xml_ids):
         model = row[0]
         record_id = row[1]
         if version_info[0] < 16:
-            query = """
-                DELETE FROM ir_translation
-                WHERE module = %s AND name LIKE %s AND res_id = %s;
-            """
-            logged_query(
-                cr,
-                query,
-                (
-                    module,
-                    model + ",%",
-                    record_id,
-                ),
-            )
+            if not field_list:
+                query = """
+                    DELETE FROM ir_translation
+                    WHERE module = %s AND name LIKE %s AND res_id = %s;
+                """
+                logged_query(
+                    cr,
+                    query,
+                    (
+                        module,
+                        model + ",%",
+                        record_id,
+                    ),
+                )
+            else:
+                name_comparisons = " OR ".join(
+                    ["name = %s" % (model + "," + field) for field in field_list]
+                )
+                query = """
+                    DELETE FROM ir_translation
+                    WHERE module = %s AND ({name_comparison}) AND res_id = %s;
+                """.format(
+                    name_comparison=name_comparisons
+                )
+                logged_query(
+                    cr,
+                    query,
+                    (
+                        module,
+                        record_id,
+                    ),
+                )
         else:
             table = model.replace(".", "_")
             # we use information_schema to assure the columns exist
@@ -2635,6 +2656,8 @@ def delete_record_translations(cr, module, xml_ids):
                 (model, table),
             )
             list_columns = [x[0] for x in cr.fetchall()]
+            if field_list:
+                list_columns = [x for x in list_columns if x in field_list]
             if not list_columns:
                 continue
             query = """
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2592,7 +2592,7 @@ def delete_record_translations(cr, module, xml_ids):
     :param xml_ids: a tuple or list of xml record IDs
     """
     if not isinstance(xml_ids, (list, tuple)):
-        do_raise("XML IDs %s must be a tuple or list!" % (xml_ids))
+        do_raise("XML IDs %s must be a tuple or list!" % (xml_ids,))
 
     cr.execute(
         """
@@ -2606,19 +2606,71 @@ def delete_record_translations(cr, module, xml_ids):
         ),
     )
     for row in cr.fetchall():
-        query = """
-            DELETE FROM ir_translation
-            WHERE module = %s AND name LIKE %s AND res_id = %s;
-        """
-        logged_query(
-            cr,
-            query,
-            (
-                module,
-                row[0] + ",%",
-                row[1],
-            ),
-        )
+        model = row[0]
+        record_id = row[1]
+        if version_info[0] < 16:
+            query = """
+                DELETE FROM ir_translation
+                WHERE module = %s AND name LIKE %s AND res_id = %s;
+            """
+            logged_query(
+                cr,
+                query,
+                (
+                    module,
+                    model + ",%",
+                    record_id,
+                ),
+            )
+        else:
+            table = model.replace(".", "_")
+            # we use information_schema to assure the columns exist
+            cr.execute(
+                """
+                SELECT isc.column_name
+                FROM information_schema.columns isc
+                JOIN ir_model_fields imf ON (
+                    imf.name = isc.column_name AND imf.model = %s)
+                WHERE isc.table_name = %s AND imf.translate""",
+                (model, table),
+            )
+            list_columns = [x[0] for x in cr.fetchall()]
+            if not list_columns:
+                continue
+            query = """
+                SELECT {checks}
+                FROM {table}
+                WHERE id = %s""".format(
+                table=table,
+                checks=", ".join(
+                    [
+                        "{x} IS NOT NULL AND ({x} ? 'en_US') AND "
+                        "(SELECT count(*) FROM jsonb_object_keys({x})) > 1".format(x=x)
+                        for x in list_columns
+                    ]
+                ),
+            )
+            cr.execute(query, (record_id,))
+            checks = cr.fetchall()[0]
+            for i, (column, check) in enumerate(zip(list_columns, checks)):
+                if not check:
+                    list_columns.pop(i)
+            if not list_columns:
+                continue
+            query = """
+                UPDATE {table}
+                SET ({columns}) = ({values})
+                WHERE id = %s""".format(
+                table=table,
+                columns=", ".join(list_columns),
+                values=", ".join(
+                    [
+                        '\'{{"en_US": {x} -> "en_US"}}\'::jsonb'.format(x=x)
+                        for x in list_columns
+                    ]
+                ),
+            )
+            logged_query(cr, query, (record_id,))
 
 
 # flake8: noqa: C901


openupgradelib/openupgrade.py
```diff
@@ -2606,19 +2606,80 @@ def delete_record_translations(cr, module, xml_ids):
...
+            cr.execute(
+                query,
+                (record_id,),
+            )
+            list_translations = [x[0] for x in cr.fetchall()]
+            new_translations = []
+            for translations in list_translations:
+                field_translations = translations or {}
+                if "en_US" in field_translations:
+                    clean_value = {"en_US": field_translations["en_US"]}
```
MiquelRForgeFlow, line 2661: Not sure if I should conserve the en_US translation or delete it too.

remytms: I seams to me that it should be kept, because leaving a field with `{}` as a value seams to make it disappear in the interface.

I tried setting value `{}` to the name field of a res_country and after that I can no longer find it on the interface. Updating the `en_US` translation via the interface, did not bring back the translation for this res_country.

MiquelRForgeFlow: Thanks!

openupgradelib/openupgrade.py
```diff
@@ -2606,19 +2606,90 @@ def delete_record_translations(cr, module, xml_ids):
...
+                SELECT {columns}
+                FROM {table}
+                WHERE id = %s""".format(
+                table=table,
+                columns=columns,
+            )
+            cr.execute(
+                query,
+                (record_id,),
+            )
```
pedrobaeza, line 2655:
```suggestion
            cr.execute(query, (record_id,))
```

openupgradelib/openupgrade.py
```diff
@@ -2606,19 +2606,90 @@ def delete_record_translations(cr, module, xml_ids):
...
+            list_translations = [x[0] for x in cr.fetchall()]
+            new_translations = []
+            for column, translations in zip(list_columns, list_translations):
+                field_translations = translations or {}
+                if "en_US" in field_translations:
+                    clean_value = {"en_US": field_translations["en_US"]}
+                else:
+                    clean_value = {
+                        "en_US": (
+                            column[:-4]
```
pedrobaeza, line 2665: Don't understand this. Please add comments.

openupgradelib/openupgrade.py
```diff
@@ -2606,19 +2606,90 @@ def delete_record_translations(cr, module, xml_ids):
...
+                        )
+                        .replace("_", " ")
+                        .title()
+                    }
+                new_translations += [Json(clean_value)]
+            query = """
+                UPDATE {table}
+                SET %s = %s
+                WHERE id = %s;
+            """.format(
```
pedrobaeza, line 2679: Save lines:

```suggestion
            query = "UPDATE {table} SET %s = %s WHERE id = %s".format(
```

duong77476:
Nice @MiquelRForgeFlow

openupgradelib/openupgrade.py
```diff
@@ -2606,19 +2606,90 @@ def delete_record_translations(cr, module, xml_ids):
...
+                    record_id,
+                ),
+            )
+        else:
+            from psycopg2.extras import Json
+
+            table = model.replace(".", "_")
+            cr.execute(
+                """
+                SELECT isc.column_name
```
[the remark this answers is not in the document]
MiquelRForgeFlow: It was a way to ensure the column exists...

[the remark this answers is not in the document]
pedrobaeza: OK, please indicate so on a comment before this instruction.

pedrobaeza:
As commented in OU, I'm seeing that this method is removing all translations for all fields, but there's a lot of times where this is not needed, so we should add a granularity for passing (as keyword argument), a field list for resetting translations.

openupgradelib/openupgrade.py
MiquelRForgeFlow: This part was taken from Odoo:

Selection_1472

MiquelRForgeFlow:
@pedrobaeza fields_list optional parameter added.

openupgradelib/openupgrade.py
MiquelRForgeFlow: And comment added.

pedrobaeza:
Last question: what happens in a DB with `en_US` disabled and another single language enabled? Are translations still stored with `en_US` JSON value?

openupgradelib/openupgrade.py
MiquelRForgeFlow: I just deleted this part, as it was not making sense:

Selection_1473

With my code, that empty cells would have been filled with `{"en_US": "Description"}`

MiquelRForgeFlow:
> Are translations still stored with en_US JSON value?

Yes. I just tested it.

pedrobaeza:
I know it's a bit picky, but `fields_list` is a bad variable name, as the `s` is going to be typoed most of the time. Please call `field_list` as normalized way of naming variables.

openupgradelib/openupgrade.py
MiquelRForgeFlow: I think we should set `NULL`, not `{}` in order to avoid it disappearing from the interface. When Odoo detects the null value, it will load translations. (Not tested, but it's my two cents of my logic).

Selection_1473

pedrobaeza: But this is usually called after `load_data` call, so in `en_US` we already have the proper updated value. A doubt I have right now is that that I think we are forcing or not `en_US` in such `load_data` method.

pedrobaeza:
I'm also wondering if we have done it correctly all this time, doing the `delete_record_translations` on post, instead of pre. That way, we require another update for loading the translations. What do you think about this?

MiquelRForgeFlow:
> I'm also wondering if we have done it correctly all this time, doing the delete_record_translations on post, instead of pre. That way, we require another update for loading the translations. What do you think about this?

But the update of the noupdate changed fields is done in post, so if you delete the translations in pre, during the load they will be wrongly set, as they don't have the noupdate changes yet. I mean, to do it in pre, you would also need to apply the noudate changes in pre, which is wrong.

pedrobaeza:
I don't think so, as the search is not done by the source term, but the comment. And now on v16 the rules change. It's not very important, as we all perform a second update after the migration itself for the cleanup, but it we let the system the best on the first round, it would be great.

openupgradelib/openupgrade.py
MiquelRForgeFlow: > But this is usually called after load_data call, so in en_US we already have the proper updated value.

Are you sure?

> A doubt I have right now is that that I think we are forcing or not en_US in such load_data method.

We need to check.

pedrobaeza: That's the purpose of the `load_data` method.

MiquelRForgeFlow:
I delete now all translations using NULL. Let's check in a simple migration if this works or if it messes up.

pedrobaeza:
Well, it's replacing the `load_data` loaded data. And I'm also wondering now how this new JSON fields behaves with noupdate=1 data and translations.

MiquelRForgeFlow:
@pedrobaeza I fully tested it. It's ready and safe.

openupgradelib/openupgrade.py
```diff
@@ -2606,19 +2608,88 @@ def delete_record_translations(cr, module, xml_ids):
...
+            list_translations = cr.fetchall()[0]
+            new_values = []
+            for column, translations in zip(list_columns, list_translations):
+                field_translations = translations or {}
+                if "en_US" in field_translations:
+                    clean_value = {"en_US": field_translations["en_US"]}
+                    new_values += [Json(clean_value)]
+                else:
+                    new_values += [None]
+            query = "UPDATE {table} SET ({columns}) = ({values}) WHERE id = %s".format(
```
pedrobaeza, line 2683: Can we easily avoid to update if no change?

openupgradelib/openupgrade.py
```diff
@@ -2606,19 +2608,88 @@ def delete_record_translations(cr, module, xml_ids):
...
+                table=table,
+                columns=columns,
+            )
+            cr.execute(query, (record_id,))
+            list_translations = cr.fetchall()[0]
+            new_values = []
+            for column, translations in zip(list_columns, list_translations):
+                field_translations = translations or {}
+                if "en_US" in field_translations:
+                    clean_value = {"en_US": field_translations["en_US"]}
```
pedrobaeza, line 2679: Can we also avoid to read each time the translation, but only to remove extra values. I think JSON SQL allows this.

MiquelRForgeFlow: How?

openupgradelib/openupgrade.py
MiquelRForgeFlow: How?

openupgradelib/openupgrade.py
pedrobaeza: And what about to just assign on the UPDATE the existing value? The same you do something like `UPDATE table SET field = field + 1`, you would be able to do something similar with JSON SQL. Invented example: `UPDATE table SET field_json = {"en_US": field_json["en_US"]}`. 

https://www.postgresql.org/docs/current/functions-json.html

openupgradelib/openupgrade.py
MiquelRForgeFlow: Done

openupgradelib/openupgrade.py
MiquelRForgeFlow: This contradicts the previous suggestion. Or I read the translations and filter them to avoid doing stupid updates, or I don't read them and just do a complicated update query that will be executed always.

pedrobaeza: Well, more or less. Maybe the SQL `WHERE` can include also a condition to say "if `en_US` is not the only key in the JSON". But I'm fine-tuning a lot, I know.

pedrobaeza:
I see that you are still doing a select to fetch the column values, and that's exactly what I want to avoid on my concern: only do the `UPDATE` query, handling it through JSON operators in SQL.

openupgradelib/openupgrade.py
MiquelRForgeFlow: Then, a separate UPDATE query for each column will be done. I cannot batch them into one if we need to check for each one if  they include the en_US key or not.

MiquelRForgeFlow:
> handling it through JSON operators in SQL

Done!

pedrobaeza (approved):
OK, this seems good enough for now. Thanks for the efforts.

legalsylvain:
Hi @all thanks for this work. 
The recent merge is breaking openupgrade 16.0 CI. 
See : https://github.com/OCA/OpenUpgrade/actions/runs/5460474242/jobs/9937497789?pr=4042#step:15:5315

an existing script (in auth_signup) is now failing. Could you take a look ? Thanks ! 

```
   File "/home/runner/work/OpenUpgrade/OpenUpgrade/openupgrade/openupgrade_scripts/scripts/auth_signup/16.0.1.0/post-migration.py", line 17, in migrate
    openupgrade.delete_record_translations(
  File "/opt/hostedtoolcache/Python/3.10.12/x64/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 2696, in delete_record_translations
    logged_query(cr, query, (record_id,))
  File "/opt/hostedtoolcache/Python/3.10.12/x64/lib/python3.10/site-packages/openupgradelib/openupgrade.py", line 1681, in logged_query
    cr.execute(query, args)
  File "/home/runner/work/OpenUpgrade/OpenUpgrade/odoo/odoo/sql_db.py", line 321, in execute
psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type json
LINE 3: ...   SET (name, body_html, subject, description) = ('{"en_US":...
                                                             ^
DETAIL:  Token "name" is invalid.
CONTEXT:  JSON data, line 1: {"en_US": name...
```

MiquelRForgeFlow:
@legalsylvain sorry, that breaking line was a last untested change. It is being fixed in https://github.com/OCA/openupgradelib/pull/335.

PR #331, opened 2023-06-09
[IMP] merge_records: support Serialized and 'Jsonb' fields (adapt to Odoo v16)

MiquelRForgeFlow:
I added Serialized fields support as they are very similar to 'Jsonb' fields and could be done in the same operation.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -286,6 +286,8 @@ def _change_reference_refs_orm(
 def _change_translations_orm(
     env, model_name, record_ids, target_record_id, exclude_columns
 ):
+    if version_info[0] > 15:
+        return
     if ("ir_translation", "res_id") in exclude_columns:
         return
     translation_obj = env["ir.translation"]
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_merge_records.py
```diff
@@ -553,18 +582,18 @@ def _adjust_merged_values_orm(
         if not field.store or field.compute or field.related:
             continue  # don't do anything on these cases
         op = field_spec.get(field.name, False)
-        if field.type != "reference":
-            _list = all_records.mapped(field.name)
-        else:
+        if field.type == "reference":
```
pedrobaeza, line 585: This changes is not needed.

MiquelRForgeFlow: I know, it was just changing the order. It bothers you? I think it's better doing this way, because it's more readable listing the specific case in the `if` and the rest in the `else`. If in the future we need to add anther case, a simple `elif` would only be added.

pedrobaeza: Readability is debatable. I think it's the same, but the main principle is to reduce the diff on unrelated things. This is totally unrelated to the goal of the PR. You can do such refactoring in other PR labeling it `[REF]`.

MiquelRForgeFlow: Done

pedrobaeza (approved):

PR #332, opened 2023-06-16
[FIX] merge_records: Adapt to v16

pedrobaeza:
@Tecnativa


MiquelRForgeFlow:
I already included this in https://github.com/OCA/openupgradelib/pull/331.

PR #336, opened 2023-07-05
[ADD] get_model2table: method to map nonstandard table names

MiquelRForgeFlow:
Extracted from https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/base/16.0.1.3/pre-migration.py#L65C9-L65C64, in order to all versions get the benefit.

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -10,7 +10,7 @@ from psycopg2 import IntegrityError, ProgrammingError, sql
 from psycopg2.errorcodes import UNDEFINED_COLUMN, UNIQUE_VIOLATION
 from psycopg2.extensions import AsIs
 
-from .openupgrade import logged_query, version_info
+from .openupgrade import get_model2table, logged_query, version_info
 from .openupgrade_tools import column_exists, table_exists
 
 logger = logging.getLogger("OpenUpgrade")
@@ -216,11 +216,11 @@ def _change_reference_refs_sql(
     for row in rows:
         try:
             model = env[row[0]]
+            if not model._auto:  # Discard SQL views
+                continue
+            table = model._table
         except KeyError:
-            continue
-        if not model._auto:  # Discard SQL views
-            continue
-        table = model._table
+            table = get_model2table(row[0])
         if not table_exists(cr, table):
             continue
         column = row[1]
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Not sure if this may vary across versions.

MiquelRForgeFlow:
> Not sure if this may vary across versions.

 I have checked from v8 to v16 and this is the only change:
Selection_1476

So, I think we are being safe enough. If in the future any of these models/tables changes, then we will adapt this new method accordingly.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -217,10 +217,11 @@ def _change_reference_refs_sql(
         try:
             model = env[row[0]]
         except KeyError:
-            continue
-        if not model._auto:  # Discard SQL views
-            continue
-        table = model._table
+            table = get_model2table(row[0])
+        else:
+            if not model._auto:  # Discard SQL views
```
pedrobaeza, line 222: This is not evaluated in the proper place.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -758,8 +759,10 @@ def _change_generic(
         try:
             model = env[model_to_replace].with_context(active_test=False)
         except KeyError:
-            continue
-        if (model._table, res_id_column) in exclude_columns:
+            table = get_model2table(model_to_replace)
+        else:
+            table = model._table
```
pedrobaeza, line 764: Don't do this in `else`, but next to the `model = env[model_to_replace].with_context(active_test=False)`. The same for all occurrences of this type.

openupgradelib/openupgrade_merge_records.py
```diff
@@ -861,7 +864,10 @@ def _delete_records_sql(
     env, model_name, record_ids, target_record_id, model_table=None
 ):
     if not model_table:
-        model_table = env[model_name]._table
+        try:
```
pedrobaeza, line 867: Not sure about all of this new `try`. If the model is not available, another bigger problem may arise, so I'm thinking that better to detect it sooner than later.

openupgradelib/openupgrade_merge_records.py
MiquelRForgeFlow: What do you mean? The `else` part is executed if no exception is triggered. Thus, the code is fine.

openupgradelib/openupgrade_merge_records.py
MiquelRForgeFlow: This is `_delete_records_sql()`, which used only when method='sql'. My point is we should be able to use the merge records with method='sql' in tables that are obsolete or not currently in the env in the moment the merge is called.

pedrobaeza (approved):

PR #337, opened 2023-07-06
[FIX] delete_record_translations: error when only update for 1 one column

duong77476:


legalsylvain (approved):

MiquelRForgeFlow:
Please, don't duplicate code. Try to do it in only one query. You can calculate columns and values before the query, and then add `()` in case > 1.

duong77476:
> Please, don't duplicate code. Try to do it in only one query. You can calculate columns and values before the query, and then add `()` in case > 1.

Yes, i understand, i'll will update and rerun again

openupgradelib/openupgrade.py
```diff
@@ -2680,18 +2680,27 @@ def delete_record_translations(cr, module, xml_ids, field_list=None):
...
+                    "jsonb_build_object('en_US', {x} -> 'en_US')".format(x=x)
+                    for x in list_columns
+                )
+                columns = "({})".format(columns)
+                values = "({})".format(values)
+            else:
+                columns = list_columns[0]
+                values = "jsonb_build_object('en_US', {x} -> 'en_US')".format(
+                    x=list_columns[0]
+                )
```
MiquelRForgeFlow, line 2697:
```suggestion
            columns = ", ".join(list_columns)
            values = ", ".join(
                "jsonb_build_object('en_US', {x} -> 'en_US')".format(x=x)
                for x in list_columns
            )
            if len(list_columns) > 1:
                columns = "({})".format(columns)
                values = "({})".format(values)
```

duong77476: Thank you, but i prefer keeping my comment is that ok?

duong77476: Got it, done sir

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
Thanks @duong77476

Issue #339, opened 2023-07-17
Error migrating from version 9.0 to version 10.0


zamberjo:
I know that these versions are very old but in order to bump to a more current version you must migrate version by version to the desired version.

## Describe the bug
The error appears after the merge of https://github.com/OCA/openupgradelib/pull/336.

```
2023-07-17 05:18:03,199 129 CRITICAL migration odoo.service.server: Failed to initialize database `migration`.
Traceback (most recent call last):
    File "/opt/odoo/custom/src/repos/odoo/odoo/service/server.py", line 974, in preload_registries
      registry = Registry.new(dbname, update_module=update_module)
    File "/opt/odoo/custom/src/repos/odoo/odoo/modules/registry.py", line 83, in new
      odoo.modules.load_modules(registry._db, force_demo, status, update_module)
    File "/opt/odoo/custom/src/repos/odoo/odoo/modules/loading.py", line 375, in load_modules
      report=report, models_to_check=models_to_check, upg_registry=upg_registry)
    File "/opt/odoo/custom/src/repos/odoo/odoo/modules/loading.py", line 211, in load_module_graph
      migrations.migrate_module(package, 'post')
    File "/opt/odoo/custom/src/repos/odoo/odoo/modules/migration.py", line 178, in migrate_module
      mod.migrate(self.cr, installed_version)
    File "/home/odoo/.local/lib/python2.7/site-packages/openupgradelib/openupgrade.py", line 2275, in wrapped_function
      version,
    File "/opt/odoo/custom/src/repos/odoo/odoo/addons/base/migrations/10.0.1.3/post-migration.py", line 142, in migrate
      merge_country_states(env)
    File "/opt/odoo/custom/src/repos/odoo/odoo/addons/base/migrations/10.0.1.3/post-migration.py", line 73, in merge_country_states
      correct_country_states[-1].id,
    File "/home/odoo/.local/lib/python2.7/site-packages/openupgradelib/openupgrade_merge_records.py", line 1003, in merge_records
      _change_generic(*args, method=method)  # pylint: disable=E1124
    File "/home/odoo/.local/lib/python2.7/site-packages/openupgradelib/openupgrade_merge_records.py", line 766, in _change_generic
      if not model._fields.get(model_column) or not model._fields.get(
    UnboundLocalError: local variable 'model' referenced before assignment
```

MiquelRForgeFlow:
[...]

PR #340, opened 2023-07-17
[FIX] merge_records: assure don't break on method="orm" if KeyError on model

MiquelRForgeFlow:
Fixes https://github.com/OCA/openupgradelib/issues/339

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -759,6 +759,8 @@ def _change_generic(
             model = env[model_to_replace].with_context(active_test=False)
             table = model._table
         except KeyError:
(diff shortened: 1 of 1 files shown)


zamberjo (approved):

pedrobaeza:
See my comment at https://github.com/OCA/openupgradelib/issues/339#issuecomment-1637803600

Probably this problem is because of an unexisting model in such version.

MiquelRForgeFlow:
> Probably this problem is because of an unexisting model in such version.

Yeah, but before https://github.com/OCA/openupgradelib/commit/7819744da0af84cf38ec5629ae91a578aafd70e1, the case was not breaking. Now breaks, so we need to avoid the break.

PR #343, opened 2023-08-28
Implement m2m_to_o2m

carmenbianca:
Ported over from https://github.com/OCA/OpenUpgrade/pull/3911 in the hope that it can be more broadly useful.

Tagging @remytms

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1956,7 +1956,7 @@ def m2m_to_o2m(
     source_relation_table,
     relation_parent_field,
     relation_child_field,
-    strategy: M2mToO2mStrategy,
+    strategy,
 ):
     """Transform many2many relations into one2many (with possible data loss).
     The data loss occurs when two (or more) parent records are linked to one
@@ -2071,7 +2071,6 @@ def m2m_to_o2m(
             )
             target_id = env[columns.comodel_name].browse(child)
             for parent in parents:
-                parent_id = env[model].browse(parent)
                 logger.warning(
                     "Making a copy of %(child_model)s(%(target_id)s,) and"
                     " assigning %(parent_model)s(%(parent_id)s,) to its"
@@ -2084,6 +2083,7 @@ def m2m_to_o2m(
                         "target_field": target_field,
                     },
                 )
+                parent_id = env[model].browse(parent)
                 target_copy = target_id.copy()
                 setattr(target_copy, target_field, parent_id)
 
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -171,6 +171,8 @@ __all__ = [
     "get_legacy_name",
     "get_model2table",
     "m2o_to_x2m",
+    "m2m_to_o2m",
+    "m2m_to_o2m_dataloss_warn",
     "float_to_integer",
     "message",
     "check_values_selection_field",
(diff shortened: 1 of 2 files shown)


openupgradelib/openupgrade.py
```diff
@@ -1935,6 +1937,119 @@ def m2o_to_m2m(cr, model, table, field, source_field):
...
+            "relation_comodel_field": AsIs(relation_comodel_field),
+        },
+    )
+
+
+def m2m_to_o2m_dataloss_warn(
+    cr, source_relation_table, relation_source_field, relation_comodel_field
+):
+    """Warn user about data loss when migrating data from many2many to
+    many2one.
```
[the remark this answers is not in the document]
carmenbianca: Good catch! Fixed.

openupgradelib/openupgrade.py
```diff
@@ -1935,6 +1937,119 @@ def m2o_to_m2m(cr, model, table, field, source_field):
...
+            "target_table": AsIs(target_table),
+            "target_field": AsIs(target_field),
+            "source_relation_table": AsIs(source_relation_table),
+            "relation_source_field": AsIs(relation_source_field),
+            "relation_comodel_field": AsIs(relation_comodel_field),
+        },
+    )
+
+
+def m2m_to_o2m_dataloss_warn(
```
pedrobaeza, line 2003: Instead of 2 methods, put a keyword argument for warning.

pedrobaeza (commented):
Not sure about the use case, but this will probably needs a migration from one record with m2m to several records with m2o.

openupgradelib/openupgrade.py
```diff
@@ -1935,6 +1937,119 @@ def m2o_to_m2m(cr, model, table, field, source_field):
...
+def m2m_to_o2m(
+    env,
+    model,
+    field,
+    source_relation_table,
+    relation_source_field,
+    relation_comodel_field,
+):
+    """Transform many2many relations into one2many (with possible data
+    loss).
```
legalsylvain, line 1949: Could you explicit dataloss. With some exemple like : 
In VX : a1 related to B1, B2, B3. A2 related to B3, B4 , B5.
In VX+1 : ?

openupgradelib/openupgrade.py
carmenbianca: good call, done.

openupgradelib/openupgrade.py
carmenbianca: added better documentation + example.

carmenbianca:
Ready for review again. Changes since last update:

- Better documentation/naming. I decided to call the entities in the One2many relationship parent/child instead of source/comodel, which was not clear.
- Even more logging.
- Added a 'COPY' strategy that, instead of causing data loss, copies the child record for each parent record that wants one.

The copy strategy uses the Odoo ORM (`copy()`). I don't know if this is good/kosher/permissible. Doing this manually in SQL seemed like a Really Bad Idea™ to me.

carmenbianca:
Oof, syntax error because Travis uses Python 3.6.

legalsylvain:
> Oof, syntax error because Travis uses Python 3.6.
> edit: ImportError because Travis uses Python 3.6 :'(

We have to switch to github actions, but not done until now (#322)...

carmenbianca:
Turns out Python 2 was the problem, not Python 3.6. I added a backport dependency, but it wouldn't be very difficult to get rid of the Enum and use something more primitive.

MiquelRForgeFlow:
I don't think it's useful. No need the data loss. You need to duplicate records and link them. See https://github.com/OCA/OpenUpgrade/pull/4719.

PR #345, opened 2023-09-13
[IMP] rename_modules: rename rating mixin parent_res_model as well

duong77476:
If not , rating feature will be gone for good

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1079,6 +1079,12 @@ def rename_models(cr, model_spec):
                     "UPDATE mail_activity SET res_model=%s where res_model=%s",
                     (new, old),
                 )
+        if table_exists(cr, "rating_rating"):
+            logged_query(
+                cr,
+                "UPDATE rating_rating SET parent_res_model=%s where parent_res_model=%s",
+                (new, old),
+            )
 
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -1079,6 +1079,12 @@ def rename_models(cr, model_spec):
                     "UPDATE mail_activity SET res_model=%s where res_model=%s",
                     (new, old),
                 )
+        if is_module_installed(cr, "rating"):
```
legalsylvain, line 1082: Better check if table existe.

legalsylvain (approved):

openupgradelib/openupgrade.py
duong77476: @legalsylvain ok as you wish sir

duong77476: @pedrobaeza Okay, done sir, i saw the above check both so i copy :v

pedrobaeza: Yes, it's true that the module installed check is used up on the code for `mail`. I suppose that one was for not having to check a lot of table existence and as the module name is expected to be constant. In this case, I think one check is enough.

duong77476: Okay, that 's fine by me

Issue #347, opened 2023-09-14
[11.0] `rating_rating` does not have `parent_res_model` column


zamberjo:
## Describe the bug
The PR https://github.com/OCA/openupgradelib/pull/345 uses a column called `parent_res_model` which is added in odoo version 11.0. 

Since this is executed in the pre-migrations, this column does not exist and gives an error:
```bash
2023-09-13 22:42:01,349 1167 ERROR migration_db odoo.sql_db: bad query: UPDATE rating_rating SET parent_res_model='ir.actions.report' where parent_res_model='ir.actions.report.xml'
ERROR: column "parent_res_model" does not exist
LINE 1: ...ng SET parent_res_model='ir.actions.report' where parent_res...
```

## To Reproduce
Migrate an odoo10 to odoo11

---

Could we apply the following change?

```python
if table_exists(cr, "rating_rating"):
```
to 
```python
if column_exists(cr, "rating_rating", "parent_res_model"):
```

zamberjo:
ping @pedrobaeza @duong77476 @legalsylvain

pedrobaeza:
It should be both conditions. Can you please make a PR?

zamberjo:
done

PR #348, opened 2023-09-14
[FIX] `rating_rating` does not have `parent_res_model` column

zamberjo:
SEE [Issue](https://github.com/OCA/openupgradelib/issues/347)

@pedrobaeza I have added only the `column_exists` condition as it implicitly checks whether the table exists or not:
Screenshot_299

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1079,7 +1079,7 @@ def rename_models(cr, model_spec):
                     "UPDATE mail_activity SET res_model=%s where res_model=%s",
                     (new, old),
                 )
-        if table_exists(cr, "rating_rating"):
+        if column_exists(cr, "rating_rating", "parent_res_model"):
             logged_query(
                 cr,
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -1079,7 +1079,7 @@ def rename_models(cr, model_spec):
                     "UPDATE mail_activity SET res_model=%s where res_model=%s",
                     (new, old),
                 )
-        if table_exists(cr, "rating_rating"):
+        if column_exists(cr, "rating_rating", "parent_res_model"):
```
pedrobaeza, line 1082:
```suggestion
        if table_exists(cr, "rating_rating") and column_exists(cr, "rating_rating", "parent_res_model"):
```

pedrobaeza: OK, seen that it works only with the column_exists.

pedrobaeza (approved):

PR #350, opened 2023-10-31
[ADD] openupgrade_160: fill_analytic_distribution

pedrobaeza:
Convert v15 analytic tags with distributions to v16 analytic distributions.

Extracted and adapted from Viindoo original queries in their migration scripts.

@Tecnativa

### DIFF
--- a/openupgradelib/openupgrade_160.py
+++ b/openupgradelib/openupgrade_160.py
@@ -401,3 +401,76 @@ def _convert_field_bootstrap_4to5_sql(cr, table, field, ids=None):
                 new_content,
                 id_,
             )
+
+
+def fill_analytic_distribution(
+    env,
+    table,
+    m2m_rel,
+    m2m_column1,
+    m2m_column2="account_analytic_tag_id",
+    column="analytic_distribution",
+):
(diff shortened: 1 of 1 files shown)


legalsylvain (commented):
my knowledge of analytical accounting is far too limited for me to give an opinion. 
But thank you for your contribution!

pedrobaeza:
It's already used successfully in `account`, which I already generalize a bit:

https://github.com/OCA/OpenUpgrade/blob/95a57861be937e9dc0bef2d56761a6c4e9e1b08a/openupgrade_scripts/scripts/account/16.0.1.2/pre-migration.py#L329

Being needed also in `sale`, `purchase`, and so on, I'm extracting it and generalizing for not repeating code. If any patch is needed, then we only need to touch in one place.

Merging then and next step is to test it replacing in `account` the code by the call to this.

marielejeune:
Hi @pedrobaeza, thanks for extracting this method.
It is sadly not working for sale order lines because `sale_order_line` has no line `analytic_account_id`. This column is on the related sale.order. Any idea on how to integrate this in the `sale` module migration?
See my comment here: https://github.com/OCA/OpenUpgrade/pull/4240#discussion_r1409470410

PR #351, opened 2023-11-02
[FIX] openupgrade_160: _convert_field_bootstrap_4to5_sql ids

chienandalu:
related: https://github.com/OCA/OpenUpgrade/pull/4205#issuecomment-1790261528

```python
>>> ids = [1, 3, 5]
>>> (ids,)
([1, 3, 5],)
>>> (tuple(ids),)
((1, 3, 5),)
```

cc @Tecnativa TT45734

ping @pedrobaeza

### DIFF
--- a/openupgradelib/openupgrade_160.py
+++ b/openupgradelib/openupgrade_160.py
@@ -389,7 +389,7 @@ def _convert_field_bootstrap_4to5_sql(cr, table, field, ids=None):
     params = ()
     if ids:
         sql += "WHERE id IN %s"
-        params = (ids,)
+        params = (tuple(ids),)
     cr.execute(sql, params)
     for id_, old_content in cr.fetchall():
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_160.py
```diff
@@ -389,7 +389,7 @@ def _convert_field_bootstrap_4to5_sql(cr, table, field, ids=None):
     params = ()
     if ids:
         sql += "WHERE id IN %s"
-        params = (ids,)
+        params = (*ids,)
```
pedrobaeza, line 392: Incorrect patch

```suggestion
        params = (tuple(ids),)
```

chienandalu: Right, I would have expected to work both ways but it doesn't

pedrobaeza (approved):

PR #352, opened 2023-11-07
[FIX] convert_html_fragment: Don't wrap full XMLs

pedrobaeza:
If the HTML string to convert is a fully XML compliant file including the header `<?xml version="...?>`, the wrapping process is breaking the final result, as the expected XML syntax doesn't include `?` in the middle and lxml returns a weird result `<?xml version="...??>` with 2 interrogations.

With this commit, we are not wrapping if we detect the XML header.

Fine-tune of #349 

@Tecnativa


chienandalu (approved):
:+1:

PR #338, opened 2023-07-11
[IMP] openupgrade_160, openupgrade_tools: BS4 to BS5 transformation yeahhhhh

duong77476-viindoo:
BS4 to BS5 transformation yeahhhhh

PR #353, opened 2023-11-13
[FIX] _convert_field_bootstrap_4to5_sql: convert parameters

chienandalu:
Fixes https://github.com/OCA/openupgradelib/pull/338#discussion_r1390249774

- Fix update query
- Avoid templating the sql directly from the method parameters.

ping @lijoantony

### DIFF
--- a/openupgradelib/openupgrade_160.py
+++ b/openupgradelib/openupgrade_160.py
@@ -8,7 +8,7 @@ import itertools
 import logging
 from itertools import product
 
-from psycopg2.extensions import AsIs
+from psycopg2 import sql
 
 from odoo.tools.translate import _get_translation_upgrade_queries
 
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_160.py
```diff
@@ -385,21 +385,23 @@ def _convert_field_bootstrap_4to5_sql(cr, table, field, ids=None):
     :param list ids:
         List of IDs, to restrict operation to them.
     """
-    sql = "SELECT id, %s FROM %s " % (field, table)
-    params = ()
+    sql = "SELECT id, %s FROM %s "
+    params = (AsIs(field), AsIs(table))
```
pedrobaeza, line 389: In fact, this should be `sql.Identifier`, and built with `sql.SQL`

chienandalu: Done :+1:

pedrobaeza (approved):

lijoantony:
Tested. It is working. 
Thank you very much!

PR #319, opened 2023-05-09
[IMP] update chained fields name in ir_exports_line and ir_filters tables

ernestotejeda:
Cc @Tecnativa

 
ping @pedrobaeza

PR #355, opened 2023-11-23
[IMP] rename_fields: Be able to use a cursor instead of Environment

rousseldenis:
In some cases, we want to use the rename_fields() method without using a full environment but with a cursor instead.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -658,7 +658,9 @@ def rename_fields(env, field_spec, no_deep=False):
     :param no_deep: If True, avoids to perform any operation that involves
       the environment. Not used for now.
     """
-    cr = env.cr
+    # This method has never implemented Environment usage apart cursor.
+    # To keep backward compatibility, check if env == Environment
+    cr = env.cr if isinstance(env, core.api.Environment) else env
     for model, table, old_field, new_field in field_spec:
         if column_exists(cr, table, old_field):
             rename_columns(cr, {table: [(old_field, new_field)]})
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):

openupgradelib/openupgrade.py
```diff
@@ -632,9 +632,16 @@ def rename_columns(cr, column_spec):
 
 
 def rename_fields(env, field_spec, no_deep=False):
```
StefanRijnhart, line 634: Would it make sense to allow an `env_or_cr` attribute here and deprecate passing an env when the major Odoo version is 17 or higher?

MiquelRForgeFlow: Suppose someday a PR like https://github.com/OCA/openupgradelib/pull/298 gets merged.

In this case, we would have:

```python
def rename_fields(env, field_spec, no_deep=False):
    rename_fields_cr(env.cr, field_spec)
    if no_deep:
        _rename_field_on_custom_view(env, ...)

def rename_fields_cr(cr, field_spec):
    ...
```

If we go by `env_or_cr`, and forcing cr in a version, then this version will not have this extra method that uses env.

StefanRijnhart: ok

StefanRijnhart (approved):

pedrobaeza (commented):
`deep=False` parameter is already for not doing stuff with ORM if not available, but you can get any environment even in `base` pre-migration, or fake it with a class containing the property `cr`.

@rousseldenis why do you want to do this?

IMO it's to complicate the number of methods without reason.

pedrobaeza (changes requested):
By the stated above

MiquelRForgeFlow:
Seeing @pedrobaeza comment, what do you think of doing this instead this?

```python
def rename_fields(env_or_cr, field_spec, no_deep=False):
    if not no_deep:
        env = env_or_cr
        cr = env.cr
    else:
        cr = env_or_cr
    (...code using cr here...)
    if not no_deep:
        (...code using env here...)
```
maybe we could rename `no_deep` to `no_env`... @pedrobaeza what do you think?

pedrobaeza:
OK about the renaming. It has this name for also not doing all the stuff we have in #319

PR #358, opened 2024-01-02
[ADD] convert_translatable_field_to_html method

AryanFanid:
Feature to convert translate=True fields to html


pedrobaeza:
I think this should be handled in the existing method, detecting if translatable (or passing an extra arg).

AryanFanid:
> I think this should be handled in the existing method, detecting if translatable (or passing an extra arg).

I started coding it like that (using a parameter) but then I thought this is specific for v16.0. So, i wrote it in openupgrade_160 file.

PR #360, opened 2024-02-14
[FIX] fill_analytic_distribution get analytic account column

remi-filament:
When trying to use fill_analytic_distribution for migrating purchase module (https://github.com/OCA/OpenUpgrade/pull/4303), I got the following error :
`column line.analytic_account_id does not exist`

The problem here is that the column is named account_analytic_id...
I then added a named parameter to fill_analytic_distribution() function to be able to set the column name holding analytic account id as done for the other columns related to analytic distribution


pedrobaeza (approved):

remi-filament:
Thanks @pedrobaeza

PR #363, opened 2024-04-05
[ADD] clean_transient_models

MiquelRForgeFlow:
clean_transient_models
-------

Clean transient models to prevent possible issues due to chained data.

To be run at the base post-migration script for having a general scope.Only works on > v8.

Used in https://github.com/OCA/OpenUpgrade/pull/4374.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -194,6 +194,7 @@ __all__ = [
     "convert_to_company_dependent",
     "cow_templates_mark_if_equal_to_upstream",
     "cow_templates_replicate_upstream",
+    "clean_transient_models",
 ]
 
 
@@ -3492,3 +3493,32 @@ def cow_templates_replicate_upstream(cr, mark_colname=None):
         """
         ).format(mark_identifier),
     )
+
+
+def clean_transient_models(cr):
+    """Clean transient models to prevent possible issues due to
+    chained data.
+
+    To be run at the base pre-migration script for having a general scope.
+    Only works on > v8.
+
+    :param cr: Database cursor.
+    """
+    if version_info[0] < 9:
+        raise Exception("Not supported Odoo version for this method.")
+    cr.execute("SELECT model FROM ir_model WHERE transient")
+    table_names = [get_model2table(x[0]) for x in cr.fetchall()]
+    for table_name in table_names:
+        if not table_exists(cr, table_name):
+            continue
+        try:
+            with cr.savepoint():
+                table = sql.Identifier(table_name)
+                query = sql.SQL(
+                    """DELETE FROM {} WHERE
+                    COALESCE(write_date, create_date, (now() at time zone 'UTC'))::timestamp
+                    < ((now() at time zone 'UTC') - interval '1 seconds')"""
+                ).format(table)
+                cr.execute(query)
+        except Exception as e:
+            logger.warning("Failed to clean transient table %s\n%s", table_name, str(e))


pedrobaeza (approved):
@StefanRijnhart is it OK for you?

openupgradelib/openupgrade.py
```diff
@@ -3492,3 +3493,29 @@ def cow_templates_replicate_upstream(cr, mark_colname=None):
...
+def clean_transient_models(env):
+    """Clean transient models to prevent possible issues due to
+    chained data.
+
+    To be run at the base post-migration script for having a general scope. Only
+    assured to work on > v9.
+
+    :param env: Environment/pool variable.
+    """
+    for mname in env:
```
StefanRijnhart, line 3507: Does not solve the issue. The transient models that are not loaded yet are not going to be in the env. Would you consider using the `transient` field in `ir_model` and predicting the table name from the model name?

MiquelRForgeFlow: Thanks

openupgradelib/openupgrade.py
```diff
@@ -3492,3 +3493,31 @@ def cow_templates_replicate_upstream(cr, mark_colname=None):
...
+        try:
+            with cr.savepoint():
+                table = sql.Identifier(table_name)
+                query = sql.SQL(
+                    """DELETE FROM {} WHERE
+                    COALESCE(write_date, create_date, (now() at time zone 'UTC'))::timestamp
+                    < ((now() at time zone 'UTC') - interval '1 seconds')"""
+                ).format(table)
+                cr.execute(query)
+            cr.commit()  # pylint: disable=E8102
```
MiquelRForgeFlow, line 3521: Not sure If we should avoid this line.

openupgradelib/openupgrade.py
```diff
@@ -3492,3 +3493,31 @@ def cow_templates_replicate_upstream(cr, mark_colname=None):
...
+
+    To be run at the base post-migration script for having a general scope.
+    Only works on > v8.
+
+    :param cr: Database cursor.
+    """
+    if version_info[0] < 9:
+        raise Exception("Not supported Odoo version for this method.")
+    cr.execute("SELECT model FROM ir_model WHERE transient")
+    table_names = [x[0] for x in cr.fetchall()]
```
StefanRijnhart, line 3510: Please replace dots by underscores and skip if the resulting table name does not exist in the database.

openupgradelib/openupgrade.py
StefanRijnhart: Yes, that would break incremental upgrade attempts during development/debugging of migration scripts. And is not necessary due to the saveoint.

openupgradelib/openupgrade.py
MiquelRForgeFlow: Done

openupgradelib/openupgrade.py
```diff
@@ -3492,3 +3493,32 @@ def cow_templates_replicate_upstream(cr, mark_colname=None):
         """
         ).format(mark_identifier),
     )
+
+
+def clean_transient_models(cr):
+    """Clean transient models to prevent possible issues due to
+    chained data.
+
+    To be run at the base post-migration script for having a general scope.
```
pedrobaeza, line 3502: Now this can be run on pre-migration:

```suggestion
    To be run at the base pre-migration script for having a general scope.
```

pedrobaeza (approved):

StefanRijnhart (approved):

Issue #362, opened 2024-03-27
chunked: should it flush ?


ivantodorovich:
The `chunked` method is invalidating the cache on each iteration, but it's never flushing.

I'm encountering an issue with this in Odoo `16.0`, where by using `chunked` I get into CacheMiss errors due to the cache being lost, including the previous one before it's being called ⚠️ .

By looking at the code I see it's invalidating the cache but it's never flushing. This looks like a red flag to me, although I've never encountered any issues with `chunked` before.

Perhaps it's due to https://github.com/odoo/odoo/pull/66938

pedrobaeza:
`chunked` was programmed a long time ago, and it's not adapted to newer versions.

ivantodorovich:
Ouch, that explains it then 😢 
I'm worried now about all the other openupgradelib's method I used in the past 😓

💡 Perhaps we could have a warning system to manually whitelist methods for new versions

pedrobaeza:
Well, this is the only method where it mixes ORM. The rest is thought for handling database things, so I don't think the rest are affected.

StefanRijnhart:
It's a good question, and I don't think anybody explicitly checked this after 16.0 was released, but `invalidate_all` actually executes a flush by default: https://github.com/odoo/odoo/blob/16.0/odoo/api.py#L727-L736

ivantodorovich:
> It's a good question, and I don't think anybody explicitly checked this after 16.0 was released, but `invalidate_all` actually executes a flush by default: https://github.com/odoo/odoo/blob/16.0/odoo/api.py#L727-L736

Indeed it does, since: https://github.com/odoo/odoo/pull/66938

The issue is we're not calling, the thing is we've stopped calling `invalidate_all` for version > `10` (we call `env.cache.invalidate` instead).

The fix could be to simply go back to calling `env.invalidate_all` for version >= `16`.

However I'm wondering if we should explicitly call `flush` for version >= `12` instead, when flush was introduced. By reading the code, IMO it's a miracle that it works. I can't explain why the data is not lost in versions 12, 13, 14 and 15; there must be something else that's flushing and saving it. The intention of this method is clearly to flush our changes to the db, whilst clearing the cache before the next iteration.

StefanRijnhart:
@ivantodorovich Thanks for clarifying! I think it's a very good idea of yours to go back to `invalidate_all` now that it is reintroduced, and to flush explicitely for the earlier versions that support it.

[...]

PR #364, opened 2024-04-10
[FIX] chunked: explicitly flush on every iteration

ivantodorovich:
Fixes:
- https://github.com/OCA/openupgradelib/issues/362

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -3261,17 +3261,11 @@ def chunked(records, single=True):
     """Memory and performance friendly method to iterate over a potentially
     large number of records. Yields either a whole chunk or a single record
     at the time. Don't nest calls to this method."""
-    if version_info[0] > 10:
-        invalidate = records.env.cache.invalidate
-    elif version_info[0] > 7:
-        invalidate = records.env.invalidate_all
-    else:
-        raise Exception("Not supported Odoo version for this method.")
     size = core.models.PREFETCH_MAX
     model = records._name
     ids = records.with_context(prefetch_fields=False).ids
     for i in range(0, len(ids), size):
-        invalidate()
+        openupgrade_tools.invalidate_cache(records.env, flush=True)
         chunk = records.env[model].browse(ids[i : i + size])
         if single:
             for record in chunk:
--- a/openupgradelib/openupgrade_tools.py
+++ b/openupgradelib/openupgrade_tools.py
@@ -26,6 +26,8 @@ import logging
 from lxml.etree import tostring
 from lxml.html import fromstring
 
+logger = logging.getLogger(__name__)
+
 
 def table_exists(cr, table):
     """Check whether a certain table or view exists"""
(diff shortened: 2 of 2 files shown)


openupgradelib/openupgrade.py
```diff
@@ -3261,17 +3261,43 @@ def chunked(records, single=True):
...
-        invalidate = records.env.cache.invalidate
-    elif version_info[0] > 7:
-        invalidate = records.env.invalidate_all
+    # Warning on possibly untested versions where chunked might not work as expected
+    if version_info[0] > 17:  # unreleased version at this time
+        logger.warning(
+            f"chunked method might not work as expected in Odoo {version_info[0]}. "
+            "Please report any issue you may find and consider bumping this warning "
+            "to the next version otherwise."
+        )
```
StefanRijnhart, line 3270: How does this relate to the rest of OpenUpgrade? Would we need a similar warning for every method?

openupgradelib/openupgrade.py
```diff
@@ -3261,17 +3261,43 @@ def chunked(records, single=True):
...
+    # Before each iteration we must flush everything and clear the cache.
+    # This differs depending on the Odoo version.
+    if version_info[0] >= 16:
+
+        def reset():
+            records.env.invalidate_all(flush=True)
```
StefanRijnhart, line 3276: `True` is the default argument to `flush`. Would you consider leaving it out, as is conventional for defaults?

openupgradelib/openupgrade.py
```diff
@@ -3261,17 +3261,43 @@ def chunked(records, single=True):
...
+            records.env.invalidate_all()
+
     else:
         raise Exception("Not supported Odoo version for this method.")
     size = core.models.PREFETCH_MAX
     model = records._name
     ids = records.with_context(prefetch_fields=False).ids
     for i in range(0, len(ids), size):
-        invalidate()
+        reset()
```
StefanRijnhart, line 3300: I like having a name like `invalidate` as it reflects the aim more clearly and is aligned with the underlying Odoo methods it calls. Can you tell my why you changed it?

openupgradelib/openupgrade.py
```diff
@@ -3261,17 +3261,43 @@ def chunked(records, single=True):
...
+            records.env.cache.invalidate()
+
+    elif version_info[0] >= 11:
+
+        def reset():
+            records.env.cache.invalidate()
+
+    elif version_info[0] >= 8:
+
+        def reset():
```
StefanRijnhart, line 3291: A version dependent shorthand method to invalidate the cache might proof to be useful outside of `chunked`. Would you consider moving it to `openupgrade_tools.py`? It might also be an aesthetic improvement, depending on your taste of course.

openupgradelib/openupgrade.py
ivantodorovich: Quoting @pedrobaeza here: https://github.com/OCA/openupgradelib/issues/362#issuecomment-2023092055

> Well, this is the only method where it mixes ORM. The rest is thought for handling database things, so I don't think the rest are affected.

I don't think I have enough knowledge of this tool to answer this.

I was just bitten by the fact that I tried to use this method in `16.0` and got corrupted data as a result because of this silent error 😓 My motivation is preventing other people from going through the same in future versions, in case something like this changes again.

openupgradelib/openupgrade.py
ivantodorovich: It is. I wanted to make it clear that flush *is* happening, which is something you may not immediately realize by just reading `records.env.invalidate_all()`. But I can remove it and let the default play out if you think it's best

openupgradelib/openupgrade.py
ivantodorovich: My reasoning comes from the fact that we're not only invalidating the "read cache", we're also flushing the "to write cache"

Something like `flush_and_invalidate` seemed more accurate but it's long and explicit. Then I thought for each iteration a "reset" of the cache is done, hence the name.

In any case, I've extracted the invalidation/flushing to a separate method, as you suggested

openupgradelib/openupgrade.py
ivantodorovich: Sure ;) Check out the latests changes.

StefanRijnhart (approved):
Thanks for the updates!
(code review, no test)

PR #369, opened 2024-05-14
[FIX] load_data: update to version 17.0

royle-viindoo:
changed on 17.0
(https://github.com/odoo/odoo/commit/5bf1207c8cb7575fba4501aff4932c2cc952bf9f)

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -271,7 +271,7 @@ def check_values_selection_field(cr, table_name, field_name, allowed_values):
     return res
 
 
-def load_data(cr, module_name, filename, idref=None, mode="init"):
+def load_data(env_or_cr, module_name, filename, idref=None, mode="init"):
     """
     Load an xml, csv or yml data file from your post script. The usual case for
     this is the
@@ -287,6 +287,8 @@ def load_data(cr, module_name, filename, idref=None, mode="init"):
     Leave it to the user to actually delete existing resources that are
     marked with 'noupdate' (other named items will be deleted
     automatically).
+
+    Notes: Argument "env_or_cr" is an cr until 16 and is an env is required since 17.
 
 
     :param module_name: the name of the module
@@ -305,6 +307,15 @@ def load_data(cr, module_name, filename, idref=None, mode="init"):
         filled which are not contained in the data file.
     """
 
+    cr = env_or_cr
+    if version_info[0] >= 17:
+        if not isinstance(env_or_cr, core.api.Environment):
+            logger.error(
+                "Pass argument 'env_or_cr' as Environment parameter since 17.0"
+            )
+        cr = env_or_cr.cr
+    elif not isinstance(env_or_cr, core.sql_db.Cursor):
+        logger.error("Pass argument 'env_or_cr' as Cursor parameter until 16.0")
     if idref is None:
         idref = {}
     logger.info("%s: loading %s" % (module_name, filename))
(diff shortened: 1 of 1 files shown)


legalsylvain:
I dont get it.

pedrobaeza:
On v16-, you call the method `load_data(cr, ...)`. On v17+, you call it `load_data(env, ...)`.

legalsylvain:
OK ! 

You propose to replace 
```
def load_data(cr, module_name, filename, idref=None, mode="init"):
```

by 
```
def load_data(env_or_cr, module_name, filename, idref=None, mode="init"):
```
 
Adding a comment to say that an cr is required until 16 and an env is required since 17, right ?


Personnaly, I prefer to keep the same signature ``def load_data(cr, module_name, filename, idref=None, mode="init"):``

Then, add : 
```
env_or_cr = api.Environment(cr, SUPERUSER_ID, {}) if version_info[0] >= 17 else cr
```
then simply replace ```tools.convert_xml_import(cr, module_name, fp, idref, mode=mode)```
by ```tools.convert_xml_import(env_or_cr, module_name, fp, idref, mode=mode)```

But both works, so it's maybe a matter of taste. I think that it can maybe generate more errors to change the type of the argument, depending of the version of the module that call the function.

royle-viindoo:
> OK !
> 
> You propose to replace
> 
> ```
> def load_data(cr, module_name, filename, idref=None, mode="init"):
> ```
> 
> by
> 
> ```
> def load_data(env_or_cr, module_name, filename, idref=None, mode="init"):
> ```
> 
> Adding a comment to say that an cr is required until 16 and an env is required since 17, right ?
> 
> Personnaly, I prefer to keep the same signature `def load_data(cr, module_name, filename, idref=None, mode="init"):`
> 
> Then, add :
> 
> ```
> env_or_cr = api.Environment(cr, SUPERUSER_ID, {}) if version_info[0] >= 17 else cr
> ```
> 
> then simply replace `tools.convert_xml_import(cr, module_name, fp, idref, mode=mode)` by `tools.convert_xml_import(env_or_cr, module_name, fp, idref, mode=mode)`
> 
> But both works, so it's maybe a matter of taste. I think that it can maybe generate more errors to change the type of the argument, depending of the version of the module that call the function.



> OK !
> 
> You propose to replace
> 
> ```
> def load_data(cr, module_name, filename, idref=None, mode="init"):
> ```
> 
> by
> 
> ```
> def load_data(env_or_cr, module_name, filename, idref=None, mode="init"):
> ```
> 
> Adding a comment to say that an cr is required until 16 and an env is required since 17, right ?
> 
> Personnaly, I prefer to keep the same signature `def load_data(cr, module_name, filename, idref=None, mode="init"):`
> 
> Then, add :
> 
> ```
> env_or_cr = api.Environment(cr, SUPERUSER_ID, {}) if version_info[0] >= 17 else cr
> ```
> 
> then simply replace `tools.convert_xml_import(cr, module_name, fp, idref, mode=mode)` by `tools.convert_xml_import(env_or_cr, module_name, fp, idref, mode=mode)`
> 
> But both works, so it's maybe a matter of taste. I think that it can maybe generate more errors to change the type of the argument, depending of the version of the module that call the function.

@legalsylvain Thank you for your solution. I find it very interesting and low risk. I have fixed it according to your solution.

pedrobaeza:
Why I was talking to pass `env` is for the same reason as stated in the Odoo commit: to re-use the environment and avoid to regenerate it each time, and to share the context. Don't you think it's worth?

royle-viindoo:
> Why I was talking to pass `env` is for the same reason as stated in the Odoo commit: to re-use the environment and avoid to regenerate it each time, and to share the context. Don't you think it's worth?

@pedrobaeza I have updated it according to your suggestion, please review it

openupgradelib/openupgrade.py
```diff
@@ -329,21 +337,21 @@ def load_data(cr, module_name, filename, idref=None, mode="init"):
         if ext == ".csv":
             noupdate = True
             tools.convert_csv_import(
-                cr, module_name, pathname, fp.read(), idref, mode, noupdate
+                env_or_cr, module_name, pathname, fp.read(), idref, mode, noupdate
             )
         elif ext == ".yml":
             yaml_import(cr, module_name, fp, None, idref=idref, mode=mode)
```
pedrobaeza, line 343: This one hasn't been changed, isn't it?

royle-viindoo: Has been changed at [5bf1207c8cb7575fba4501aff4932c2cc952bf9f](https://github.com/odoo/odoo/commit/5bf1207c8cb7575fba4501aff4932c2cc952bf9f)

StefanRijnhart: `yaml_import` was removed from Odoo after 11 anyway.

StefanRijnhart (approved):

openupgradelib/openupgrade.py
pedrobaeza: I know, but right now, the code will fail in 11-, as `cr` has no value for that case. Assign previously `cr = env_or_cr`, or directly put here `env_or_cr`.

royle-viindoo: you are right, I added more at line 310

pedrobaeza (approved):
Thanks for the patience attending our comments.

PR #370, opened 2024-06-10
[IMP] BS4->BS5: performance

chienandalu:
Avoid record values which html is empty using the same rules that odoo.tools.is_html_empty uses. This way we can simplify the call avoiding a default domain as well. In any case, we can use that domain to refine certain models.

cc @Tecnativa TT46020

please check @pedrobaeza

### DIFF
--- a/openupgradelib/openupgrade_160.py
+++ b/openupgradelib/openupgrade_160.py
@@ -17,6 +17,7 @@ from .openupgrade import logged_query, table_exists, update_field_multilang
 from .openupgrade_tools import (
     convert_html_fragment,
     convert_html_replacement_class_shortcut as _r,
+    not_html_empty_where_clause,
     replace_html_replacement_attr_shortcut as _attr_replace,
     replace_html_replacement_class_rp_by_inline_shortcut as _class_rp_by_inline,
 )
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
nice improvment.

pedrobaeza (approved):

hhgabelgaard:
@chienandalu @pedrobaeza What's the Python version compability requirements for OpenUpgradelib?
2.7 according to  setup.py https://github.com/OCA/openupgradelib/blob/master/setup.py#L53

So please avoid f-strings - Some of us are using OpenUpgrade on older Odoo/Python versions

pedrobaeza:
We are using them too in old versions, but look that this is something exclusively for version 16.0. Where are the problem?

StefanRijnhart:
@pedrobaeza this is the problem
```
>>> import openupgradelib
>>> from openupgradelib import openupgrade_tools
  File "<stdin>", line 1, in <module>
  File "openupgradelib/openupgrade_tools.py", line 415
    )"""
       ^
SyntaxError: invalid syntax
```

StefanRijnhart:
Maybe we should bump the version with a major and drop compatibility with Python 2.x.

pedrobaeza:
OK, I think we can remove the f-string, as it's not strictly needed. @chienandalu can you?

chienandalu:
No problem

chienandalu:
Addressing fixes here: https://github.com/OCA/openupgradelib/pull/371

hhgabelgaard:
@chienandalu Thanks!

This community rocks!

PR #371, opened 2024-06-11
[FIX][IMP] _convert_field_bootstrap_4to5_orm: translatable fields

chienandalu:
The improvement made in 40741d960477bfcfea5c1e6a518863a7e6f92065 couldn't deal with json columns of translatable fields.

In any case a more simpler approach can be took if we simply discard all the records which don't even have a `class` attribute in their html code, meaning that we don't need to make the Bootstrap migration at all.

cc @Tecnativa TT46020

### DIFF
--- a/openupgradelib/openupgrade_160.py
+++ b/openupgradelib/openupgrade_160.py
@@ -11,13 +11,13 @@ from itertools import product
 from psycopg2 import sql
 from psycopg2.extras import Json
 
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow (approved):

MiquelRForgeFlow:
We should go fast with this.

pedrobaeza (approved):

PR #373, opened 2024-06-14
[IMP] disable_invalid_filters: log filter probes

chienandalu:
To have a hint on possible bottleneck issues for filters with heavy queries. We can disable it setting the new `verbose` parameter to `False`.

cc @Tecnativa TT46020

check @pedrobaeza

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2743,7 +2743,7 @@ def delete_record_translations(cr, module, xml_ids, field_list=None):
 
 
 # flake8: noqa: C901
-def disable_invalid_filters(env):
+def disable_invalid_filters(env, verbose=True):
     """It analyzes all the existing active filters to check if they are still
     correct. If not, they are disabled for avoiding errors when clicking on
     them, or worse, if they are default filters when opening the model/action.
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -2814,6 +2814,13 @@ def format_message(f):
             with savepoint(env.cr):
                 # Strange artifact found in a filter
                 domain = f.domain.replace("%%", "%")
+                if verbose:
+                    logger.info(
+                        "Probing filter '%s' on model %s with domain %s",
```
pedrobaeza, line 2819:
```suggestion
                        "Checking filter '%s' on model %s with domain %s",
```

`Probing` is not a valid gerund: https://translate.google.com/details?sl=en&tl=es&text=probing&op=translate

chienandalu: done

openupgradelib/openupgrade.py
```diff
@@ -2814,6 +2814,13 @@ def format_message(f):
             with savepoint(env.cr):
                 # Strange artifact found in a filter
                 domain = f.domain.replace("%%", "%")
+                if verbose:
```
pedrobaeza, line 2817: I think you should add the message at the beginning of the loop.

pedrobaeza (approved):

PR #374, opened 2024-06-18
[FIX] migrate_translations_to_jsonb: avoid ir_translation terms cleanup

chienandalu:
We're making use of core's `_get_translation_upgrade_queries` which aside from provide the proper queries to upgrade the translated fields to the new json structure tries to cleanup the former ir_tanslation records.

In OpenUpgrade we try to keep historic info after migration so we're able to quickly react to possible inconsistencies.

cc @Tecnativa TT49615

### DIFF
--- a/openupgradelib/openupgrade_160.py
+++ b/openupgradelib/openupgrade_160.py
@@ -4,7 +4,6 @@
 """This module provides simple tools for OpenUpgrade migration, specific for
 the >=16.0 migration.
 """
-import itertools
 import logging
 from itertools import product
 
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_160.py
```diff
@@ -54,9 +53,14 @@ def migrate_translations_to_jsonb(env, fields_spec):
...
-                _get_translation_upgrade_queries(env.cr, field)
-            ):
+            # Ignore cleanup queries as we want to keep the original ir_translation
+            # table records in order to be able to fix possible inconsistencies once
+            # we're migrated.
+            migrate_queries, _cleanup_queries = _get_translation_upgrade_queries(
+                env.cr, field
+            )
+            for query in migrate_queries:
+
```
pedrobaeza, line 63: Remove this empty line

chienandalu:
There's a limitation here as the core method executes a query inside the method that expects the table `ir_translation` to be renamed as `_ir_translation`... so we we should do it in `base` `pre-migration` to be able to use it: https://github.com/OCA/OCB/blob/1e74b85c260fc0c6d15e7e4e13d9afcb9c2e9909/odoo/tools/translate.py#L1616-L1631

pedrobaeza:
I recommend then that this method renames the table before, and restores it after.

pedrobaeza:
Now that I see the diff, why previous approach replacing inside the query didn't serve?

pedrobaeza:
OK, understood

@marielejeune FYI

pedrobaeza (approved):

PR #379, opened 2024-08-22
latest setuptool_scm version is not compatible with python 3.7

JohnnyVM:
Version 8.0.0 of setuptools_scm onwards imports Protocol from typing but Protocol is not available for versions of python3.7, which is the one used by the odoo:13.0 and odoo:14.0 docker

https://github.com/pypa/setuptools_scm/blob/b5dbba7a154b90f401ac1b9ca37dc920b5e82c33/src/setuptools_scm/_config.py#L11

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,8 @@ setup(
     include_package_data=True,
     setup_requires=[
         "setuptools_scm<6.0.0; python_version <= '3.6'",
-        "setuptools_scm; python_version > '3.6'",
+        "setuptools_scm<8; python_version < '3.8'",
+        "setuptools_scm; python_version >= '3.8'",
     ],
     install_requires=[
         "lxml",


JohnnyVM:
the documentation action is failing becouse:
       WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate (_ssl.c:1007)'))': /oca-simple-and-pypi/setuptools-scm/
       
I think is not related to my PR, can someone fix it in master? after that i will rebase the PR

sbidoul:
The certificate issue was due to a DNS error, should be resolved in 5 minutes.

setup.py
```diff
@@ -33,7 +33,8 @@
     include_package_data=True,
     setup_requires=[
         "setuptools_scm<6.0.0; python_version <= '3.6'",
-        "setuptools_scm; python_version > '3.6'",
+        "setuptools_scm<8.0.0; python_version < '3.8'",
+        "setuptools_scm; python_version >= '3.8'",
```
sbidoul, line 37: This should not be necessary has setuptools_scm 8 declares it needs python >= 3.8, so pip will not pick it and use a lower version with python 3.7. Unless you have a very old pip ?

JohnnyVM: i am using docker to migrate a database from 13 to 14, i dont know if that is a workflow supported, but the pip i am using is the pip is installed with docker:14.0

JohnnyVM:
> The certificate issue was due to a DNS error, should be resolved in 5 minutes.

i cannot rerun the action, I don't have the necessary permissions, can someone re-launch it for me?

setup.py
```diff
@@ -33,7 +33,8 @@
     include_package_data=True,
     setup_requires=[
         "setuptools_scm<6.0.0; python_version <= '3.6'",
-        "setuptools_scm; python_version > '3.6'",
+        "setuptools_scm<8.0.0; python_version < '3.8'",
```
sbidoul, line 36:
```suggestion
        "setuptools_scm<8; python_version < '3.8'",
```

nit

sbidoul: BTW `<= 3.6` should be replaced by `<3.7` because I think 3.6.1 would match <=3.6 which not what's intended.

JohnnyVM:
if all is okay can the PR be merged?

StefanRijnhart:
One step towards that would be to squash commits and align the commit message with the guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

JohnnyVM:
Hi, i had time to back
Message fixed :)
Am i missing something more?
Kind regards

StefanRijnhart (approved):

StefanRijnhart:
@sbidoul Can you update your review?

sbidoul (commented):
That said, I still don't understand why this PR is necessary, as I tried installing the `master` branch in a python 3.7 venv and it works fine:

```
git clone https://github.com/oca/openupgradelib
python3.7 -m venv venv
venv/bin/pip install -U pip wheel setuptools
venv/bin/pip install ./openupgradelib
venv/bin/pip list
```

JohnnyVM:
How to reproduce:
```Dockerfile
FROM odoo:14

USER root

RUN apt update --fix-missing && apt update && apt install -y git
RUN python3  -m pip install wheel setuptools git+https://github.com/OCA/openupgradelib.git@master#egg=openupgradelib

USER odoo

ENTRYPOINT ["/entrypoint.sh"]
CMD ["odoo"]
```

Error:
```
WARN[0005] SHELL is not supported for OCI image format, [/bin/bash -xo pipefail -c] will be ignored. Must use `docker` format
--> a8586f9af0a2
STEP 4/7: RUN python3  -m pip install wheel setuptools git+https://github.com/OCA/openupgradelib.git@master#egg=openupgradelib
Collecting wheel
  Downloading https://files.pythonhosted.org/packages/c7/c3/55076fc728723ef927521abaa1955213d094933dc36d4a2008d5101e1af5/wheel-0.42.0-py3-none-any.whl (65kB)
Requirement already satisfied: setuptools in /usr/lib/python3/dist-packages (40.8.0)
Collecting openupgradelib from git+https://github.com/OCA/openupgradelib.git@master#egg=openupgradelib
  Cloning https://github.com/OCA/openupgradelib.git (to revision master) to /tmp/pip-install-we0hvqrr/openupgradelib
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-we0hvqrr/openupgradelib/setup.py", line 62, in <module>
        tests_require=test_requirements,
      File "/usr/lib/python3/dist-packages/setuptools/__init__.py", line 145, in setup
        return distutils.core.setup(**attrs)
      File "/usr/lib/python3.7/distutils/core.py", line 108, in setup
        _setup_distribution = dist = klass(attrs)
      File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 447, in __init__
        k: v for k, v in attrs.items()
      File "/usr/lib/python3.7/distutils/dist.py", line 292, in __init__
        self.finalize_options()
      File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 739, in finalize_options
        ep.load()(self, ep.name, value)
      File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2411, in load
        return self.resolve()
      File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2417, in resolve
        module = __import__(self.module_name, fromlist=['__name__'], level=0)
      File "/tmp/pip-install-we0hvqrr/openupgradelib/.eggs/setuptools_scm-8.1.0-py3.7.egg/setuptools_scm/__init__.py", line 8, in <module>
        from ._config import DEFAULT_LOCAL_SCHEME
      File "/tmp/pip-install-we0hvqrr/openupgradelib/.eggs/setuptools_scm-8.1.0-py3.7.egg/setuptools_scm/_config.py", line 13, in <module>
        from typing import Protocol
    ImportError: cannot import name 'Protocol' from 'typing' (/usr/lib/python3.7/typing.py)

    ----------------------------------------
    ```
    
    current work around? update setupt tools previously to install openupgradelib:
    ```Dockerfile
FROM odoo:14

USER root

RUN apt update --fix-missing && apt update && apt install -y git
RUN python3  -m pip install -U wheel setuptools
RUN python3  -m pip install git+https://github.com/OCA/openupgradelib.git@master#egg=openupgradelib

USER odoo

ENTRYPOINT ["/entrypoint.sh"]
CMD ["odoo"]
    ```
    
    It seem like update setuptools fix the issue because newest versions do some type of fix for the import, but default docker setuptools fails.
    
    I think is more elegant fix the setup.py but give that exist a work around feel free of close the PR, if not i am going to solve the @sbidoul  comment  and reupdate in a couple of hours

sbidoul:
Since the issue seems to be in setuptools and there is a workaround with a simple upgrade I personally find it preferable to recommend that instead of adding complexity in this project.

StefanRijnhart:
@sbidoul It is undeniably the case that setuptool_scm 8.0.0 dropped support for python 3.7, introducing the offending import of `typing.Protocol` that is causing OP's issue: https://github.com/pytest-dev/pytest/pull/11152/files#diff-af0c5977e27eeffde9235704eaccb3f3351575f23ea385f6b1e8f48a87249036
As such I don't think it is unreasonable to distinguish between versions here.

sbidoul:
@StefanRijnhart  Sure but that setuptools_scm commit you mention also adds `python_requires = >=3.8` so a sufficiently recent and correct version of setuptools will not select setuptools_scm >= 8 on python 3.7. In that sense, this PR works around a bug in a very old version of setuptools. Or do I misunderstand something? 

So the question is do we want to drag along that work around along in openupgradelib, or simply recommand users to upgrade setuptools.

StefanRijnhart:
@sbidoul Thanks for the clarification. After having worked for many years without this *basic* dependency management feature, I wasn't really aware that setuptools caught up at some point :grin:

@JohnnyVM do you have a compelling reason not to update setuptools beyond the version that your Python 3.7 distribution came with?

JohnnyVM:
No, the work around work for me.
I am going to close the PR then, anyway it was interesting :)
Kind regards

StefanRijnhart:
@JohnnyVM Thanks for confirming!

PR #380, opened 2024-08-30
[FIX] load_data: fail early on missing path

yajo:
Without this patch, if you call this function with a filename that can't be found, you'll get a confusing error like `NameError: name 'fp' is not defined`, that gives no details about the real problem.

@moduon MT-7047

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -333,6 +333,11 @@ def load_data(env_or_cr, module_name, filename, idref=None, mode="init"):
                     break
                 except OSError:  # pylint: disable=W7938
                     pass
+            else:
+                raise OSError(
+                    "Couldn't find file %s in upgrade path (%s)"
+                    % (filename, tools.config["upgrade_path"])
+                )
         else:
             raise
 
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -333,6 +333,11 @@ def load_data(env_or_cr, module_name, filename, idref=None, mode="init"):
                     break
                 except OSError:  # pylint: disable=W7938
                     pass
+            else:
+                raise OSError(
+                    "Couldn't find file %s in upgrade path (%s)"
```
pedrobaeza, line 338:
```suggestion
                    "Couldn't find file %s in the upgrade paths (%s)"
```

pedrobaeza:
Please squash both commits into one. Remember that doing `commit --amend` + `push -f` is also valid in GitHub, as it puts the diff in the timeline.

legalsylvain (approved):
Lgtm.

yajo:
> Please squash

You can just click on "squash and merge", since there's no ocabot involved here 🙂

pedrobaeza:
Right, I didn't remember in this project, we do direct merges without ocabot. Anyway, it's better to be consistent across all the contributions IMO.

PR #382, opened 2024-09-25
[ADD] validate_fk_constraints function

thomaspaulb:
This is useful in cases where you need to delete a lot of records, and you speed the deletion up by disabling FK triggers, but you want to validate integrity after deletion.

Sample use:

```
env.cr.execute('alter table res_partner disable trigger all')
env.cr.execute(delete from res_partner where id = 738152')
validate_fk_constraints(env.cr, 'res_partner', 'id')
# of course this will fail, so then you can iterate until referential integrity is achieved again
```

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -195,6 +195,7 @@ __all__ = [
     "cow_templates_mark_if_equal_to_upstream",
     "cow_templates_replicate_upstream",
     "clean_transient_models",
+    "validate_fk_constraints",
 ]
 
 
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -3547,3 +3548,50 @@ def clean_transient_models(cr):
                 cr.execute(query)
         except Exception as e:
             logger.warning("Failed to clean transient table %s\n%s", table_name, str(e))
+
+
+VALIDATE_FK_QUERY = """
```
MiquelRForgeFlow, line 3553: why outside of validate_fk_constraints? I would put inside...

PR #383, opened 2024-09-30
[FIX] Convert tax tags based on move type instead of amount.

ddejong-therp:
Like the title says.


pedrobaeza (commented):
You may have tax tags on other moves different from invoices (for example, payroll moves), so I don't think this is correct.

PR #386, opened 2024-10-10
[FIX] logged_query: deprecation warning for tools.ustr for version >= 18

marielejeune:
`odoo.tools.ustr` is deprecated since Odoo 18. See full comment here: https://github.com/odoo/odoo/blob/e65d31339d383169bfebab7619073d7117198cfa/odoo/loglevels.py#L65

This improvement avoids raising the warning in the logs.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -127,6 +127,11 @@ if version_info[0] > 7:
 else:
     api = False
 
+if version_info[0] < 18:
+    decode = tools.ustr
+else:
+    decode = lambda s: s.decode("utf-8", errors="strict")
+
 # Setting the target version as an environment variable allows OpenUpgrade
 # to skip methods that are called in every version but really only need to
 # run in the target version. Make the target version available to OpenUpgrade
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade.py
```diff
@@ -1721,9 +1721,17 @@ def logged_query(cr, query, args=None, skip_no_result=False):
         duration = datetime.now() - start
         if log_msg:
             try:
-                full_query = tools.ustr(cr._obj.query)
+                if version_info[0] < 18:
+                    full_query = tools.ustr(cr._obj.query)
+                else:
+                    full_query = cr._obj.query.decode("utf-8", errors="strict")
```
StefanRijnhart, line 1727: Thanks! We keep a small number of functions with a version depending definition at the top of this file (around line 123). How about adding a custom `decode` function to them?

marielejeune: Good idea, but how to "import" python `decode` method? 
Doing something like 
```
if version_info[0] < 18:
    custom_decode = tools.ustr
else:
    custom_decode = decode
```
won't work because `decode` is not defined?

StefanRijnhart: I would just call it `decode`, so something like

```
decode = lambda s: s.decode("utf-8", errors="strict")
```

marielejeune: Done

StefanRijnhart (approved):

pedrobaeza (approved):

PR #387, opened 2024-10-10
[FIX] convert_field_to_html: fix `format` call on string

GuillemCForgeFlow:


MiquelRForgeFlow (approved):

pedrobaeza:
I don't get the fix... and this should be done using `sql.SQL` and `sql.Identifier` for proper SQL wrapping.

MiquelRForgeFlow:
> I don't get the fix

`.format` was being called outside of the `cr.execute()`

pedrobaeza:
Ah, OK, didn't see it.

PR #389, opened 2024-10-28
[UPD] m2o_to_x2m: docstring recommends outdated usage

hbrunn:
it's not true for some versions of OpenUpgrade already that we need to retain the column's value manually, because we centrally [patch away dropping columns](https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_framework/odoo_patch/odoo/addons/base/models/ir_model.py#L34).

The recommendation leads to [confusion](https://github.com/OCA/OpenUpgrade/pull/4465#discussion_r1806480453), so I suggest to restrict it to ancient versions of OpenUpgrade.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1867,8 +1867,9 @@ def get_model2table(model):
 def m2o_to_x2m(cr, model, table, field, source_field):
     """
     Transform many2one relations into one2many or many2many.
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

hbrunn:
thanks Pedro. Just to be sure for posterity: We also don't need this for unpatched modern Odoos, because it will drop the column only at the end of the migration

PR #390, opened 2024-10-28
[ADD] convert_from_company_dependent

hbrunn:
this is the function I propose in https://github.com/OCA/OpenUpgrade/pull/4457#discussion_r1819214161, will be useful elsewhere too

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -64,6 +64,13 @@ if hasattr(core, "release"):
 else:
     import release
 
+ir_property_TYPE2FIELD = {}
+try:
+    from odoo.addons.base.models.ir_property import TYPE2FIELD as ir_property_TYPE2FIELD
+except ImportError:  # pylint: disable=except-pass
+    # don't fail for very old Odoos
+    pass
+
 Many2many = False
 One2many = False
 one2many = False
@@ -198,6 +205,7 @@ __all__ = [
     "delete_sql_constraint_safely",
     "set_xml_ids_noupdate_value",
     "convert_to_company_dependent",
+    "convert_from_company_dependent",
     "cow_templates_mark_if_equal_to_upstream",
     "cow_templates_replicate_upstream",
     "clean_transient_models",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But I think this depends on the business logic behind, as you are converting a multi-valued field (potentially one per company) into only one value, so you need to decide what to do in case of conflict depending on the use case.

hbrunn:
I think the general case will be pretty much the same everywhere:

v(n-1): model has no company_id and property property_whatever
v(n): model has company_id, property_whatever becomes field whatever

Then the user of the function duplicates every row of model for each company where it makes sense (that's the business logic depending part), and calls `convert_from_company_dependent(model, property_whatever, whatever)`, which is invariant wrt business logic

pedrobaeza:
But in such duplication, `res_id` will be another, and thus, this method won't serve for transferring the value of the corresponding property for the duplicated records.

hbrunn:
it will, because the caller sets company_id for the cloned records as it's appropriate.

consider the [account migration](https://github.com/OCA/OpenUpgrade/pull/4457/files#diff-406c7e73ba75feeb54738e35feff14fb8fe419b838b4461e864cf033b2fa0f52R480), there they duplicate tax groups as it makes sense into the different companies, and in the next function I can replace the horrible incomplete sql statements with three calls to the function I propose.

Wherever Odoo decides to go from properties to company aware on model level, we need the exact same thing.

openupgradelib/openupgrade.py
```diff
@@ -3458,6 +3466,75 @@ def convert_to_company_dependent(
...
+    return logged_query(
+        env.cr,
+        f"""
+        UPDATE {table_name}
+        SET {destination_field_name} = (
+            SELECT {value_expression}
+            FROM ir_property ip
+            WHERE ip.fields_id={field_id} --- {origin_field_name}
+            AND (
+                ip.res_id = '{model_name}.' || {table_name}.id
```
pedrobaeza, line 3524: I insist that this is the problem. The duplicated record has an ID that is not on `ir_property` table, as the corresponding one is the previous record ID.

hbrunn:
aaah wait, I get it now. This indeed needs the extra field I propose on the PR to link the clones to their origin record. That's a good thing to have in those cases anyways, will do next week

openupgradelib/openupgrade.py
hbrunn: thanks for insisting, see below

hbrunn:
I've just grepped through the other analysis files and didn't find another place where we need this for the v17 migration, so for the time being I put the function in the v17 account migration, and we move it to openupgradelib when the next similar case comes up

PR #391, opened 2024-11-01
[IMP]openupgrade: disable_invalid_filters

leemannd:
### Trying to solve
After a migration it was not possible to open the settings submenu: "User-defined filters".
This was due to invalid user filters.

### Cause
The current script is not testing action_id. Some user filters contained deleted `ir.actions.actions`.
It was then failing when performing the `web_read`.

In this PR I'm adding a test related to the inhexistance of the related `ir.actions.actions`.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2898,6 +2898,13 @@ def disable_invalid_filters(env, verbose=True):
                     )
                     f.active = False
                     break
+        # ACTION_ID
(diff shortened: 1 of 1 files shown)


StefanRijnhart (approved):
Thanks. I see the field `action_id` has been around since 8.0, as has the `exists` method so this should be safe.

pedrobaeza (approved):
Thanks for the extra check

PR #392, opened 2024-12-04
[IMP] copy_columns: Only copy not null values

etobella:
@pedrobaeza @MiquelRForgeFlow @hbrunn @legalsylvain


StefanRijnhart (commented):
Makes sense. Is this just a performance improvement or does it solve errors?
I think I would prefer the commit to include a little more detail. Maybe `[IMP] copy_columns: Only copy not null values`?

etobella:
It should increase the speed. The code itself is right.

StefanRijnhart (approved):
Thanks!

MiquelRForgeFlow (approved):
LGTM

pedrobaeza (approved):

PR #395, opened 2025-01-13
[IMP] rename_models: add condition to check version and existence of ir_property table

xaviedoanhduy:
model `ir.property` was removed in commit https://github.com/odoo/odoo/commit/de302c2 in version `18.0` so if new databases is initialized from 18.0 onwards will not have `ir_property` table then there will be an error exception that `relation ir_property does not exist` when using the `rename_models` function

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -977,19 +977,6 @@ def rename_models(cr, model_spec):
         )
         logged_query(
             cr,
-            """
-            UPDATE ir_property
-            SET res_id = replace(res_id, %(old_string)s, %(new_string)s)
-            WHERE res_id like %(old_pattern)s""",
-            {
-                "old_pattern": "%s,%%" % old,
-                "old_string": "%s," % old,
-                "new_string": "%s," % new,
-            },
-        )
-        # Handle properties that reference to this model
-        logged_query(
-            cr,
             "SELECT id FROM ir_model_fields "
             "WHERE relation = %s AND ttype = 'many2one'",
             (old,),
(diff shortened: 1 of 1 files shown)


jeroenev (approved):
Looks good!

openupgradelib/openupgrade.py
```diff
@@ -975,48 +975,49 @@ def rename_models(cr, model_spec):
...
             )
+            # Handle properties that reference to this model
+            logged_query(
+                cr,
+                "SELECT id FROM ir_model_fields "
+                "WHERE relation = %s AND ttype = 'many2one'",
+                (old,),
+            )
+            field_ids = [x[0] for x in cr.fetchall()]
+            logged_query(
```
pedrobaeza, line 999: This should go outside the `if`.

pedrobaeza (changes requested):
You need to handle the renaming for 18+, changing model inside the JSON field.

openupgradelib/openupgrade.py
```diff
@@ -975,48 +975,49 @@ def rename_models(cr, model_spec):
...
-        logged_query(
-            cr,
-            "UPDATE ir_model_fields SET relation = %s WHERE relation = %s",
-            (
-                new,
-                old,
-            ),
-        )
-        if field_ids:
+        if version_info[0] < 18 and table_exists(cr, "ir_property"):
```
StefanRijnhart, line 978: What would be the reason to check if ir_property exists as a table?

xaviedoanhduy: I'm just denying the case if the `ir_property` table exists and the version is >= 18.0. because in fact if you use a database from 17.0 or earlier to migrate data to version >= 18.0, the `ir_property` table still exists and is just no longer used.

StefanRijnhart: Agreed that even in version 18 no operations should be performed in the table even if it exists! I would say checking the version is enough because we can rely on the table existing in earlier versions.

xaviedoanhduy: thanks, maybe it's redundant. i will update it in next patch

StefanRijnhart: @xaviedoanhduy thanks for your changes. I see you resolved the comment, but I prefer that the comments be resolved by the reviewers on rereview.

openupgradelib/openupgrade.py
xaviedoanhduy: Thank @StefanRijnhart, for the clarification! However, the query to the `ir_model_fields` table seems to be just a `SELECT` statement, right?

Would it be correct for me to move the code handling `ir_property` updates into the  check version `18.0` block while keeping the `SELECT` statement for `ir_model_fields` outside of the condition?

xaviedoanhduy: ah sorry, it seems I didn't notice it. Thank you for pointing out the details. I updated it again and took it out of the check version condition and appears above the condition so that the `field_ids` variable is declared

StefanRijnhart (approved):
Looking good now!

pedrobaeza:
Have you studied the model renaming effect in the new properties stored in JSON in the own table?

StefanRijnhart:
@pedrobaeza It's only ids, not model names in the JSON formatted property fields in Odoo 18.
```
select property_payment_term_id from res_partner where property_payment_term_id is not null limit 1;
 property_payment_term_id 
--------------------------
 {"1": 4}
```

pedrobaeza (approved):
Great then

PR #396, opened 2025-01-20
[ADD] openupgrade_180.convert_company_dependent

hbrunn:
plus a small patch to allow migration signatures (env, version). This needs to go to openupgrade.py because otherwise you can't say

```python
from openupgradelib import openupgrade

openupgrade.migrate()
migration(env, version=None):
```

This is extracted from https://github.com/OCA/OpenUpgrade/pull/4709 where you can see it in use.


openupgradelib/openupgrade.py
```diff
@@ -64,6 +64,15 @@ def __exit__(self, exc_type, exc_value, traceback):
 else:
     import release
 
+try:
+    from odoo.modules.migration import VALID_MIGRATE_PARAMS
```
StefanRijnhart, line 68: Can you document this was introduced in 18?

hbrunn: thanks, done

StefanRijnhart (approved):
Nice!

PR #397, opened 2025-02-14
[IMP] merge_record: keep data from origin

chienandalu:
Sometimes it might be needed to keep the data from the origin record instad of the target one.

A use case might be this one
https://github.com/OCA/product-variant/pull/373 where we are assigning the significant record to a brand new one which data isn't important at that moment.

cc @Tecnativa TT55092

ping @pedrobaeza 

fyi @lmignon 

Pending:

- [ ] Perform some testing


pedrobaeza (approved):

chienandalu:
Maybe a new release is needed to pass tests in https://github.com/OCA/product-variant/pull/373 ?

pedrobaeza:
Well, you need to include the git reference, but let's publish it.

PR #398, opened 2025-02-20
[IMP] merge_records: delete before update

chienandalu:
In some case we might be intereseted in delete the origin records before updating the target with the gathered values to avoid some possible constraints like unique key values.

For example, when updating product_product barcode with a field specification of first_from_origin, the merge will fail as barcode is a unique key for that table.

To change this behavior, a new key `openupgradelib_delete_before_update` can be added to the field_spec dict.

cc @Tecnativa TT55184

please check @pedrobaeza

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -522,7 +522,7 @@ def apply_operations_by_field_type(
 
 
 def _adjust_merged_values_orm(
-    env, model_name, record_ids, target_record_id, field_spec
+    env, model_name, record_ids, target_record_id, field_spec, delete=True
 ):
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Why doing this optional and only in SQL mode and not for both?

chienandalu:
> Why doing this optional

To avoid any issue with existing scripts relaying on the current behavior

> only in SQL mode and not for both?

Added to ORM mode as well.

pedrobaeza:
I still think it shouldn't be an option. The problem is clear, and it will happen always that you have a unique constraint, so it should be applied always. I can't imagine also any side effect.

chienandalu:
Done and tested locally

PR #287, opened 2022-04-03
Add pyproject.toml

sbidoul:
Explicitly declare that setuptools is the build backend
for this project, so it can be installed in environment
where setuptools is not pre-installed.

closes #276

PR #400, opened 2025-02-24
[FIX] Deprecate setup_requires for pyproject.toml

chienandalu:
Follow PEP 517/518 packaging standards to avoid dependency issues.

cc @Tecnativa @josep-tecnativa


marcelsavegnago (approved):

pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[build-system]
+requires = [
+    "setuptools>=61.0.0",
```
StefanRijnhart, line 3: This breaks the installation in Python 2.7. Can you add a lower version specifically for that one?

StefanRijnhart (commented):
Thank! I was hit by this one as well.

hparfr:
does it replace https://github.com/OCA/openupgradelib/pull/287 ?

pyproject.toml
pedrobaeza: Done, @StefanRijnhart should I put `python_version < '2.8'`, or is it OK this way no matter the revision?

sbidoul: Normally `setuptools` should be sufficient, since pip will select the latest version that is compatible with the python version used. Same for `setuptools_scm`.

pyproject.toml
```diff
@@ -0,0 +1,10 @@
...
+    "setuptools_scm>=8.0.0; python_version >= '3.7'",
+    "setuptools_scm<6.0.0; python_version < '3.7'",
+    "wheel"
```
sbidoul, line 6:
```suggestion
```

requirements.txt
```diff
@@ -4,4 +4,5 @@ flake8
 pep8-naming
 lxml<=4.3.4
 psycopg2==2.7.3.1
-setuptools<58.0
+setuptools>=61.0.0 ; python_version > '2.7'
+setuptools<58.0 ; python_version <= '2.7'
```
sbidoul, line 8: Where is this `requirements.txt` file used? 

I find a reference to it in in `.travis.yml` but that is obsolete.

pyproject.toml
```diff
@@ -0,0 +1,10 @@
...
+]
```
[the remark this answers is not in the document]
sbidoul: Setuptools was not using this requirements.txt before, so if no-one knows what where it is used, let's remove it.

requirements.txt
pedrobaeza: I don't know. That's why I'm asking for it, as I don't use pip deployment as you know, and don't know how this should be deployed. Does `pyproject.toml` serve  for all the Python versions, including Python 2.7?

sbidoul: Just remove `requirements.txt` (it was not used before and is not used in this PR either), and `.travis.yml` while you are at it.

> Does pyproject.toml serve for all the Python versions, including Python 2.7?

Yes, pip 20.3.4 (the last version that supports python2) understands pyproject.toml.

pyproject.toml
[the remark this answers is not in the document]
pedrobaeza: OK, removed. Should we put the requirements inside setup.py.

requirements.txt
pedrobaeza: Done.

pedrobaeza:
@sbidoul as you are the expert on "pip" thing, please give your final blessing to merge this and fix the problem in deployments.

sbidoul (approved):
This change looks good, and produces the same openupgradelib wheel as before. So :+1:

PR #401, opened 2025-02-24
[FIX] Don't pin setuptools

pedrobaeza:
For avoiding this incompatibility:

```
ERROR: setuptools==57.5.0 is used in combination with setuptools-scm>=8.x
```

Linked with https://github.com/Tecnativa/doodba-copier-template/issues/520

Being honest, not sure why we need such libraries. I know this is for something related with the publishing to PyPi, but didn't dig into details.

@Tecnativa


pedrobaeza:
Replaced by #400

PR #402, opened 2025-02-28
[ADD] update_module_moved_translations

MiquelRForgeFlow:
Enable rename_xmlids method to change module translations (below v16).

Handles https://github.com/OCA/OpenUpgrade/pull/3882.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -186,6 +186,7 @@ __all__ = [
     "table_exists",
     "update_module_moved_fields",
     "update_module_moved_models",
+    "update_module_moved_translations",
     "update_module_names",
     "add_ir_model_fields",
     "get_legacy_name",
@@ -1287,6 +1288,10 @@ def rename_xmlids(cr, xmlids_spec, allow_merge=False):
         if not old_row:
             logger.info("XMLID %s not found when renaming to %s", old, new)
             continue
+        if old.split(".")[0] != new.split(".")[0]:
+            update_module_moved_translations(
+                cr, old.split(".")[1], old.split(".")[0], new.split(".")[0]
+            )
         if allow_merge:
             cr.execute(get_data_query, tuple(new.split(".")))
             new_row = cr.fetchone()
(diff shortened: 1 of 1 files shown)


StefanRijnhart (approved):

openupgradelib/openupgrade.py
```diff
@@ -2683,6 +2688,57 @@ def rename_property(cr, model, old_name, new_name):
...
+    :param xml_id: a xml record ID
+    :param old_module: previous module of the xml_id
+    :param new_module: new module of the xml_id
+
+    .. deprecated:: 16.0
+       Translations in version 16.0 are in table model, they not depend on module anymore.
+    """
+    if version_info[0] > 15:
+        return
+    logger.info(
```
pedrobaeza, line 2703: I think this is too verbose

openupgradelib/openupgrade.py
```diff
@@ -2683,6 +2688,57 @@ def rename_property(cr, model, old_name, new_name):
...
+        WHERE module = %s AND name = %s
+        """,
+        (
+            old_module,
+            xml_id,
+        ),
+    )
+    row = cr.fetchone()
+    if not row:
+        do_raise("Cannot find xml_id %s.%s." % (old_module, xml_id))
```
pedrobaeza, line 2722: Don't raise on this IMO.

openupgradelib/openupgrade.py
```diff
@@ -2683,6 +2688,57 @@ def rename_property(cr, model, old_name, new_name):
...
+    model = row[0]
+    record_id = row[1]
+    query = """
+    UPDATE ir_translation
```
pedrobaeza, line 2726: Indent these lines for better reading

openupgradelib/openupgrade.py
```diff
@@ -2683,6 +2688,57 @@ def rename_property(cr, model, old_name, new_name):
     )
 
 
+def update_module_moved_translations(cr, xml_id, old_module, new_module):
+    """Cleanup translations of specific records in a module.
```
pedrobaeza, line 2692: Cleanup?

openupgradelib/openupgrade.py
```diff
@@ -2683,6 +2688,52 @@ def rename_property(cr, model, old_name, new_name):
...
+    cr.execute(
+        """
+        SELECT model, res_id
+        FROM ir_model_data
+        WHERE module = %s AND name = %s
+        """,
+        (
+            old_module,
+            xml_id,
+        ),
```
pedrobaeza, line 2713: Please reduce the number of lines for something not adding value:

```suggestion
        (old_module, xml_id),
```

openupgradelib/openupgrade.py
```diff
@@ -2683,6 +2688,52 @@ def rename_property(cr, model, old_name, new_name):
...
+    )
+    row = cr.fetchone()
+    if not row:
+        return
+    model = row[0]
+    record_id = row[1]
```
pedrobaeza, line 2719: These 2 variables are used only once. Apply them directly over the query for reducing the number of lines for this tiny method. The readability is barely affected.

openupgradelib/openupgrade.py
```diff
@@ -2683,6 +2688,52 @@ def rename_property(cr, model, old_name, new_name):
...
+    logged_query(
+        cr,
+        query,
+        (
+            new_module,
+            old_module,
+            model + ",%",
+            record_id,
+        ),
+    )
```
pedrobaeza, line 2734:
```suggestion
    logged_query(
        cr, query, (new_module, old_module, row[0] + ",%", row[1])
    )
```

pedrobaeza (commented):
I'm wondering if this also affects when a field moves from one module to another. Are their translations affected?

pedrobaeza:
What about my question?

> I'm wondering if this also affects when a field moves from one module to another. Are their translations affected?

MiquelRForgeFlow:
> I'm wondering if this also affects when a field moves from one module to another. Are their translations affected?

Yes, but as far as I see in the code, the translation issue is already handled in `update_module_moved_fields`, `update_module_moved_models` and `update_module_names` methods, so it seems to me there is no need to do more changes. Am I missing something?

pedrobaeza:
The problem is that we are not calling that methods since a lot of versions because it was "not needed". Luckily, since 16, there shouldn't be such problem.

pedrobaeza (approved):

PR #403, opened 2025-02-28
[FIX] rename_models: don't call get_model2table

MiquelRForgeFlow:
The names in ir_model_data use the model converted name instead of table name.


pedrobaeza (changes requested):
I don't get the problem

MiquelRForgeFlow:
This is from v16:

Selection_4356

You can see that is shows `model_mailing_contact_subscription` instead of `model_mailing_contact_list_rel`. The same in base module, where `base.model_ir_actions_actions` exists instead of `base.model_ir_actions`. Same argument for the model data of fields.

pedrobaeza (approved):
OK, now I understand. Good catch!

PR #407, opened 2025-03-19
[IMP] fill_analytic_distribution: Make tags optional

pedrobaeza:
In some cases, there's no analytic tags field, and you only want to convert the previous analytic account field to analytic distribution. As we can reuse the same method for that, let's do it.

### DIFF
--- a/openupgradelib/openupgrade_160.py
+++ b/openupgradelib/openupgrade_160.py
@@ -442,17 +442,18 @@ def _convert_field_bootstrap_4to5_sql(cr, table, field, ids=None):
 def fill_analytic_distribution(
     env,
     table,
-    m2m_rel,
-    m2m_column1,
+    m2m_rel=False,
+    m2m_column1=False,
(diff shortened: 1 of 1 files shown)


openupgradelib/openupgrade_160.py
```diff
@@ -466,6 +467,24 @@ def fill_analytic_distribution(
         env.cr,
         f"ALTER TABLE {table} ADD COLUMN IF NOT EXISTS {column} jsonb",
     )
+    query_union = ""
+    if m2m_rel:
```
MiquelRForgeFlow, line 471:
```suggestion
    if m2m_rel and m2m_column1:
```

pedrobaeza: Well, it's your responsibility to call correctly the method if you want to process analytic tags. The other parameter is the control one, but OK, modified to your taste.

MiquelRForgeFlow (approved):

Issue #409, opened 2025-03-27
OpenUpgrade error during migration


patrickt-oforce:
After odoo merge PRs
- 16.0 https://github.com/odoo/odoo/pull/202014
- 17.0 https://github.com/odoo/odoo/pull/202583
 
moving the migration process in a dedicated method, the key 'pkg' is not available anymore in `frame.locals`, i think the correct refernce is `addon` parameter of method.
So when try to access that value in https://github.com/OCA/openupgradelib/blob/19ba25c49f15225a428d872d2e3710278928c072/openupgradelib/openupgrade.py#L2309 migration process break, because the key is not existent anymore.

The issue opend on Odoo is: https://github.com/odoo/odoo/issues/203731

Related issue, https://github.com/OCA/l10n-spain/issues/4092

patrickt-oforce:
@pedrobaeza

pedrobaeza:
Fixed in #410

PR #410, opened 2025-03-27
[FIX] openupgrade: fix module name retrieval

patrickt-oforce:
Fix the error indicated in https://github.com/OCA/l10n-spain/issues/4092 after https://github.com/odoo/odoo/pull/202014 merge, and potentially close the issue #409

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2306,7 +2306,15 @@ def migrate(no_version=False, use_env=None, uid=None, context=None):
                 try:
                     frame = inspect.getargvalues(inspect.stack()[1][0])
                     stage = frame.locals["stage"]
-                    module = frame.locals["pkg"].name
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Diff review. 

Not expert here, but looks correct.

thanks for your work.

patrickt-oforce:
> I'm still perplexed that Odo has done this change...

The PR of Odoo was originated by the need to use their upgrade library outside of migration process, with this PRs they allow to use the method merge_modules and rename_modules of their library before base is updated

pedrobaeza:
Yeah, I mean that breaking change... It's true that we were using an "internal variable" and very specific Python code outside Odoo modules, but their stability policy sometimes is not strict where it should be, and it's the excuse other times for not accepting clear patches. Let's move on and thanks again for finding the issue.

PR #412, opened 2025-04-07
[ADD] add_column: split from add_fields

MiquelRForgeFlow:
Sometimes, you only want to precreate the column, not the `ir.model.fields` entry, `ir.model.data` entry, etc. In fact, since v12, `add_fields` is less needed as Odoo always add the XML-ID entry, so you may be interested only in the SQL column.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1945,6 +1945,7 @@ def get_field2column_type(field_type, translatable=False):
         "selection": "varchar",  # Can be sometimes integer. Force it manually
         "text": "text",
         "serialized": "text",
+        "json": "jsonb",
     }
     if version_info[0] > 15 and field_type in ["char", "text", "html"] and translatable:
         return "jsonb"
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -163,6 +163,7 @@ __all__ = [
     "migrate",
     "logging",
     "load_data",
+    "add_columns",
     "add_fields",
     "copy_columns",
     "copy_fields_multilang",
@@ -3029,6 +3030,64 @@ def disable_invalid_filters(env, verbose=True):
             f.active = False
 
 
+def add_columns(env, field_spec):
+    """This method adds a new column.
+
+    It's intended for being run in pre-migration scripts for pre-populating
+    columns to avoid the execution of new stored computed fields.
+
+    :param: field_spec: List of tuples with the following expected elements
+      for each tuple:
+
+      * model name
+      * field name
+      * field type: binary, boolean, char, date, datetime, float, html,
+        integer, many2many, many2one, many2one_reference, monetary, one2many,
+        reference, selection, text, serialized. The list can vary depending on
+        Odoo version or custom added field types.
+      * (optional) initialization value: if included in the tuple, it is set
+        in the column for existing records.
+      * (optional) SQL table name
+      * (optional) SQL field type: If the field type is custom or if it's one
+        of the special cases (see get_field2column_type), you need to indicate
+        here the SQL type to use (from the valid PostgreSQL types):
+        https://www.postgresql.org/docs/9.6/static/datatype.html
+    """
+    cr = env.cr
+    for vals in field_spec:
+        model_name = vals[0]
+        field_name = vals[1]
+        field_type = vals[2]
+        init_value = vals[3] if len(vals) > 3 else False
+        table_name = vals[4] if len(vals) > 4 else False
+        sql_type = vals[5] if len(vals) > 5 else False
+        sql_type = sql_type or get_field2column_type(field_type)
+        if not table_name:
+            try:
+                table_name = env[model_name]._table
+            except KeyError:
+                table_name = get_model2table(model_name)
+        if sql_type and not column_exists(cr, table_name, field_name):
+            query = sql.SQL("ALTER TABLE {} ADD COLUMN {} {}").format(
+                sql.Identifier(table_name),
+                sql.Identifier(field_name),
+                sql.SQL(sql_type),
+            )
+            args = []
+            if init_value:
+                query += sql.SQL(" DEFAULT %s")
+                args.append(init_value)
+            logged_query(cr, query, args)
+            if init_value:
+                logged_query(
+                    cr,
+                    sql.SQL("ALTER TABLE {} ALTER COLUMN {} DROP DEFAULT").format(
+                        sql.Identifier(table_name),
+                        sql.Identifier(field_name),
+                    ),
+                )
+
+
 def add_fields(env, field_spec):
     """This method adds all the needed stuff for having a new field populated
     in the DB (SQL column, ir.model.fields entry, ir.model.data entry...).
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Instead of that, you can reuse the existing method and add an extra keyword argument for not executing whole method.

MiquelRForgeFlow:
I know, but I think my way is cleaner. The method `add_fields` expects to add a field as the name suggests, not only the column. To me, calling `add_fields` with a lot of parameters only to add a column is confusing.

openupgradelib/openupgrade.py
```diff
@@ -2970,36 +2971,26 @@ def format_message(f):
...
 
 
-def add_fields(env, field_spec):
-    """This method adds all the needed stuff for having a new field populated
-    in the DB (SQL column, ir.model.fields entry, ir.model.data entry...).
+def add_columns(cr, field_spec):
+    """This method adds a new column.
 
     It's intended for being run in pre-migration scripts for pre-populating
     fields that are going to be declared later in the module.
```
pedrobaeza, line 2978: Then this docstring is not correct.

MiquelRForgeFlow: Changed :)

openupgradelib/openupgrade.py
```diff
@@ -2970,36 +2971,26 @@ def format_message(f):
...
-    this method for consistency and for avoiding to know the internal PG
-    column type.
+    columns to avoid the execution of new stored computed fields.
 
     :param: field_spec: List of tuples with the following expected elements
       for each tuple:
 
       * field name
-      * model name
       * SQL table name: Put `False` if the model is already loaded in the
```
pedrobaeza, line 2984: The `False` part text is not correct either.

openupgradelib/openupgrade.py
```diff
@@ -2970,36 +2971,26 @@ def format_message(f):
...
         registry and thus the SQL table name can be obtained that way.
       * field type: binary, boolean, char, date, datetime, float, html,
```
pedrobaeza, line 2986: This is not covering the new translatable fields that are JSONB.

MiquelRForgeFlow: This should be done in another PR, as you always say.

MiquelRForgeFlow: BTW, I don't how would should handle it, as there is not a field type, only the translate attribute that makes use of jsonb sql type.

MiquelRForgeFlow: Ok, I added a note to pass the sql_type explicitly in case of translatable. Also moved back the mapping to add_fields.

pedrobaeza:
That's not what I said to do. It's losing the charming of this method, which is to auto-convert the Odoo types to PG types, to abstract them. This way, it's even simpler to call `logged_query(env.cr, "ALTER TABLE ADD ...")`.

Please revert back to what I said.

MiquelRForgeFlow:
Well, I would like a second opinion here. The origin of this PR was due to a comment of @hbrunn, which asked if we could have the column creation (and dropping default) in a separated method. And I just wanted to keep it simple (by simple, I mean, less parameters to pass). If you find simpler to write those two queries explicitly again and again in the migration scripts, feel free to do it that way.

pedrobaeza:
What I want is a useful method, not one that requires to know the internal DB types.

StefanRijnhart (commented):
I like having it as a separate method, with the ability to use it with only ORM terminology, passing a model name and an ORM field type as a minimum.

Given that there is no grouping in the current `field_spec`, it may be less useful to allow for multiple items. By implementing `add_column` rather than `add_columns` we can start using keyword parameters, making the calls less convoluted.

Or we could consider defining a `collections.namedtuple`

```
NewColumn = namedtuple(
    "NewColumn",
    ["model_name", "field_name", "field_type", "init_value", "table_name", "sql_type"],
)

def add_columns(cr, *new_columns):
```

MiquelRForgeFlow:
Changed `add_columns` -> `add_column`, it got simpler ✅

> I like having it as a separate method, with the ability to use it with only ORM terminology, passing a model name and an ORM field type as a minimum.

I am not convinced haha

openupgradelib/openupgrade.py
```diff
@@ -2970,6 +2971,39 @@ def format_message(f):
             f.active = False
 
 
+def add_column(cr, field_name, table_name, sql_type, init_value=False):
```
pedrobaeza, line 2974:
```suggestion
def add_column(cr, field_name, table_name, sql_type, init_value=False, translatable=False):
```

and switch the type for char accordingly.

MiquelRForgeFlow: This would make sense if I was passing the field_type, but at the moment I am passing only the sql_type, then the person who call this method should put "jsonb" in sql_type directly instead of using this translatable flag.

pedrobaeza: Then I insist this is not what we are looking for. We want the field_type, not the sql_type. Stefan agrees with me.

StefanRijnhart: And analogous to that, maybe the model name?

MiquelRForgeFlow: @hbrunn but this `add_columns` should be called with `field_type` or `sql_type`? with `model_name` or `table_name`?

hbrunn: fine with me

openupgradelib/openupgrade.py
```diff
@@ -1899,6 +1901,39 @@ def get_model2table(model):
     return model2table.get(model, model.replace(".", "_"))
 
 
+def get_field2column_type(field_type):
```
pedrobaeza, line 1904: This is still missing `translate=False` keyword argument.

MiquelRForgeFlow: It's not missing. It's a parameter of `add_columns`.

openupgradelib/openupgrade.py
```diff
@@ -2982,6 +3017,71 @@ def format_message(f):
...
+        https://www.postgresql.org/docs/9.6/static/datatype.html
+      * (optional) translatable: From >=v16, if field is translatable then
+        SQL field type is changed to 'jsonb'.
+    """
+    cr = env.cr
+    for vals in field_spec:
+        model_name = vals[0]
+        field_name = vals[1]
+        field_type = vals[2]
+        init_value = vals.get(3, False)
```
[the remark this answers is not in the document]
MiquelRForgeFlow: Oh, my bad.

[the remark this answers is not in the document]
MiquelRForgeFlow: Fixed.

StefanRijnhart (approved):

openupgradelib/openupgrade.py
pedrobaeza: But semantically it should be here, as this method intents to return the absolute SQL type for any field type.

MiquelRForgeFlow: Done.

openupgradelib/openupgrade.py
```diff
@@ -1912,6 +1914,44 @@ def get_model2table(model):
...
+        "many2one_reference": "int4",
+        "monetary": "numeric",
+        "one2many": False,  # No need to create SQL column
+        "reference": "varchar",
+        "selection": "varchar",  # Can be sometimes integer. Force it manually
+        "text": "text",
+        "serialized": "text",
+        "json": "jsonb",
+    }
+    if version_info[0] > 15 and translatable:
```
pedrobaeza, line 1950:
```suggestion
    if version_info[0] > 15 and field_type in ... and translatable:
```

MiquelRForgeFlow: Done. Yes, only those three are. I checked.

pedrobaeza (approved):
Thanks for the changes.

Issue #421, opened 2025-05-20
[ODOO.SH] Build is considered failed because of `logger.error`


emi-odoo:
In function `migrate` we log a message in case a package is missing: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L2386

The type of log is `error`. SH's behavior considers builds failed if a log of type error is logged

## To Reproduce
**Affected versions**:

1. Create a SH project
2. Install an OCA module that uses the `migrate` decorator (https://github.com/OCA/l10n-italy/blob/16.0/l10n_it_fatturapa/migrations/16.0.1.1.1/post-migrate.py)
3. Migration is run -> message is logged
4. If the migration is running on SH, the build rolls back

PR #422, opened 2025-05-20
[FIX] prevent build rollback on odoo.sh

emi-odoo:
Odoo.sh consider builds as failed if an error is logged (regardless of the fact that we continue with the migration script or not). Following odoo PR #202014, this will happen more frequently in the `migrate` function as `pkg` is not available anymore

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -2383,7 +2383,7 @@ def migrate(no_version=False, use_env=None, uid=None, context=None):
                     # Python3: fetch pyfile from locals, not fp
                     filename = frame.locals.get("pyfile") or frame.locals["fp"].name
                 except Exception as exc:
-                    logger.error(
+                    logger.warning(
                         "'migrate' decorator: failed to inspect the frame above: %s",
                         exc,
                     )


emi-odoo:
Linked to issue: https://github.com/OCA/openupgradelib/issues/421

StefanRijnhart (approved):
Thanks! Please don't create a separate issue if you are going to propose a fix as well. You can put all the info in the PR.

emi-odoo:
@StefanRijnhart indeed! I was not sure I would have been able to create the fix too

pedrobaeza:
But this was already fixed putting .get before. Are you trying with latest library version?

emi-odoo:
@pedrobaeza hi! Not sure which version is being used, I was helping another partner! I will check that :)

StefanRijnhart:
Ah yes, https://github.com/OCA/openupgradelib/commit/b1ae535f7520acb1a4bb2afbdb983f1b24e96781. Retracting my approval for now.

emi-odoo:
@pedrobaeza @StefanRijnhart thanks guys! I rushed to this solution! I will investigate some more and come back to you

emi-odoo:
Checked the issue, the partner was using an old version of openupgradelib. Solution -> manually update it

PR #425, opened 2025-10-01
[FIX] openupgrade_merge_records: safely merge jsonb and serialized fields

xaviedoanhduy:
when merging records using `openupgrade_merge_records.merge_records()`,
the helper `apply_operations_by_field_type` attempted to apply the `|=` operator directly on string values. however jsonb fields may be wrapped types (e.g. psycopg2.extras.Json), causing:
` TypeError: unsupported operand type(s) for |=: 'dict' and 'str'`

maybe not as expected the result returned like `method="sql"` ([ref](https://github.com/OCA/openupgradelib/blob/8a86014577d567883f73e36eea436af67096078d/openupgradelib/openupgrade_merge_records.py#L700-L709)), will return dict value like `{"en_US": "Awaiting"}`. but unfortunately in `method="orm"` even though `name` field in database is jsonb but in ORM level it is still `Char` type and return a string like "Awaiting" so the error occurred.

this proposed fix will convert string value to dict based on current language (mostly `en_US`). In case field type is `Serialized`, the old logic remains the same because they are mostly `dicts`. And in python versions like `3.8` it seems that the `|=` operator is not supported yet so I replaced it with the `.update()` function

on this occasion also corrected the warning sush as: `UserWarning: unsupported operand type(s) for "==": 'res.company()' == '1'` in the second commit https://github.com/OCA/openupgradelib/commit/96be2e9aa29525d3f22bbee776a7129046af8764

### DIFF
--- a/openupgradelib/openupgrade_merge_records.py
+++ b/openupgradelib/openupgrade_merge_records.py
@@ -654,13 +654,21 @@ def _adjust_merged_values_orm(
         return
     # Curate values that haven't changed
     new_vals = {}
-    for f in vals:
-        if model._fields[f].type != "many2many":
-            if vals[f] != getattr(target_record, f):
-                new_vals[f] = vals[f]
+    for f, new_val in vals.items():
+        field = model._fields[f]
+        current_val = getattr(target_record, f)
+
+        if field.type == "many2many":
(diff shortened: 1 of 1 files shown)


MiquelRForgeFlow (approved):
I forgot 😅

PR #427, opened 2025-10-31
[IMP] load_data: support v19 upgrade_path

hbrunn:
in v18 and before, upgrade_path was a comma separated [string](https://github.com/OCA/OCB/blob/18.0/odoo/tools/config.py#L123), while it's parsed as a [list](https://github.com/OCA/OCB/blob/19.0/odoo/tools/config.py#L243) in v19


pedrobaeza:
Waiting for full v19 check to make a new release.

hbrunn:
I'm nearly done with the migration script of the base module, up to now that's the only issue I encountered. But I agree there's no haste to release this

PR #448, opened 2026-05-08
[FIX] openupgrade: eager-import odoo submodules for Odoo 19 compatibility

Paerrin:
In Odoo <= 18, `odoo/__init__.py` eager-imported submodules like `fields`, `api`, `tools`, `exceptions`, etc., so `import odoo as core` was enough to access them as `core.fields.Many2many`, `core.api`, ...

In Odoo 19, `odoo/__init__.py` no longer does these eager imports. As a result, openupgradelib fails to import under Odoo 19 with:

    AttributeError: module 'odoo' has no attribute 'fields'
    AttributeError: module 'odoo' has no attribute 'api'

at module load time (lines that read `core.fields.Many2many` and `api = core.api`). This breaks the OCA/OpenUpgrade 19.0 migration path: `openupgrade_framework` is loaded as a server-wide module, which imports openupgradelib, which fails before any migration script can run.

Eager-import the submodules referenced at module level right after `import odoo as core`. Each import is wrapped in try/except so older Odoo versions with a slightly different module layout still work.

Verified end-to-end: Odoo 18.0.20250924 → 19.0.20260508 OpenUpgrade migration completes cleanly (`Modules loaded`, registry loaded in 414s, zero ERROR/CRITICAL) on a vanilla Community install with this patch.

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -18,7 +18,7 @@ try:
 except ImportError:
     from io import StringIO
 
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 
 try:
     from psycopg2 import IntegrityError, ProgrammingError, errorcodes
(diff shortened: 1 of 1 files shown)


tate11:
Tested on Odoo 18 → 19 migration (v18.0 to v19.0). This fix resolves the AttributeError: module 'odoo' has no attribute 'fields' issue we encountered when loading openupgrade_framework as a server-wide module during migration. After applying this patch, Many2many, One2many, api, and other submodule references work correctly at import time. The full migration of ~157 modules completed successfully with OpenUpgrade 19.0.

    LGTM, this is ready to merge.

hbrunn (changes requested):
I don't see what problem this solves.

openupgrade_framework works just fine without this

PR #432, opened 2026-01-20
[UPD] HTML documentation

CesarSant2000:

PR #453, opened 2026-05-30
[FIX] docs: add 14.0 + 18.0 method sections to API.rst

dnplkndll:
## What
Add the missing **14.0** and **18.0** method sections to `docsource/API.rst`. Both modules (`openupgrade_140.py`, `openupgrade_180.py`) exist but were never wired into the API docs, so the published page stopped at 16.0.

## Also
Fixes an RST indentation warning in `convert_field_html_string_13to14`'s docstring (missing blank line before the `:param` list) that `sphinx -W` surfaces once the 14.0 module is documented.

Source-only — `docs/` is regenerated by `documentation-commit.yml` on push. Verified: `sphinx -W` builds clean; both `convert_html_string_13to14` and `convert_company_dependent` now render.

Supersedes #432.

### DIFF
--- a/docsource/API.rst
+++ b/docsource/API.rst
@@ -71,6 +71,16 @@ release.
 .. automodule:: openupgradelib.openupgrade_130
    :members:
 
+Methods for OpenUpgrade 14.0
+----------------------------
+
+The following specific methods for 14.0 are available. These have been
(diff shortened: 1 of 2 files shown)


dnplkndll:
@pedrobaeza

pedrobaeza (approved):

PR #459, opened 2026-07-01
[IMP] Load additional update_module_names spec from environment

david-banon-tecnativa:
Related to https://github.com/OCA/OpenUpgrade/pull/5765

@pedrobaeza @hbrunn 
@tecnativa TT63118

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -1879,14 +1879,32 @@ def logged_query(cr, query, args=None, skip_no_result=False):
     return cr.rowcount
 
 
-def update_module_names(cr, namespec, merge_modules=False):
+def update_module_names(cr, namespec, merge_modules=False, environment_namespec=False):
     """Deal with changed module names, making all the needed changes on the
     related tables, like XML-IDs, translations, and so on.
 
     :param namespec: list of tuples of (old name, new name)
     :param merge_modules: Specify if the operation should be a merge instead
         of just a renaming.
+    :param environment_namespec: If ``True``, parse JSON from the appropriate
+        environment variable (``OPENUPGRADE_MERGED_MODULES`` for merges or
+        ``OPENUPGRADE_RENAMED_MODULES`` for renames) and merge the resulting
+        mapping into ``namespec``.
+
     """
+    if environment_namespec:
+        if merge_modules:
+            env_var = "OPENUPGRADE_MERGED_MODULES"
+        else:
+            env_var = "OPENUPGRADE_RENAMED_MODULES"
+        try:
+            env_namespec = json.loads(os.environ.get(env_var, "{}"))
+        except ValueError:
+            raise ValueError(
+                "Environment variable %s doesn't contain valid json" % env_var
+            )
+        namespec = list(namespec)
+        namespec.extend(list(env_namespec.items()))
     for (old_name, new_name) in namespec:
         query = "SELECT id FROM ir_module_module WHERE name = %s"
         cr.execute(query, [new_name])
--- a/tests/test_openupgradelib.py
+++ b/tests/test_openupgradelib.py
@@ -1,3 +1,4 @@
+import json
 import os
 import unittest
 from io import StringIO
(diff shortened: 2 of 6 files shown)


pedrobaeza:
This is not the idea, but hard-coding here the environment variables for reducing the complexity on the OpenUpgrade side not having to copy from other version the environment variables names, or have the risk of some typos

openupgradelib/openupgrade.py
```diff
@@ -1879,14 +1879,22 @@ def logged_query(cr, query, args=None, skip_no_result=False):
...
 
-def update_module_names(cr, namespec, merge_modules=False):
+def update_module_names(cr, namespec, merge_modules=False, environment_namespec=False):
     """Deal with changed module names, making all the needed changes on the
     related tables, like XML-IDs, translations, and so on.
 
     :param namespec: list of tuples of (old name, new name)
     :param merge_modules: Specify if the operation should be a merge instead
         of just a renaming.
+    :param environment_namespec: If true, will load extra namespec from environment.
```
pedrobaeza: Indicate here the environment variables names.

.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ exclude: |
     # You don't usually want a bot to modify your legal texts
     (LICENSE.*|COPYING.*)
 default_language_version:
-    python: python3
+    python: python3.8
```
pedrobaeza: Why touching this?

david-banon-tecnativa: Pre-commit fails when running on python version 3.12+, so i think its better to pin it to the same version CI uses.

pedrobaeza: But fails where? In your local environment? In the rest of the pre-commmit config files, there's no such restriction. Example: https://github.com/OCA/l10n-spain/blob/19.0/.pre-commit-config.yaml#L24

david-banon-tecnativa: Yes, but openupgradelib pre-commit runs on ci in python 3.8, and the flake8 hook fails on python 3.12+
https://pastebin.com/VmRxyKw1
Its not pre-commit itself, but the config used in this repo that fails on python 3.12.

pedrobaeza: Then maybe the pre-commit config should be aligned with the rest of the OCA repos? @hbrunn you worked on this. Is there any reason to not be aligned?

david-banon-tecnativa: It is a copier update away from being aligned since we merged this https://github.com/OCA/oca-addons-repo-template/pull/326, every repo will pin pre-commit python to the same version used on CI if it gets updated.

pedrobaeza: Uhm, OK, although I don't like that too much, as it's not going to work in most modern distributions, as Python 3.8 is not included.

pedrobaeza (approved):

.pre-commit-config.yaml
hbrunn: no, we really don't want this

hbrunn: note there's no copier involved in this repo

david-banon-tecnativa: On newer versions you can run it with `pipx --python 3.8 run pre-commit run`.

Or install pre-commit with [pre-commit-uv ](https://github.com/tox-dev/pre-commit-uv/)which will use [UV](https://docs.astral.sh/uv/) under the hood and will download & use the correct python version for every repo.
```
pipx install pre-commit
pipx inject pre-commit pre-commit-uv
```
Or just use UV
`uv tool install pre-commit --with pre-commit-uv --force-reinstall`

openupgradelib/openupgrade.py
```diff
@@ -1879,14 +1879,26 @@ def logged_query(cr, query, args=None, skip_no_result=False):
...
+        ``OPENUPGRADE_RENAMED_MODULES`` for renames) and merge the resulting
+        mapping into ``namespec``.
+
     """
+    if environment_namespec:
+        if merge_modules:
+            env_var = "OPENUPGRADE_MERGED_MODULES"
+        else:
+            env_var = "OPENUPGRADE_RENAMED_MODULES"
+        env_namespec = json.loads(os.environ.get(env_var, "{}"))
```
hbrunn: please handle invalid json here and raise a message containing the env var name

hbrunn (changes requested):
please use this opportunity to add a test for this function

.pre-commit-config.yaml
pedrobaeza: @hbrunn says to not do the change here, and about your instructions, if the Python executable is not present in the host, I doubt they will work, but in any case, they seems not very straight-forward...

openupgradelib/openupgrade.py
david-banon-tecnativa: Added a test for update_module_names and an error message if the ENV doesn't contain valid JSON

.pre-commit-config.yaml
david-banon-tecnativa: Removed the python version pin

tests/test_openupgradelib.py
```diff
@@ -206,6 +207,48 @@ def test_disable_invalid_filters(self):
...
+        renamed_module = Mod.search([("name", "=", new), ("state", "=", "installed")])
+        self.assertFalse(dummy_module)
+        self.assertTrue(renamed_module)
+        self.env.cr.execute(
+            "SELECT name FROM ir_module_module WHERE id = %s", (renamed_module.id,)
+        )
+        name = self.env.cr.fetchone()[0]
+        self.assertEqual(name, new)
+        # rename dummy module back into original name using namespec
+        os.environ["OPENUPGRADE_RENAMED_MODULES"] = json.dumps({new: old})
```
hbrunn: please use mock.patch.dict here, this might have unpleasant side effects otherwise some time in the future

hbrunn: and add a line where you set the variable to invalid json

david-banon-tecnativa: Replaced direct environment edition with mock.patch.dict and added test that invalid JSON raises an error

tests/test_openupgradelib.py
```diff
@@ -206,6 +207,64 @@ def test_disable_invalid_filters(self):
...
+        dummy_module = Mod.search([("name", "=", old), ("state", "=", "installed")])
+        renamed_module = Mod.search([("name", "=", new), ("state", "=", "installed")])
+        self.assertFalse(dummy_module)
+        self.assertTrue(renamed_module)
+        self.env.cr.execute(
+            "SELECT name FROM ir_module_module WHERE id = %s", (renamed_module.id,)
+        )
+        name = self.env.cr.fetchone()[0]
+        self.assertEqual(name, new)
+        # ENsure invalid JSON raises an error
```
hbrunn:
```suggestion
        # Ensure invalid JSON raises an error
```

tests/test_openupgradelib.py
```diff
@@ -206,6 +207,64 @@ def test_disable_invalid_filters(self):
...
+            "SELECT name FROM ir_module_module WHERE id = %s", (dummy_module.id,)
+        )
+        name = self.env.cr.fetchone()[0]
+        self.assertEqual(name, old)
+        # Merge dummy module into base
+        openupgrade.update_module_names(
+            self.cr, {old: "base"}.items(), merge_modules=True
+        )
+        dummy_module = Mod.search([("name", "=", new), ("state", "=", "installed")])
+        renamed_module = Mod.search([("name", "=", old), ("state", "=", "installed")])
```
hbrunn:
```suggestion
        dummy_module = Mod.search([("name", "=", new)])
        renamed_module = Mod.search([("name", "=", old)])
```

hbrunn: actually, I wonder why you search for state above at all? and then via sql? shouldn't it suffice to prove the point with the two searches I propose?

david-banon-tecnativa: The state search is unnecessary, removed.
The sql search was because at some point the following code was failing at the assertEqual due to the ORM cache and didn't know if invalidate_cache worked in all versions, but i agree it is also unnecessary with the 2 searches, also removed.
```python
        dummy_module = Mod.search([("name", "=", old)])
        renamed_module = Mod.search([("name", "=", new)])
        self.assertFalse(dummy_module)
        self.assertTrue(renamed_module)
        self.assertEqual(renamed_module.name, new)
```

hbrunn (commented):
thanks!

pedrobaeza:
@hbrunn is it OK for you then?

hbrunn (approved):
sure

david-banon-tecnativa:
Can we release a new openupgradelib version to include this change?

pedrobaeza:
Done, although it's not really needed installing from GitHub.

https://github.com/OCA/openupgradelib/releases/tag/3.13.5

PR #460, opened 2026-07-10
[IMP] check_valid_filters: Use search_count and add missing context fields

david-banon-tecnativa:
Some filters on non optimized models (sale.report.delivered) were taking a really long time to check because search is fetching all the fields, I think we can safely assume that if it doesn't crash with search_count it also won't crash once migrated.

Also, pivot views use different context keys that were not being checked.

@pedrobaeza @sergio-teruel 
TT62385 @tecnativa

### DIFF
--- a/openupgradelib/openupgrade.py
+++ b/openupgradelib/openupgrade.py
@@ -3192,10 +3192,17 @@ def disable_invalid_filters(env, verbose=True):
             with savepoint(env.cr):
                 # Strange artifact found in a filter
                 domain = f.domain.replace("%%", "%")
-                model.search(
-                    safe_eval(domain, globaldict),
-                    limit=1,
-                )
+                if version_info[0] < 16:
+                    model.search(
+                        safe_eval(domain, globaldict),
+                        limit=1,
+                        count=True,
+                    )
+                else:
+                    model.search_count(
+                        safe_eval(domain, globaldict),
+                        limit=1,
+                    )
         except Exception as e:
             logger.warning(
                 format_message(f) + "as it contains an invalid domain %s. Detail: %s",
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
`Some filters on optimized models` > `Some filters on non optimized models`

sergio-teruel (approved):

openupgradelib/openupgrade.py
```diff
@@ -3192,7 +3192,7 @@ def format_message(f):
             with savepoint(env.cr):
                 # Strange artifact found in a filter
                 domain = f.domain.replace("%%", "%")
-                model.search(
+                model.search_count(
                     safe_eval(domain, globaldict),
                     limit=1,
                 )
```
MiquelRForgeFlow:
```suggestion
                if version_info[0] < 16:
                    model.search(
                        safe_eval(domain, globaldict),
                        limit=1,
                    )
                else:
                    model.search_count(
                        safe_eval(domain, globaldict),
                        limit=1,
                    )
```
`limit` in `search_count` was implemented in v16.

david-banon-tecnativa: Done, thanks for the review.

MiquelRForgeFlow (approved):

openupgradelib/openupgrade.py
```diff
@@ -3192,10 +3192,16 @@ def format_message(f):
             with savepoint(env.cr):
                 # Strange artifact found in a filter
                 domain = f.domain.replace("%%", "%")
-                model.search(
-                    safe_eval(domain, globaldict),
-                    limit=1,
-                )
+                if version_info[0] < 16:
+                    model.search(
```
pedrobaeza, line 3196: Isn't still better to use `search_count` (without limit) instead of `search` in this case?

david-banon-tecnativa: We can't really know without trying it on real databases, however i've looked at the search method  on v15 all the way down to v6.0, and it supports a count param that doesn't run .browse(ids) and so doesn't fetch all the fields.

PR #464, opened 2026-08-26
fix bad logging within map_values

Hugo-Trentesaux:
When model is an empty recordset (`env["res.partner"]`), the condition is falsy, where we want it to be truthy instead (model._table is actually defined).


pedrobaeza (approved):

pedrobaeza:
Please rephrase the commit message to `map_values: detect model absence correctly`

Hugo-Trentesaux:
```
git commit --amend -m "map_values: detect model absence correctly"
```

hbrunn (approved):

MiquelRForgeFlow:
unittests fails

hbrunn:
> unittests fails

that's github's shaky infrastructure, fixed in https://github.com/OCA/openupgradelib/pull/467. merging manually as ocabot didn't seem to do its job here

PR #465, opened 2026-09-16
[FIX] merge_records: data loss in m2m relations

hbrunn:
the existing fallback code doesn't work for cases like:

company1 with id 1, company2 with id 2, user1 with id 42, user2 with id 43

user1.company_ids = [1, 2]
user2.company_ids = [2]

when we now merge company2 into company1, user2 will end up without company because the fallback code for uniqueness errors will run

```sql
update res_company_users_rel set cid=1 where cid=2
```

which fails because of user1's entry, and afterwards all tuples with cid=2 are deleted, hence no companies for user2

what we actually want is to run

```sql
update res_company_users_rel set cid=1 where cid=2 and user_id=42 -- this fails, expected
update res_company_users_rel set cid=1 where cid=2 and user_id=43 -- this goes through
```

i took the liberty to change some of the names to align better with what's going on, it took me quite a while to wrap my head around this one.


openupgradelib/openupgrade_merge_records.py
```diff
@@ -81,31 +81,46 @@ def _change_foreign_key_refs(
...
-            for row in list(set([x[0] for x in env.cr.fetchall()])):
+            for id_or_other_value, record_id in env.cr.fetchall():
                 env.cr.execute("SAVEPOINT sp2")
                 try:
                     logged_query(
                         env.cr,
-                        """UPDATE %(table)s
-                        SET "%(column)s" = %(target_record_id)s
-                        WHERE %(target_column)s = %(record_id)s""",
+                        f"""UPDATE {table}
```
pedrobaeza: You shouldn't change this to f-strings for several reasons:

- Compatibility with older versions.
- Let psycopg2 to stringify the field names.

pedrobaeza (commented):
I don't think the nomenclature is incorrect, talking about target_column for knowing where is the source and the destination.

openupgradelib/openupgrade_merge_records.py
hbrunn: good point about old python versions, psycopg2 however doesn't do anything with parameters passed AsIs

hbrunn:
the names are horrible, because `target_column` implies this column has something to do with target_ids, and it decidedly doesn't. what we're targeting here is *column*. and `row` is bad because it it's either the row's id or the id of some other record in the many2many case.

celm1990 (approved):
I found the same issue and was about to create a PR, but I just found this. I tested it, and it works for me. LGTM.

pedrobaeza (approved):

pedrobaeza:
@hbrunn can you resolve conflicts?

hbrunn:
@pedrobaeza done

MiquelRForgeFlow (approved):
LGTM

pedrobaeza (approved):

PR #466, opened 2026-09-17
[IMP] delete_records_safely_by_xml_id: recursive child deletion by parent field

MiquelRForgeFlow:
I tried to solve all issues that surfaced in https://github.com/OCA/openupgradelib/pull/455 discussion. Hope you like it.


openupgradelib/openupgrade.py
```diff
@@ -3675,13 +3730,23 @@ def delete_records_safely_by_xml_id(env, xml_ids, delete_childs=False):
...
+            if record._name == "ir.ui.view" and not record_parent_field_name:
+                record_parent_field_name = "inherit_id"
+            elif delete_childs and not record_parent_field_name:
+                record_parent_field_name = _get_parent_field_name(record)
+            if delete_childs and record_parent_field_name:
+                child_records = _get_child_records(record, record_parent_field_name)
+                if child_records:
+                    delete_records_safely_by_xml_id(
+                        env,
+                        [x for x in child_records.get_external_id().values() if x],
```
hbrunn: child records aren't guaranteed to have an xmlid, that seems to be pretty much the point of the whole enterprise to me. so this should try to find xmlids, but delete the children without one regardless

openupgradelib/openupgrade.py
```diff
@@ -3675,13 +3730,23 @@ def delete_records_safely_by_xml_id(env, xml_ids, delete_childs=False):
...
             if not record:
                 continue
-            if delete_childs:
-                child_and_parent_records = env["ir.ui.view"].search(
-                    [("inherit_id", "child_of", record.id)], order="id desc"
-                )
-                safe_unlink(child_and_parent_records, do_raise=True)
-            else:
-                safe_unlink(record, do_raise=True)
+            # Resolved per record, as the XML-IDs can belong to several models
```
hbrunn: this comment baffles me

openupgradelib/openupgrade.py
```diff
@@ -3675,13 +3730,23 @@ def delete_records_safely_by_xml_id(env, xml_ids, delete_childs=False):
...
+            record_parent_field_name = parent_field_name
+            if record._name == "ir.ui.view" and not record_parent_field_name:
+                record_parent_field_name = "inherit_id"
+            elif delete_childs and not record_parent_field_name:
+                record_parent_field_name = _get_parent_field_name(record)
```
hbrunn: I don't think this should exist. It's fine to keep backwards compatibility for ir.ui.view, but everything else should be `record_parent_field_name or model._parent name`. this will silently do nothing if the function is used wrongly

openupgradelib/openupgrade.py
```diff
@@ -3649,7 +3651,55 @@ def safe_unlink(records, do_raise=False):
...
+        records = record
+        while records:
+            records = model.search([(parent_field_name, "in", records.ids)])
+            records -= child_records | record
+            child_records |= records
+        return model.browse(sorted(child_records.ids, reverse=True))
+    return model.search(
+        [(parent_field_name, "child_of", record.id), ("id", "!=", record.id)],
+        order="id desc",
+    )
```
hbrunn: none of these guarantee the records are sorted leaf first. I think this should walk the hierarchy unconditionally (no child_of, no version dependence), and reverse the order

note btw that in a cycle free tree, there won't be duplicate records, so no need for the union operator

hbrunn (changes requested):

hbrunn (changes requested):
needs tests

MiquelRForgeFlow:
@hbrunn comments attended :)

openupgradelib/openupgrade.py
```diff
@@ -3676,12 +3703,22 @@ def delete_records_safely_by_xml_id(env, xml_ids, delete_childs=False):
...
-                safe_unlink(child_and_parent_records, do_raise=True)
-            else:
-                safe_unlink(record, do_raise=True)
+                child_records = _get_child_records(record, record_parent_field_name)
+                if child_records:
+                    delete_records_safely_by_xml_id(
+                        env,
+                        [x for x in child_records.get_external_id().values() if x],
+                        delete_childs=delete_childs,
+                        parent_field_name=record_parent_field_name,
```
hbrunn: this causes a lot of unnecessarily searches, as you already have all children

openupgradelib/openupgrade.py
```diff
@@ -3649,7 +3650,28 @@ def safe_unlink(records, do_raise=False):
...
+def _get_child_records(record, parent_field_name):
+    """Return the descendant records of the given record through the given
+    parent field, ordered leaf first, and excluding the record itself.
+
+    An empty recordset is returned if the model has no such field, as models
+    without a hierarchy simply have no childs to remove.
+    """
+    model = record.env[record._name].with_context(active_test=False)
+    if not parent_field_name or parent_field_name not in model._fields:
+        logger.info("Model %s has no parent field to get childs from.", record._name)
```
hbrunn:
```suggestion
        logger.error("Model %s has no parent field to get childs from.", record._name)
```

tests/test_openupgradelib.py
```diff
@@ -255,6 +255,101 @@ def test_update_module_names(self):
...
+                "name": "Test child view",
+                "model": "res.partner",
+                "inherit_id": view.id,
+                "arch": """
+                    <field name="name" position="after">
+                        <field name="function"/>
+                    </field>
+                """,
+            }
+        )
```
hbrunn, line 322: for really testing the sorting of _get_children, there should be a grandchild

tests/test_openupgradelib.py
```diff
@@ -255,6 +255,101 @@ def test_update_module_names(self):
...
+        self.assertFalse(partner.exists())
+        self.assertFalse(child.exists())
+
+    def test_delete_records_safely_by_xml_id_no_hierarchy(self):
+        """Models without a hierarchy are removed without complaining."""
+        group = self.env["res.country.group"].create({"name": "Test country group"})
+        xml_id = self._create_xml_id(group, "test_country_group")
+        openupgrade.delete_records_safely_by_xml_id(
+            self.env, [xml_id], delete_childs=True
```
hbrunn: assertLogs

hbrunn (changes requested):
the plural of "child" is "children". I find it jarring enough that we can't change the parameter name for api stability, but please don't perpetuate this in new code

openupgradelib/openupgrade.py
```diff
@@ -3676,12 +3703,25 @@ def delete_records_safely_by_xml_id(env, xml_ids, delete_childs=False):
...
+                    delete_records_safely_by_xml_id(
+                        env,
+                        [
+                            child_xml_ids[child.id]
+                            for child in child_records
+                            if child_xml_ids.get(child.id)
+                        ],
+                    )
+                    # The children without an XML-ID are removed here
+                    safe_unlink(child_records, do_raise=True)
```
hbrunn: I just realize this fails ie when there's a view with xmlid that's inherited by a view without xmlid.

so this should identify blocks of consecutive records with/without xmlid in order, and delete those

hbrunn (changes requested):

openupgradelib/openupgrade.py
hbrunn: please also update tests accordingly

MiquelRForgeFlow: Good catch!

hbrunn (approved):

sergiocorato (approved):
Tried in a migration from v.14 to v.18

pedrobaeza (commented):
I think the code is too complicated while there's no need for that. Calling `delete_records_safely_by_xml_id` for those children records with XML-ID doesn't add any advantage over calling a simple unlink over all of them at the same time. Or am I missing something?

hbrunn:
you're missing the set-xmlid-noupdate functionality in case of errors

MiquelRForgeFlow:
I prefer having working methods a bit more complicated than having simpler broken methods.

pedrobaeza (approved):

